PR #20, opened 2018-08-31
test: add a test to return null in badge component

cmcartaya:


src/components/Badge/__tests__/badge.spec.js
```diff
@@ -9,6 +9,12 @@ describe('<Badge/>', () => {
         );
         expect(component.find('span').text()).toBe('Testing Label in Badge');
     });
+    it('should not render the "Badge" when label and children is not passed', () => {
+        const component = mount(
+            <Badge label="Testing Label in Badge" />,
+        );
+        expect(component.find('span.rainbow-badge').exists()).toBe(true);
+    });
```
reiniergs: in this test you are passing the label, I'm confused here because you are saying `it('should not render the "Badge" when label and children is not passed',`

PR #31, opened 2018-09-01
feat: implement rainbow spinner

AlejandroYanes:
fixes: #23


src/components/Spinner/styles.scss
```diff
@@ -0,0 +1,223 @@
...
+    top: 50%;
+    left: 50%;
+    margin-top: -4px;
+    margin-left: -4px;
+    border-radius: 100px;
+    animation: ball-spin-clockwise 1s infinite ease-in-out;
+}
+
+.rainbow-spinner_xx-small {
+    height: 16px;
```
LeandroTorresSicilia: let's use rem here. no more px

PR #33, opened 2018-09-01
fix: change progress-bar to circular

AlejandroYanes:
fixes: #12 
Changed progess bar to circular and variant prop to represent the color of it


src/components/ProgressBar/index.js
```diff
@@ -59,13 +58,9 @@ ProgressBar.propTypes = {
         'medium',
         'large',
     ]),
-    /** It is the progress bar color. The default value is brand. */
-    color: PropTypes.oneOf([
-        'brand', 'success',
-    ]),
     /** The variant of the progress bar. Valid values are default and circular. */
```
LeandroTorresSicilia: change the description to:
```/** The variant of the progress bar. Valid values are brand and success. */```

File: src/components/ProgressBar/styles.scss [hunk 2]
```diff
@@ -43,7 +44,7 @@
     background: $color-success;
 }
 
-.rainbow-progress-bar_circular .rainbow-progress-bar__value {
+.rainbow-progress-bar .rainbow-progress-bar__value {
     border-radius: 0.5rem;
```
LeandroTorresSicilia: .rainbow-progress-bar already have the ```border-radius: 0.5rem;``` and remove the:
```
.rainbow-progress-bar_circular {
     border-radius: 0.5rem;
 }
```
above in the code since circular will not be more a variant
and remove all vertical class since we will not use them.

PR #114, opened 2018-09-07
styles: dataTable properties

TahimiLeonBravo:


library/styleguideComponents/ReactComponent/index.js
```diff
@@ -3,8 +3,12 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import Pathline from 'react-styleguidist/lib/rsg-components/Pathline';
 import Description from './description';
+import Card from './../../../src/components/Card';
 import './styles.css';
 
+const { FontAwesomeIcon } = require('@fortawesome/react-fontawesome');
```
LeandroTorresSicilia: use imports here

PR #171, opened 2018-09-13
feat: Implement RequiredAsterisk component.

### LINKED ISSUE #128: implement RequiredAsterisk component to use in all inputs

leyanessantiago:
fixes #128


src/components/AssistiveText/index.js
```diff
@@ -1,12 +1,14 @@
...
 export default function AssistiveText({ text }) {
-    if (text) {
-        return <span className="rainbow-assistive-text">{text}</span>;
-    }
-    return null;
+    return (
+        <RenderIf isTrue={!!text}>
+            <span className="rainbow-assistive-text">{text}</span>;
+        </RenderIf>
+    );
```
reiniergs: I'm not sure this is a good practice in this particular case, since there is component creation time added for RenderIf ... 
```
if (text) {
    return <span className="rainbow-assistive-text">{text}</span>;
}
```
its a simpler and faster solution here. We should be clear where of not we should use RenderIf ... to me is more within more markup where you want to do separation of concern in between html & js 

FYI: @LeandroTorresSicilia

PR #209, opened 2018-09-21
Implement tabset

### LINKED ISSUE #202: Implement Tabset component
Tabset component implements only the header bar of the experience.
- this means there is no content tight to the tabset implementation 
- tabset only notify the consumer via event that a tab was clicked then is up to the developer to changes the content(body) of the tab
- keyboard navigation and a11y mus be supported 

## Usage
```
<Tabset onSelect={handleTabSelection} activeTabName="issues">
    <Tab label="Github Issues" name="issues" />
    <Tab label="Pull Request" name="pr" />
</Tab> 
```

Notice the tabset & tab are pure components without stage, tab are selected based on the activeTabName

AlejandroYanes:
fixes #202


integration/specs/tabset-1.spec.js
```diff
@@ -0,0 +1,35 @@
...
+        browser.refresh();
+    });
+
+    it('should select the first tab when selected', () => {
+        const tabset = new PageTabset(TABSET);
+        const tabItem = tabset.getItem(0);
+        tabItem.click();
+        expect(tabItem.isSelected()).toBe(true);
+    });
+    it('should select the secound tab when the first has focus and down key is pressed', () => {
```
LeandroTorresSicilia: second ( but to be more precise put last tab because it will select the last tab, not always de second is the last ) and left arrow key

integration/specs/tabset-1.spec.js
```diff
@@ -0,0 +1,35 @@
...
+        const tabset = new PageTabset(TABSET);
+        const tabItem = tabset.getItem(0);
+        tabItem.click();
+        browser.keys(LEFT_ARROW);
+        const tabItem2 = tabset.getItem(1);
+        expect(tabItem2.isSelected()).toBe(true);
+    });
+    it('should select the first tab when the first has focus and down key is pressed twice', () => {
```
LeandroTorresSicilia: change this test for when press right arrow key since left is already tested

integration/specs/tabset-1.spec.js
```diff
@@ -0,0 +1,35 @@
...
+        const tabset = new PageTabset(TABSET);
+        const tabItem = tabset.getItem(0);
+        tabItem.click();
+        browser.keys(LEFT_ARROW);
+        browser.keys(LEFT_ARROW);
+        expect(tabItem.isSelected()).toBe(true);
+    });
+});
```
LeandroTorresSicilia: add tests for tab use cases

library/setup.js
```diff
@@ -4,10 +4,12 @@ import VerticalSectionOverflow from './../src/components/VerticalSectionOverflow
 import VerticalItem from './../src/components/VerticalItem';
 import GlobalHeader from './../library/exampleComponents/GlobalHeader';
 import CodeEditor from './pages/components/CodeEditor';
+import Tab from '../src/components/Tab/';
 
 global.MenuItem = MenuItem;
```
LeandroTorresSicilia: remove all components here, only keep those that are not part of the components like GlobalHeader and CodeEditor

src/components/Tab/__test__/tab.spec.js
```diff
@@ -0,0 +1,39 @@
...
+    });
+    it('should set the right class names when custom class name is passed', () => {
+        const component = mount(<Tab name="tab-1" className="custom-class-name" registerTab={registerTabMockFn} />);
+
+        expect(component.find('li.rainbow-tab.custom-class-name').exists()).toBe(true);
+    });
+    it('should call onSelect when clicked', () => {
+        const onSelectMockFn = jest.fn();
+        const component = mount(<Tab name="tab-1" onSelect={onSelectMockFn} registerTab={registerTabMockFn} />);
+        const aComponent = component.find('a');
```
LeandroTorresSicilia: a better name could be anchor. <a> stands for anchor.

src/components/Tab/__test__/tab.spec.js
```diff
@@ -0,0 +1,39 @@
...
+    });
+    it('should not call onSelect when clicked if disabled is passed', () => {
+        const onSelectMockFn = jest.fn();
+        const component = mount(<Tab name="tab-1" onSelect={onSelectMockFn} disabled registerTab={registerTabMockFn} />);
+        const aComponent = component.find('a');
+        aComponent.simulate('click');
+
+        expect(onSelectMockFn).toHaveBeenCalledTimes(0);
+    });
+});
```
LeandroTorresSicilia: add this tests:
- should set role to presentation in li element.
- should set the title passed as title in li element.
- should set role to tab in a element.
- should set aria-selected to true in a element when the tab is selected
- should set tabIndex to 0 in a element when tab is selected
- should set tabIndex to -1 in a element when tab is not selected

src/components/Tab/index.js
```diff
@@ -0,0 +1,86 @@
...
+    handleSelect() {
+        const { disabled, onSelect, name } = this.props;
+        if (!disabled) {
+            onSelect(name);
+        }
+    }
+
+    render() {
+        const { label, icon, style } = this.props;
+        const tabIndex = this.isSelected() ? '0' : '-1';
```
LeandroTorresSicilia: put this logic in a method:
```
getTabIndex() {
    if (this.isSelected()) {
        return 0;
    }
    return -1;
}
```

src/components/Tab/index.js
```diff
@@ -0,0 +1,86 @@
...
+
+        return (
+            <li
+                className={this.getTabClassName()}
```
LeandroTorresSicilia: add a title prop to that is passed as title to li element and add test it.

src/components/Tab/index.js
```diff
@@ -0,0 +1,86 @@
...
+                style={style}
+                role="presentation">
+                <a
+                    href="javascript:void(0);"
```
LeandroTorresSicilia: add an id prop for accessibility.
add an ariaControls props passed as aria-controls in a element for accessibility.
add tests for both

src/components/Tab/index.js
```diff
@@ -0,0 +1,86 @@
...
+                    role="tab"
+                    aria-selected={this.isSelected()}
+                    onClick={() => this.handleSelect()}
+                    tabIndex={tabIndex}>
+
+                    {icon}
```
LeandroTorresSicilia: remove icon prop

src/components/Tab/index.js
```diff
@@ -0,0 +1,86 @@
...
+                    <span>{label}</span>
```
LeandroTorresSicilia: remove the span

src/components/Tab/index.js
```diff
@@ -0,0 +1,86 @@
...
+export default function Tab(props) {
+    return (
+        <Consumer>
+            {context => <TabItem {...props} {...context} />}
+        </Consumer>
+    );
+}
+
+Tab.propTypes = {
+    label: PropTypes.string,
```
LeandroTorresSicilia: add descriptions for all props. Take a look to other component prop descriptions

src/components/Tab/index.js
```diff
@@ -0,0 +1,86 @@
...
+    name: PropTypes.string,
+    icon: PropTypes.node,
+    disabled: PropTypes.bool,
+    className: PropTypes.string,
+    style: PropTypes.object,
```
LeandroTorresSicilia: in ariaControls and id descriptions explain that for example:
id prop is to be associated with the aria-labelledby attribute of the container that show the content of this tab
also the ariaControls prop is associated with the id attribute of the container that show the content of this tab

src/components/Tab/index.js
LeandroTorresSicilia: label is one of string or node. Take a look to other components label prop types

src/components/Tabset/__test__/tabset.spec.js
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
+import { mount } from 'enzyme';
+import Tabset from '../index';
+import Tab from '../../Tab';
+
+const registerTabMockFn = jest.fn();
+
+describe('<Tabset />', () => {
+    it('should render the children passed', () => {
```
LeandroTorresSicilia: you don't need to pass a Tab to test this instead you could pass anything like some string

integration/specs/tabset-1.spec.js
```diff
@@ -16,20 +17,35 @@ describe('Tabset with href example', () => {
...
+    it('should select the last tab when the first has focus and left arrow key is pressed', () => {
         const tabset = new PageTabset(TABSET);
         const tabItem = tabset.getItem(0);
         tabItem.click();
         browser.keys(LEFT_ARROW);
         const tabItem2 = tabset.getItem(1);
         expect(tabItem2.isSelected()).toBe(true);
     });
-    it('should select the first tab when the first has focus and down key is pressed twice', () => {
+    it('should select the first tab when the first has focus and right arrow key is pressed twice', () => {
```
LeandroTorresSicilia: only test that when press right arrow move focus to next

src/components/Tab/__test__/tab.spec.js
```diff
@@ -31,9 +31,57 @@ describe('<Tab />', () => {
...
+
+        expect(anchor.prop('tabIndex')).toBe(-1);
+    });
+    it('should set id in tha a element.', () => {
+        const component = mount(<Tab name="tab-1" id="tab 1" />);
+        const anchor = component.find('a');
+
+        expect(anchor.prop('id')).toBe('tab 1');
+    });
+    it('should set id in tha a element.', () => {
```
LeandroTorresSicilia: change this name

src/components/Tab/__test__/tab.spec.js
```diff
@@ -31,9 +31,57 @@ describe('<Tab />', () => {
...
+
+        expect(anchor.prop('tabIndex')).toBe(0);
+    });
+    it('should set tabIndex to -1 in a element when tab is not selected', () => {
+        const component = mount(<Tab name="tab-1" activeTabName="tab-2" />);
+        const anchor = component.find('a');
+
+        expect(anchor.prop('tabIndex')).toBe(-1);
+    });
+    it('should set id in tha a element.', () => {
```
LeandroTorresSicilia: in the a

PR #222, opened 2018-09-23
feat: ui notification component

### LINKED ISSUE #220: Implement UI Notification component

TahimiLeonBravo:
fixes #220


src/components/Notification/icons/index.js
```diff
@@ -0,0 +1,54 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
+import InfoIcon from './infoIcon';
+import ErrorIcon from './errorIcon';
+import SuccessIcon from './successIcon';
+import WarningIcon from './warningIcon';
+import '../styles.css';
```
LeandroTorresSicilia: you don't need to import the css file again if it is already imported in index of Notification

src/components/Notification/icons/index.js
```diff
@@ -0,0 +1,54 @@
...
+    info: () => <InfoIcon />,
+    error: () => <ErrorIcon />,
+    warning: () => <WarningIcon />,
+    success: () => <SuccessIcon />,
+};
+
+export default function Icon(props) {
+    const {
+        className,
+        style,
```
LeandroTorresSicilia: This component does not receive a className and a style prop. Remember this props are only passed to the container component.

src/components/Notification/icons/index.js
```diff
@@ -0,0 +1,54 @@
...
+        icon,
+    } = props;
+
+    function getClassName() {
+        return classnames('rainbow-notification_icon-container', {
+            [`rainbow-notification_icon--${icon}`]: typeof icon === 'string',
+        }, className);
```
LeandroTorresSicilia: remove className

src/components/Notification/icons/index.js
```diff
@@ -0,0 +1,54 @@
...
+
+    if (typeof icon === 'string') {
+        return (
+            <div className={getClassName()} style={style}>
+                {iconMap[icon]()}
+            </div>
+        );
+    }
+    return (
+        <div className={getClassName()} style={style}>
```
LeandroTorresSicilia: Since className is not needed you can use 'rainbow-notification_icon-container' className directly here. Also remove style.

src/components/Notification/index.js
```diff
@@ -0,0 +1,88 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
+import CloseIcon from './closeIcon';
+import ButtonIcon from './../ButtonIcon';
+import RenderIf from '../RenderIf';
+import Icon from './icons/index';
```
LeandroTorresSicilia: import Icon from './icons';

src/components/Notification/index.js
```diff
@@ -0,0 +1,88 @@
...
+
+    return (
+        <div className={getClassName()} style={style}>
+            <a className="rainbow-notification_anchor">
+                <RenderIf isTrue={!!icon}>
+                    <Icon icon={icon} />
+                </RenderIf>
+                <span>
+                    <RenderIf isTrue={!!title}>
+                        <h1 className="rainbow-notification_title">{title}</h1>
```
LeandroTorresSicilia: here we should wrap this on a new component that check if title is a string:
```
export default function Header({ title }) {
    if (typeof title === 'string') {
        return (
            <h1 className="rainbow-notification_title">{title}</h1>
        );
    }
    return title;
}
```
We should leave that the user pass as title whatever he want and can be a node and if it is then it should not be wrapped by h1.

src/components/Notification/index.js
```diff
@@ -0,0 +1,88 @@
...
+                    </RenderIf>
+                    <RenderIf isTrue={!!description}>
+                        <p className="rainbow-notification_description">{description}</p>
```
LeandroTorresSicilia: here same ^^

src/components/Notification/icons/index.js
```diff
@@ -15,40 +14,32 @@ const iconMap = {
 };
 
 export default function Icon(props) {
-    const {
-        className,
-        style,
-        icon,
-    } = props;
+    const { icon } = props;
```
LeandroTorresSicilia: here you can use:
```
export default function Icon({ icon }) {
}
```
and then you don't need to use:
```const { icon } = props;```

PR #227, opened 2018-09-25
refactor: add unit test in notification component

cmcartaya:


src/components/Notification/__test__/icon.spec.js
```diff
@@ -0,0 +1,54 @@
...
+            <Icon icon={<svg />} />,
+        );
+        expect(component.find('svg').exists()).toBe(true);
+    });
+    it('should render the "InfoIcon" when the string "info" is passed as icon', () => {
+        const component = mount(
+            <Icon icon="info" />,
+        );
+        expect(component.find('InfoIcon').exists()).toBe(true);
+    }); it('should render the "ErrorIcon" when the string "error" is passed as icon', () => {
```
LeandroTorresSicilia: what is this?
new line

src/components/Notification/__test__/icon.spec.js
```diff
@@ -0,0 +1,54 @@
...
+        const component = mount(
+            <Icon icon="error" />,
+        );
+        expect(component.find('ErrorIcon').exists()).toBe(true);
+    }); it('should render the "WarningIcon" when the string "warning" is passed as icon', () => {
```
LeandroTorresSicilia: same^^

src/components/Notification/__test__/icon.spec.js
```diff
@@ -0,0 +1,54 @@
...
+        const component = mount(
+            <Icon icon="warning" />,
+        );
+        expect(component.find('WarningIcon').exists()).toBe(true);
+    }); it('should render the "SuccessIcon" when the string "success" is passed as icon', () => {
```
LeandroTorresSicilia: same^^

src/components/Notification/__test__/notification.spec.js
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import { mount } from 'enzyme';
+import Notification from './../index';
+
+describe('<Notification/>', () => {
+    it('should call onClick function when someone click over', () => {
```
LeandroTorresSicilia: should fire an event when the close button is clicked

src/components/Notification/icons/index.js
```diff
@@ -23,7 +30,7 @@ export default function Icon({ icon }) {
     if (typeof icon === 'string') {
         return (
             <div className={getClassName()}>
```
LeandroTorresSicilia: sorry, it is better if we return this container inside VariantIcon, thus if a wrong string is passed it is not rendered. So VariantIcon would be:
```
function VariantIcon({ icon }) {
    if (iconMap[icon]) {
        return (
            <div className={getClassName()}>
                {iconMap[icon]()}
            </div>
        );
    }
    return null;
}
```
Also you need to test this case (when a wrong string is passed), that null is returned or nothing is rendered.

PR #250, opened 2018-10-03
feat: accordion and accordionSection implementation

cmcartaya:


src/components/Accordion/index.js
```diff
@@ -0,0 +1,96 @@
...
+        const { activeSectionNames } = props;
+        if (activeSectionNames && activeSectionNames !== state.activeNames) {
+            return {
+                activeNames: activeSectionNames,
+            };
+        }
+        return null;
+    }
+
+    getContainerClassName() {
```
LeandroTorresSicilia: remove this.

src/components/Accordion/index.js
```diff
@@ -0,0 +1,96 @@
...
+        const { className } = this.props;
+        return classnames(className);
+    }
+
+    handleChangeExpanded(event, name) {
```
LeandroTorresSicilia: handleToggleSection

src/components/Accordion/index.js
```diff
@@ -0,0 +1,96 @@
...
+        return this.setState({ activeNames: name });
+    }
+
+    render() {
+        const { id, children, style, multiple } = this.props;
+        const { activeNames } = this.state;
+        return (
+            <ul
+                id={id}
+                className={this.getContainerClassName()}
```
LeandroTorresSicilia: here put className={className}. You don't need to use a method for this.

src/components/Accordion/index.js
```diff
@@ -0,0 +1,96 @@
...
+                style={style}>
+
+                <Provider value={{
+                    multiple,
+                    activeNames,
```
LeandroTorresSicilia: put this in a const context.

src/components/Accordion/index.js
```diff
@@ -0,0 +1,96 @@
...
+    * @ignore
+    */
+    children: PropTypes.node,
+    /** If true, expands multiples AccordionSections.
+    * This value defaults to false. */
+    multiple: PropTypes.bool,
+    /** Action fired when an AccordionSection is selected.
+    * The event params include the `name` of the selected AccordionSection. */
+    onToggleSection: PropTypes.func,
+    /** The name of the AccordionSection that is selected.
```
LeandroTorresSicilia: a better description is needed here because it can be a string or an array when multiple is true.

src/components/AccordionSection/Readme.md
```diff
@@ -0,0 +1,153 @@
...
+                    A rainbow is a meteorological phenomenon that is caused by reflection,
+                    refraction and dispersion of light in water droplets resulting in a spectrum
+                    of light appearing in the sky.
+                </AccordionSection>
+            </Accordion>
+        </Card>
+    </div>
+
+
+##### accordion with icons
```
LeandroTorresSicilia: here put only one example

src/components/AccordionSection/check-is-in-array.js
```diff
@@ -0,0 +1,3 @@
+export default function isInArray(array, item) {
```
LeandroTorresSicilia: Let's change the file name to be the same the function name: is-in-array. If we would use this name then should be check-if-is-in-array.

src/components/AccordionSection/index.js
```diff
@@ -0,0 +1,176 @@
...
+import classnames from 'classnames';
+import RenderIf from './../RenderIf';
+import { uniqueId } from './../../libs/utils';
+import { Consumer } from './../Accordion/context';
+import ButtonIcon from './../ButtonIcon';
+import RightArrow from './rightArrow';
+import isInArray from './check-is-in-array';
+import './styles.css';
+
+class AccordionItem extends Component {
```
LeandroTorresSicilia: change this name to AccordionSection

src/components/AccordionSection/index.js
```diff
@@ -0,0 +1,176 @@
...
+    constructor(props) {
+        super(props);
+        this.accordionDetailsId = uniqueId('accordion-details');
+        this.name = uniqueId('accordion-section');
+        this.handleSelect = this.handleSelect.bind(this);
+    }
+
+    getContainerClassNames() {
+        const { className, disabled } = this.props;
+        return classnames('rainbow-accordion_container', {
```
LeandroTorresSicilia: You must use rainbow-accordion-section_container since this component is the AccordionSection and not the Accordion. This apply for all classNames in this component.

src/components/AccordionSection/index.js
```diff
@@ -0,0 +1,176 @@
...
+
+    getTabIndex() {
+        const { disabled } = this.props;
+        if (disabled) {
+            return -1;
+        }
+        return undefined;
+    }
+
+    resolveNamesWhenMultiple() {
```
LeandroTorresSicilia: resolveActiveNamesWhenMultiple

src/components/AccordionSection/index.js
```diff
@@ -0,0 +1,176 @@
...
+        if (activeNames === undefined) {
+            return [nameToToggle];
+        }
+        if (isInArray(activeNames, nameToToggle)) {
+            return activeNames;
+        }
+        return activeNames.concat([nameToToggle]);
+    }
+
+    resolveNames() {
```
LeandroTorresSicilia: resolveActiveNames

src/components/AccordionSection/index.js
```diff
@@ -0,0 +1,176 @@
...
+        return (
+            <li className={this.getContainerClassNames()} style={style} disabled={disabled}>
+                <section>
+                    <div className="rainbow-accordion_summary">
+                        <h3 className="rainbow-accordion_summary-heading">
+                            <button
+                                className="rainbow-accordion_summary-button-heading"
+                                tabIndex={this.getTabIndex()}
+                                aria-controls={this.accordionDetailsId}
+                                aria-expanded={this.isSelected}
```
LeandroTorresSicilia: this.isSelected does't exist in this class component then it will be always undefined. Here you must to use this.isExpanded() instead.

src/components/AccordionSection/index.js
```diff
@@ -0,0 +1,176 @@
...
+                                onClick={this.handleSelect}>
+                                <RenderIf isTrue={!!icon}>
+                                    <span className="rainbow-accordion_summary-icon" aria-hidden="true">
+                                        {icon}
+                                    </span>
+                                </RenderIf>
+                                <RenderIf isTrue={!!label}>
+                                    <span title="Accordion Label">
```
LeandroTorresSicilia: here you need to use a className to truncate and show ellipsis when the title is too large. Something like this:
```
.rainbow-accordion-section_label {
    overflow: hidden;
    text-overflow: ellipsis;
    white-space: nowrap;
}
```

src/components/AccordionSection/index.js
```diff
@@ -0,0 +1,176 @@
...
+                            </button>
+                        </h3>
+
+                        <ButtonIcon
+                            className="rainbow-accordion_summary-button"
+                            size="x-small"
+                            disabled={disabled}
+                            onClick={this.handleSelect}
+                            assistiveText={assistiveText}
+                            aria-haspopup="true"
```
LeandroTorresSicilia: this is wrong aria-haspopup is an native attribute and ButtonIcon use a prop named ariaHaspopup and also you don't need to set it to "true", only put ariaHaspopup. In react an attribute without value means true so never use attribute="true" instead only put attribute.

src/components/AccordionSection/index.js
```diff
@@ -0,0 +1,176 @@
...
+                            icon={
+                                <RightArrow isExpanded={this.isExpanded()} disabled={disabled} />
```
LeandroTorresSicilia: since this.isExpanded() is being called several times put it up (inside the render method) in a variable in order to call it only one time.

src/components/Accordion/index.js
```diff
@@ -39,19 +33,20 @@ export default class Accordion extends Component {
     }
 
     render() {
-        const { id, children, style, multiple } = this.props;
+        const { id, children, style, multiple, className } = this.props;
         const { activeNames } = this.state;
+        const context = {
```
LeandroTorresSicilia: sorry let's put this in state because there is not need to recreate this object in each rerender:
```
this.state = {
    activeNames: props.activeSectionNames,
    multiple: props.multiple,
    privateOnToggleSection: this.handleToggleSection,
};
```
but be careful now:
```this.handleToggleSection = this.handleToggleSection.bind(this);```
must be above the state
then use:
```<Provider value={this.state}>```

File: src/components/Accordion/index.js [hunk 17]
```diff
@@ -77,7 +72,9 @@ Accordion.propTypes = {
     /** Action fired when an AccordionSection is selected.
     * The event params include the `name` of the selected AccordionSection. */
     onToggleSection: PropTypes.func,
-    /** The name of the AccordionSection that is selected.
+    /** It contain the name of the AccordionSection that is expanded.
+    * It can be an array of string when multiple is true,
```
LeandroTorresSicilia: It is an array ...

PR #254, opened 2018-10-04
fix: fixed ga tracking only in production and track when navigate bet…

### LINKED ISSUE #237: fix: remove analytics when CI run.

LeandroTorresSicilia:
fixes #237

### DIFF
--- a/library/ga.js
+++ b/library/ga.js
@@ -1,6 +1,16 @@
 import ReactGA from 'react-ga';
 
 const trackId = 'UA-126214096-1';
-ReactGA.initialize(trackId);
+const isProduction = process.env.NODE_ENV === 'production';
 
-ReactGA.pageview(window.location.hash);
+ReactGA.initialize(trackId, {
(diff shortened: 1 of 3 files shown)


library/ga.js
```diff
@@ -1,6 +1,16 @@
 import ReactGA from 'react-ga';
 
 const trackId = 'UA-126214096-1';
```
reiniergs: we should use env variable for this ua id ... it should be explicit in our github repo

library/ga.js
```diff
@@ -1,6 +1,16 @@
...
-ReactGA.initialize(trackId);
+const isProduction = process.env.NODE_ENV === 'production';
 
-ReactGA.pageview(window.location.hash);
+ReactGA.initialize(trackId, {
+    testMode: !isProduction,
```
reiniergs: avoid negation, use the condition directly here `process.env.NODE_ENV !== 'production'`

PR #276, opened 2018-10-08
refactor: add key board navigation in accordion component

cmcartaya:


src/components/Accordion/index.js
```diff
@@ -1,6 +1,11 @@
 import React, { Component } from 'react';
```
LeandroTorresSicilia: there are few issues:
* when the first item is focused and press arrow up key.
* when the first item is focused and press arrow down key (need to press two times).
* sometimes when the first item is focused and press arrow down key focus the last item.
*when combining tab with key it mess up.

src/components/Accordion/index.js
```diff
@@ -9,10 +14,23 @@ export default class Accordion extends Component {
...
+            [LEFT_KEY]: () => this.selectAccordionSection(LEFT_SIDE),
+            [DOWN_KEY]: () => this.selectAccordionSection(RIGHT_SIDE),
+            [UP_KEY]: () => this.selectAccordionSection(LEFT_SIDE),
+        };
+
         this.state = {
             activeNames: props.activeSectionNames,
             multiple: props.multiple,
             privateOnToggleSection: this.handleToggleSection,
+            accordionSection: [],
```
LeandroTorresSicilia: accordionSection? This is an accordionSection?
childrenRegistered or accordionChildren

src/components/Accordion/index.js
```diff
@@ -26,23 +44,75 @@ export default class Accordion extends Component {
...
+        this.setState({ accordionSection: newAccordionSectionChildren });
     }
 
     render() {
         const { id, children, style, className } = this.props;
         return (
             <ul
+                ref={this.containerRef}
+                onKeyDown={this.handleKeyPressed}
+                role="tablist"
```
LeandroTorresSicilia: role="tablist"? Are you sure?
please review accordion markup and check if ul element have a role.

src/components/Accordion/index.js
```diff
@@ -26,23 +44,75 @@ export default class Accordion extends Component {
...
                 id={id}
                 className={className}
                 style={style}>
 
-                <Provider value={this.state}>
+                <Provider value={{
+                    ...this.state,
+                    privateRegisterAccordionSection: this.registerAccordionSection,
+                    privateUnregisterAccordionSection: this.unregisterAccordionSection,
```
LeandroTorresSicilia: again really!!!
no more js with jxs
remember separation of concerns and also you are recreating a new object in each rerender
put all this in state an then use:
```<Provider value={this.state}>```

File: src/components/AccordionSection/index.js [hunk 5]
```diff
@@ -17,6 +17,26 @@ class AccordionItem extends Component {
...
+    }
+
+    componentDidMount() {
+        const { privateRegisterAccordionSection, name, disabled } = this.props;
+        const currentName = name || this.name;
+        if (!disabled) {
+            return setTimeout(() => privateRegisterAccordionSection(
+                {
+                    name: currentName,
+                    ref: this.accordionSectionRef.current,
```
LeandroTorresSicilia: this is ugly you can instead:
```
return setTimeout(() => privateRegisterAccordionSection({
    name: currentName,
    ref: this.accordionSectionRef.current,
}), 0);
```

PR #282, opened 2018-10-11
feat: implement GoogleMap component

LeandroTorresSicilia:


src/components/GoogleMap/component.js
```diff
@@ -0,0 +1,121 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
+import RenderIf from './../RenderIf';
+import Header from './header';
+import { Provider } from './context';
+import './styles.css';
+
+const currentInfoWindow = Symbol('infowindow');
```
reiniergs: user same name in the Symbol string argument

PR #285, opened 2018-10-12
test: add integration test in accordion component

cmcartaya:


integration/specs/accordion-1.spec.js
```diff
@@ -0,0 +1,141 @@
...
+} = require('./../constants');
+
+const ACCORDION = '#accordion-1';
+
+describe('Accordion whend multiple is not passed', () => {
+    beforeEach(() => {
+        browser.url('/#!/Accordion/1');
+        browser.refresh();
+    });
+    it('should expand the accordion section when the accordion section is collapsed and click on the button icon', () => {
```
LeandroTorresSicilia: should expand the accordion section when its button icon is clicked

integration/specs/accordion-1.spec.js
```diff
@@ -0,0 +1,141 @@
...
+        expect(accordionSection.isExpanded()).toBe(true);
+    });
+    it('should collapse the accordion section when click twice on the button icon', () => {
+        const accordion = new PageAccordion(ACCORDION);
+        const accordionSection = accordion.getItem(0);
+        accordionSection.clickButtonIcon();
+        accordionSection.clickButtonIcon();
+        expect(accordionSection.isExpanded()).toBe(false);
+    });
+    it('should close the firstaccordoin section and expand the second accordion section when the first accordion section is expanded and click on the second button icon', () => {
```
LeandroTorresSicilia: should close the first accordion section and expand the second when the first is expanded and is clicked the button icon of the second

integration/specs/accordion-1.spec.js
```diff
@@ -0,0 +1,141 @@
...
+        const accordion = new PageAccordion(ACCORDION);
+        const firstAccordionSection = accordion.getItem(0);
+        const secondAccordionSection = accordion.getItem(1);
+        firstAccordionSection.clickButtonIcon();
+        secondAccordionSection.clickButtonIcon();
+        expect(firstAccordionSection.isExpanded()).toBe(false);
+        expect(secondAccordionSection.isExpanded()).toBe(true);
+    });
+    it('should put the button icon focused when click on it', () => {
```
LeandroTorresSicilia: this test does't make sense since you are testing the native default behavior of buttons, with focus you only need to test the keyboard interaction to check that the right elements are focused.

integration/specs/accordion-1.spec.js
```diff
@@ -0,0 +1,141 @@
...
+        const accordion = new PageAccordion(ACCORDION);
+        const accordionSection = accordion.getItem(0);
+        accordionSection.clickButtonIcon();
+        expect(accordionSection.hasFocusButtonIcon()).toBe(true);
+    });
+    it('should lost the focus on button icon when it is focused and click on accordionSection', () => {
```
LeandroTorresSicilia: Same ^^ you are testing that the button is blurred, it is native behavior. And also we don't need a click method on accordion section page object.

integration/specs/accordion-1.spec.js
```diff
@@ -0,0 +1,141 @@
...
+        const accordion = new PageAccordion(ACCORDION);
+        const accordionSection = accordion.getItem(0);
+        accordionSection.clickButtonIcon();
+        accordionSection.click();
+        expect(accordionSection.hasFocusButtonIcon()).toBe(false);
+    });
+    it('should put the button icon focused when click on accordionSection and press the key "tab"', () => {
```
LeandroTorresSicilia: Same this test does't make sense so remove it.

integration/specs/accordion-1.spec.js
```diff
@@ -0,0 +1,141 @@
...
+        const accordion = new PageAccordion(ACCORDION);
+        const accordionSection = accordion.getItem(0);
+        accordionSection.click();
+        browser.keys(TAB_KEY);
+        expect(accordionSection.hasFocusButtonIcon()).toBe(true);
+    });
+    it('should expand the accordion section when the button icon is focused and press Space', () => {
```
LeandroTorresSicilia: same ^^

integration/specs/accordion-1.spec.js
```diff
@@ -0,0 +1,141 @@
...
+        const accordion = new PageAccordion(ACCORDION);
+        const accordionSection = accordion.getItem(0);
+        accordionSection.click();
+        browser.keys(TAB_KEY);
+        browser.keys(SPACE_KEY);
+        expect(accordionSection.isExpanded()).toBe(true);
+    });
+    it('should expand the accordion section when the button icon is focused and press Enter', () => {
```
LeandroTorresSicilia: same ^^

integration/specs/accordion-1.spec.js
```diff
@@ -0,0 +1,141 @@
...
+        const accordion = new PageAccordion(ACCORDION);
+        const accordionSection = accordion.getItem(0);
+        accordionSection.click();
+        browser.keys(TAB_KEY);
+        browser.keys(ENTER_KEY);
+        expect(accordionSection.isExpanded()).toBe(true);
+    });
+    it('should move focus to the next button icon when the first button icon is focused and press arrow down', () => {
```
LeandroTorresSicilia: goog test !!!!!

integration/specs/accordion-5.spec.js
```diff
@@ -0,0 +1,26 @@
+const PageAccordion = require('./../../src/components/Accordion/pageObject');
+
+const ACCORDION = '#accordion-multiple-1';
+
+describe('Accordion whend multiple is passed', () => {
```
LeandroTorresSicilia: when

PR #296, opened 2018-10-13
fix:library menu section should open with results when searching

cmcartaya:


src/components/VerticalSectionOverflow/__test__/verticalSectionOverflow.spec.js
```diff
@@ -118,4 +118,16 @@ describe('<VerticalSectionOverflow/>', () => {
         );
         expect(component.find('ul').text()).toBe('the section children');
     });
+    it('should fire an event when click in the VerticalSection and onToggleSection is a function', () => {
```
LeandroTorresSicilia: in the VerticalSectionOverflow

src/components/VerticalSectionOverflow/index.js
```diff
@@ -115,6 +129,8 @@ VerticalSectionOverflow.propTypes = {
     className: PropTypes.string,
     /** An object with custom style applied for the outer element. */
     style: PropTypes.object,
+    /** Action fired when a VerticalSection is clicked. */
```
LeandroTorresSicilia: Action fired when the component is clicked.

Issue #295, opened 2018-10-13
fix: the buttons variant: brand, success and destructive change the hover styles when is disabled


TahimiLeonBravo:
Expected behavior: 
All buttons when is disabled can't change the styles.

PR #299, opened 2018-10-13
style: fix the button disabled style when hover

cmcartaya:
fix: #295

### DIFF
--- a/src/components/Button/styles.scss
+++ b/src/components/Button/styles.scss
@@ -99,6 +99,10 @@
     &[disabled] {
         background-color: $color-gray-1;
         border-color: $color-gray-1;
+
+        &:hover {
+            color: $color-gray-2;
+        }
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/components/Button/styles.scss
```diff
@@ -99,6 +99,10 @@
     &[disabled] {
         background-color: $color-gray-1;
         border-color: $color-gray-1;
+
+        &:hover {
+            color: $color-gray-2;
+        }
```
TahimiLeonBravo: you don't need the pseudo-clase :hover
you only have add the ```color: $color-gray-2;``` to ```&[disabled]```
anything like that
```
&[disabled] {
        background-color: $color-gray-1;
        border-color: $color-gray-1;
        color: $color-gray-2;
}
```

src/components/Button/styles.scss
```diff
@@ -184,6 +188,10 @@
...
```
TahimiLeonBravo: same^^

src/components/Button/styles.scss
```diff
@@ -208,6 +216,10 @@
...
```
TahimiLeonBravo: same^^

PR #336, opened 2018-10-24
docs: fix the accordion description

TahimiLeonBravo:


TahimiLeonBravo:
fixes #328

src/components/Accordion/index.js
```diff
@@ -8,7 +8,8 @@ const RIGHT_SIDE = 1;
 const LEFT_SIDE = -1;
 
 /**
-* Accordions are used for freeform data entry.
+* An Accordion is is a collection of vertically stacked sections with multiple content areas.
```
LeandroTorresSicilia: fix the double is is

src/components/AccordionSection/index.js
```diff
@@ -11,6 +11,9 @@ import isInArray from './is-in-array';
 import removeItemFromArray from './remove-item-from-array';
 import './styles.css';
 
+/**
+* An AccordionSection is single section that is nested the Accordion component.
```
LeandroTorresSicilia: that is nested inside the ...

PR #354, opened 2018-10-27
test: add pagination tests

AlejandroYanes:


src/components/Pagination/__test__/pageButtons.spec.js
```diff
@@ -0,0 +1,33 @@
...
+        const component = mount(<PageButtons pages={4} onChange={() => {}} activePage={3} />);
+        const buttons = component.find('li');
+        expect(buttons.length).toBe(4);
+    });
+    it('should set the right className to the active page button', () => {
+        const component = mount(<PageButtons pages={7} onChange={() => {}} activePage={3} />);
+        const pageButtons = component.find('li.rainbow-pagination_button');
+        expect(pageButtons.get(2).props.className).toBe('rainbow-pagination_button rainbow-pagination_button--active');
+    });
+    it('should set aria-current to "page" to the active anchor element', () => {
```
LeandroTorresSicilia: add test when is not active that set aria-current to undefined

src/components/Pagination/__test__/pageButtons.spec.js
```diff
@@ -0,0 +1,33 @@
...
+        const component = mount(<PageButtons pages={7} onChange={() => {}} activePage={3} />);
+        const pageButtons = component.find('a');
+        expect(pageButtons.get(2).props['aria-current']).toBe('page');
+    });
+    it('should call onChange if a page button is clicked', () => {
+        const onChangeMockFn = jest.fn();
+        const component = mount(<PageButtons pages={1} onChange={onChangeMockFn} activePage={1} />);
+        const anchor = component.find('a');
+        anchor.simulate('click');
+        expect(onChangeMockFn).toHaveBeenCalledTimes(1);
```
LeandroTorresSicilia: check instead toHaveBeenCalledWith to test that it is called with the right arguments

src/components/Pagination/__test__/pageButtons.spec.js
```diff
@@ -0,0 +1,33 @@
...
+    });
```
LeandroTorresSicilia: add test for aria-label

src/components/Pagination/__test__/pagination.spec.js
```diff
@@ -0,0 +1,17 @@
...
+    it('should disable the previous button if the active page is the first', () => {
+        const component = mount(<Pagination pages={5} activePage={1} />);
+        const navButtons = component.find('NavigationButton');
+        expect(navButtons.get(0).props.disabled).toBe(true);
+    });
+    it('should disable the next button if the active page is the last', () => {
+        const component = mount(<Pagination pages={5} activePage={5} />);
+        const navButtons = component.find('NavigationButton');
+        expect(navButtons.get(1).props.disabled).toBe(true);
+    });
```
LeandroTorresSicilia: add test that check that aria-label  is set to pagination.
also add a11y test
add test for onClick in NavigationButton (both).

PR #362, opened 2018-10-28
feat: add isLoading to the Button component

cmcartaya:
fix: #349

### DIFF
--- a/src/components/Button/index.js
+++ b/src/components/Button/index.js
@@ -82,15 +82,13 @@ export default class Button extends Component {
             variant,
         } = this.props;
 
-        const isDisabled = this.isDisabled();
-
         return (
             <button
                 data-id="button-element"
                 id={id}
                 className={this.getClassNames()}
                 style={style}
-                disabled={isDisabled}
+                disabled={this.isDisabled()}
                 tabIndex={tabIndex}
                 onFocus={onFocus}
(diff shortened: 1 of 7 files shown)


src/components/Button/index.js
```diff
@@ -73,15 +78,19 @@ export default class Button extends Component {
             type,
             ariaHaspopup,
             id,
+            isLoading,
+            variant,
         } = this.props;
 
+        const isDisabled = this.isDisabled();
```
LeandroTorresSicilia: It is fine, but here you don't need to create a method for this since it is not used in any other place.

src/components/Button/styles.scss
```diff
@@ -217,3 +222,39 @@
 .rainbow-button--shaded {
     box-shadow: $shadow-1;
 }
+
+.rainbow-button--content-hidden {
```
LeandroTorresSicilia: I think that .rainbow-button_content--hidden follow the class convention since content is a subcomponent not a modifier.

src/components/Button/styles.scss
```diff
@@ -217,3 +222,39 @@
...
+    visibility: hidden;
+
+    .rainbow-button--spinner-visible {
```
LeandroTorresSicilia: same^^

Issue #347, opened 2018-10-26
feat: implement the Charts component


TahimiLeonBravo:

LeandroTorresSicilia:
closed by #377

PR #377, opened 2018-11-02
feat: implement chart component

AlejandroYanes:


AlejandroYanes:
fixes #347

src/components/Chart/index.js
```diff
@@ -0,0 +1,107 @@
...
+            ),
+        });
+    }
+
+    render() {
+        const { children } = this.props;
+        return (
+            <div>
+                <span className="rainbow-chart_dataset-container" aria-hidden>
+                    <Provider value={{ privateRegister: this.registerDataset }}>
```
reiniergs: every time you do something like prop={{ value  }} ... the object gets create in every execution of the render method. You can cache this instead by doing 
```
this.contextValue = {
    privateRegister: this.registerDataset.bind(this),
}
```

then in the component render you pass the contextValue directly .. created only one time

reiniergs: other observation, use Symbol to hide private methods and variables

package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "react-rainbow-components",
-    "version": "0.6.0",
+    "version": "0.5.1",
```
reiniergs: this is incorrect. Are we going back in time? :)

AlejandroYanes: sorry a got mixed on the merge

reiniergs:
can we implement this without React context, since inside of <Chart /> component you can pass <DataSet /> We can just iterate for the dataset in componentDidUpdate ... get the data passed then update the charts

this implementation does not guarantee reactivity since the charts only is created and updated one time. Every time props values change in Chart or DataSet the chart must be updated.

## tips:
- the charts instance doesn't need to be created multiple times check: https://www.chartjs.org/docs/latest/developers/updates.html

PR #379, opened 2018-11-03
fix: delete rest in buttonIcon

cmcartaya:


LeandroTorresSicilia:
add form and onKeyDown prop to Button and ButtonIcon and all props like ariaExpande, ariaControls that already exists in the ButtonIcon to the Button

src/components/Button/index.js
```diff
@@ -148,12 +158,24 @@ Button.propTypes = {
     tabIndex: PropTypes.number,
     /** The action triggered when the element is clicked. */
     onClick: PropTypes.func,
+    /** The action that will be run when a keyboard key is pressed. */
```
LeandroTorresSicilia: The action triggered when a keyboard key is pressed.

src/components/Button/index.js
```diff
@@ -148,12 +158,24 @@ Button.propTypes = {
...
     ariaHaspopup: PropTypes.bool,
+    /** A space-separated list of element IDs that
+    * this button controls the contents or presence of. */
+    ariaControls: PropTypes.string,
+    /** Indicates whether an element the button controls is expanded or collapsed.
+    * Valid values are 'true' or 'false'. */
+    ariaExpanded: PropTypes.bool,
+    /** Indicates that the element has been pressed. */
+    ariaPressed: PropTypes.bool,
+    /** Indicates the id associated to the form element. */
```
LeandroTorresSicilia: look at form attribute description on dash

PR #409, opened 2018-11-21
fix: fix reference check to active element

reiniergs:


src/components/Modal/index.js
```diff
@@ -31,6 +31,8 @@ export default class Modal extends Component {
     componentDidMount() {
         const { isOpen } = this.props;
         if (isOpen) {
+            document.body.style.overflow = 'hidden';
```
reiniergs: when you reset this style back? why is this necessary?

reiniergs: got it, it gets reset in the update methos

Issue #372, opened 2018-11-01
fix: the components should be able to be imported in the following way


TahimiLeonBravo:
```import { Button } from 'react-rainbow-components';```

PR #414, opened 2018-11-23
feat: add new components to global index export

AlejandroYanes:
#372

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,29 +1,87 @@
+import Accordion from './Accordion';
+import AccordionSection from './AccordionSection';
+import AssistiveText from './AssistiveText';
 import Avatar from './Avatar';
+import AvatarMenu from './AvatarMenu';
 import Badge from './Badge';
 import Button from './Button';
 import ButtonGroup from './ButtonGroup';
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -1,29 +1,87 @@
+import Accordion from './Accordion';
+import AccordionSection from './AccordionSection';
+import AssistiveText from './AssistiveText';
```
LeandroTorresSicilia: remove this since is for internal use

index.js
```diff
@@ -1,29 +1,87 @@
...
+import Notification from './Notification';
+import Pagination from './Pagination';
+import PrimitiveMenu from './PrimitiveMenu';
 import ProgressBar from './ProgressBar';
+import ProgressIndicator from './ProgressIndicator';
+import ProgressStep from './ProgressStep';
+import RadioGroup from './RadioGroup';
+import Rating from './Rating';
+import ReCaptcha from './ReCaptcha';
+import RenderIf from './RenderIf';
```
LeandroTorresSicilia: remove this since is for internal use

Issue #408, opened 2018-11-19
fix: fix the size of the image in CarouselCard component


TahimiLeonBravo:

TahimiLeonBravo:
<img width="365" alt="screen shot 2018-11-24 at 11 58 57 pm" src="https://user-images.githubusercontent.com/32083556/48976069-aebf9180-f045-11e8-9bf2-8a637c7825cc.png">

TahimiLeonBravo:
<img width="366" alt="screen shot 2018-11-24 at 11 59 11 pm" src="https://user-images.githubusercontent.com/32083556/48976076-cdbe2380-f045-11e8-9280-98e12de057a4.png">

TahimiLeonBravo:
<img width="438" alt="screen shot 2018-11-24 at 11 58 21 pm" src="https://user-images.githubusercontent.com/32083556/48976079-df9fc680-f045-11e8-8b34-13f528d88a96.png">

TahimiLeonBravo:
<img width="1092" alt="screen shot 2018-12-01 at 12 10 32 am" src="https://user-images.githubusercontent.com/32083556/49324903-c3949d00-f4fd-11e8-851a-76f0d1c110a8.png">

PR #431, opened 2018-11-29
fix: fix the size of the image in CarouselCard component

cmcartaya:
fix: #408

### DIFF
--- a/src/components/CarouselCard/index.js
+++ b/src/components/CarouselCard/index.js
@@ -38,6 +38,14 @@ export default class CarouselCard extends Component {
         return classnames('rainbow-carousel', className);
     }
 
+    getContainerStyle() {
+        const { style } = this.props;
+        const carouselCardContainerStyles = {
+            height: 340,
+        };
+        return Object.assign({}, carouselCardContainerStyles, style);
+    }
+
     setActiveItem(id) {
         this.setState({ activeItem: id, isAnimationPaused: true });
     }
@@ -76,10 +84,10 @@ export default class CarouselCard extends Component {
     }
 
     render() {
-        const { children, style, id } = this.props;
+        const { children, id } = this.props;
         const { childrenRegistred, activeItem, isAnimationPaused } = this.state;
         return (
-            <div className={this.getContainerClassName()} style={style} id={id}>
+            <div className={this.getContainerClassName()} style={this.getContainerStyle()} id={id}>
                 <span className="rainbow-carousel_autoplay">
                   <AnimationButton onClick={this.handleOnClick} isAnimationPaused={isAnimationPaused} />
                 </span>
--- a/src/components/CarouselImage/index.js
+++ b/src/components/CarouselImage/index.js
@@ -91,6 +91,13 @@ class Item extends Component {
         return activeItem !== this.carouselIndicatorID;
     }
 
+    getImageSrc() {
+        const { src } = this.props;
+        return {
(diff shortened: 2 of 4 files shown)


src/components/CarouselCard/__tests__/utils.spec.js
```diff
@@ -57,4 +58,38 @@ describe('<CarouselCard/> utils', () => {
...
+
+    describe('carouselCardContainerStyles', () => {
+        it('should return null if a ref is not passed', () => {
+            expect(carouselCardContainerStyles()).toBeNull();
+        });
+        it('should return null if ref passed does not have a parentNode', () => {
+            const ref = {};
+            expect(carouselCardContainerStyles(ref)).toBeNull();
+        });
+        it('should return an object with a heght of 100% if the parentNode the ref passed has an height set', () => {
```
LeandroTorresSicilia: height

src/components/CarouselCard/__tests__/utils.spec.js
```diff
@@ -57,4 +58,38 @@ describe('<CarouselCard/> utils', () => {
...
+                    style: {
+                        height: 100,
+                    },
+                },
+            };
+            expect(carouselCardContainerStyles(ref)).toEqual({
+                height: '100%',
+            });
+        });
+        it('should return an object with a heght of 340 if the parentNode of the ref passed does not have an height set', () => {
```
LeandroTorresSicilia: same ^^

src/components/CarouselCard/index.js
```diff
@@ -5,7 +5,7 @@ import classnames from 'classnames';
 import { Provider } from './context';
 import Indicators from './indicators';
 import AnimationButton from './animationButton';
-import { getItemIndex, getChildTabNodes, insertChildOrderly } from './utils';
+import { getItemIndex, getChildTabNodes, insertChildOrderly, carouselCardContainerStyles } from './utils';
```
LeandroTorresSicilia: getContainerStyles since it is a function, would be carouselCardContainerStyles if it was a const.

src/components/CarouselCard/__tests__/utils.spec.js
LeandroTorresSicilia: the alt attribute is only for img element so here we need to use the AssistiveText component in order to screen readers take the assistive text. Then:
```
<div className="rainbow-carousel-image_content-image-container">
    <div className="rainbow-carousel-image_image" style={this.getImageSrc()} />
    <AssistiveText text={assistiveText} />
</div>
```

src/components/CarouselImage/index.js
```diff
@@ -128,19 +135,19 @@ class Item extends Component {
                     className="rainbow-carousel-image"
                     tabIndex={this.getTabIndex()}
                     ref={this.itemRef}>
-                    <div>
-                        <img className="rainbow-carousel-image_image" src={src} alt={assistiveText} />
+                    <div className="rainbow-carousel-image_content-image-container">
+                        <div className="rainbow-carousel-image_image" style={this.getImageSrc()} alt={assistiveText} />
```
LeandroTorresSicilia: the alt attribute is only for img element so here we need to use the AssistiveText component in order to screen readers take the assistive text. Then:
```
<div className="rainbow-carousel-image_content-image-container">
    <div className="rainbow-carousel-image_image" style={this.getImageSrc()} />
    <AssistiveText text={assistiveText} />
</div>
```

PR #432, opened 2018-11-30
style: fixing the exampleCard in utility section.

### LINKED ISSUE #403: style: fix the exampleCard in utility section
Requirements
-The example card description should have only three lines
-Each example card should be next to the other.
-The image shold have a margin-top: 1.2rem; and margin-bottom: 1.2rem.
screen shot 2018-11-24 at 12 14 19 am

leyanessantiago:
fixes #403

### DIFF
--- a/library/styleguideComponents/ReactComponent/Utils/styles.css
+++ b/library/styleguideComponents/ReactComponent/Utils/styles.css
@@ -17,7 +17,7 @@
     margin: 0 16px 20px 16px;
 }
 
-.react-rainbow-utils_link:hover, .react-rainbow-utils_link:focus, .react-rainbow-utils_example-link:hover, .react-rainbow-utils_example-link:focus, {
+.react-rainbow-utils_link:hover, .react-rainbow-utils_link:focus, .react-rainbow-utils_example-link:hover, .react-rainbow-utils_example-link:focus {
     text-decoration: none;
 }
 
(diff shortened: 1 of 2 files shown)


library/styleguideComponents/ReactComponent/Utils/styles.css
```diff
@@ -94,6 +94,31 @@
     color: #576574;
     align-self: flex-start;
     text-align: justify;
+    overflow: hidden;
+    max-height: 4rem;
+}
+
+.react-rainbow-utils-item_description::before {
```
TahimiLeonBravo: I think this is not a good solution. We can use the`text-overflow: ellipsis` style

leyanessantiago: The `text-overflow: ellipsis` work for a single line combined to `white-space: nowrap` but not for multiple line.

library/styleguideComponents/ReactComponent/Utils/styles.css
```diff
@@ -94,6 +94,31 @@
...
+    content:'...';
+    position: absolute;
+    top: 10.6875rem;
+    right: 0.375rem;
```
LeandroTorresSicilia: with this solution you are always set the ellipsis and we need that when the description is short and fit in 3 lines then should not have ellipsis

LeandroTorresSicilia: sorry I saw that this works

PR #443, opened 2018-12-12
fix: added prop name to textarea element

AlejandroYanes:


src/components/CheckboxGroup/index.js
```diff
@@ -88,6 +89,7 @@ CheckboxGroup.propTypes = {
     label: PropTypes.oneOfType([
         PropTypes.string, PropTypes.node,
     ]),
+    name: PropTypes.string,
```
LeandroTorresSicilia: The name of the checkbox group

src/components/Textarea/index.js
```diff
@@ -149,6 +151,8 @@ Textarea.propTypes = {
     ]).isRequired,
     /** A boolean to hide the textarea label */
     hideLabel: PropTypes.bool,
+    /** The name of th input */
```
LeandroTorresSicilia: The name of the textarea

PR #444, opened 2018-12-12
feat: implement table component

### LINKED ISSUE #289: feat: implement dataTable component (1st version)
Requirements:
-in the first version, only one should list the data
-export the component in the root index
-follow the zeplin design
https://zpl.io/blPO9Kz

AlejandroYanes:
fixes #289


src/components/Table/index.js
```diff
@@ -0,0 +1,54 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import resolveColumns from './resolve-columns';
+import Rows from './body';
+import Headers from './head';
+import './styles.css';
+
```
TahimiLeonBravo: Add the component description

src/components/Table/readme.md
```diff
@@ -0,0 +1,40 @@
+#Basic Table
```
TahimiLeonBravo: you need change this title by `##### simple Table`

src/components/Table/index.js
```diff
@@ -0,0 +1,63 @@
...
+ * so that users can look for patterns and insights. */
+export default class Table extends Component {
+    constructor(props) {
+        super(props);
+        this.columns = undefined;
+    }
+
+    componentDidMount() {
+        if (!this.columns) {
+            this.getColumns();
```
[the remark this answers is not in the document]
reiniergs: I think if the columns is always generated from the column passed by the developer and the change of the column imply render of the table then we should put it in the state of the component

src/components/Table/readme.md
```diff
@@ -0,0 +1,40 @@
...
+        <GlobalHeader className="rainbow-m-bottom_xx-large" src="images/user/user3.jpg">
+            <ButtonGroup className="rainbow-m-right_medium">
+                <ButtonIcon variant="border-filled" disabled icon={<FontAwesomeIcon icon={faCog} />} />
+                <ButtonIcon variant="border-filled" disabled icon={<FontAwesomeIcon icon={faEllipsisV} />} />
+            </ButtonGroup>
+        </GlobalHeader>
+        <Table data={tableData} style={tableStyles}>
+            <Column header="Name" field="name" />
+            <Column header="Company" field="company" />
+            <Column header="email" field="email" />
```
reiniergs: we should add more examples that use the `component` prop of the Column component

src/components/Table/index.js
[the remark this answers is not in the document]
AlejandroYanes: @reiniergs but setting it in the `state` would cause an extra render cycle and as it`s pretty much a read-only property, it'll work fine as a class property

LeandroTorresSicilia:
start the library and open the console when develop, there are lot of warnings, and one is very important:
```Error: Maximum update depth exceeded.```
currently componentDidMount is being called infinite times

PR #468, opened 2018-12-23
fix: change vertical and topBar in the library

cmcartaya:


library/styleguideComponents/ReactComponent/index.js
```diff
@@ -27,6 +29,13 @@ export default class ReactComponent extends Component {
         this.setState({ activeTabName: tab });
     }
 
+    editLinkClassName() {
```
LeandroTorresSicilia: getEditLinkClassName

library/styleguideComponents/ReactComponent/index.js
```diff
@@ -27,6 +29,13 @@ export default class ReactComponent extends Component {
...
+        const isExpanded = window.location.href.includes('/#!/');
+        return classnames('react-rainbow-component_example-edit-link-container', {
+            'react-rainbow-component_example-edit-link-container--compress': isExpanded,
+        });
+    }
```
LeandroTorresSicilia: You can put this function outside the component since don't use this. Also this enables you to remove the eslint rule added.

library/styleguideComponents/StyleGuide/index.js
```diff
@@ -38,6 +39,7 @@ export default class StyleGuide extends React.Component {
                     {toc}
                 </aside>
                 <main className="react-rainbow-main-content">
+                    <ProjectSelector />
```
LeandroTorresSicilia: Why ProjectSelector inside main?
I think that it can be placed the most outer possible.

Issue #453, opened 2018-12-21
feat: add new feature to the table component


TahimiLeonBravo:
Requirements:
-The table header should be fixed in the top
-The user should  resize the columns
-The user should make the soring in the table

TahimiLeonBravo:
Fix the movement of the elements in the table when you focus on a cell

PR #488, opened 2018-12-27
add new features to the table component

AlejandroYanes:
fixes: #453


src/components/Column/index.js
```diff
@@ -31,9 +31,14 @@ Column.propTypes = {
...
     field: PropTypes.string.isRequired,
+    /** Sets whether the column should control the sorting order of the data */
+    sortable: PropTypes.bool,
+    width: PropTypes.string,
 };
 
 Column.defaultProps = {
     header: undefined,
     component: undefined,
+    sortable: true,
```
reiniergs: column shouldn't be sortable by default

src/components/Table/body/cell.js
```diff
@@ -18,9 +18,9 @@ CellValue.defaultProps = {
     value: undefined,
 };
 
-export default function Cell({ component, value }) {
+export default function Cell({ component, value, width }) {
     return (
-        <td className="rainbow-table_cell">
+        <td className="rainbow-table_cell" style={{ width }}>
```
LeandroTorresSicilia: don't use a style object directly, instead create a const:
```const cellStyle = { width };```

File: src/components/Table/head/arrowDown.js [hunk 3]
```diff
@@ -0,0 +1,35 @@
...
+        </svg>
+    );
+}
+
+ArrowDown.propTypes = {
+    direction: PropTypes.oneOf(['asc', 'desc']),
+};
+
+ArrowDown.defaultProps = {
+    direction: 'desc',
```
LeandroTorresSicilia: the default is asc

src/components/Table/head/header.js
```diff
@@ -0,0 +1,122 @@
+/* eslint-disable no-param-reassign,class-methods-use-this */
```
LeandroTorresSicilia: add tests for this file

src/components/Table/head/header.js
```diff
@@ -0,0 +1,122 @@
...
+import RenderIf from '../../RenderIf';
+import ArrowDown from './arrowDown';
+
+function getClassName(sortable, selected) {
+    return classnames(
+        'rainbow-table_header',
+        {
+            'rainbow-table_header--sortable': sortable,
+            'rainbow-table_header--selected': selected,
+        },
```
LeandroTorresSicilia: this is right but if we always put this inside the class why here it is outside?

src/components/Table/head/header.js
```diff
@@ -0,0 +1,122 @@
...
+                onClick={this.handleClick}
+                ref={this.headerContainer}>
+
+                <div className="rainbow-table_header-content-wrapper">
+                    <span title={this.getHeaderTitle()} className="rainbow-table_header-content">{content}</span>
+                    <RenderIf isTrue={isSelected}>
+                        <ArrowDown direction={sortDirection} />
+                    </RenderIf>
+                </div>
+                <div
```
LeandroTorresSicilia: here we need to use a input type="range" to make this accessible

src/components/Table/index.js
```diff
@@ -23,34 +26,68 @@ export default class Table extends Component {
         }
     }
 
+    onColumnSelect(e, columnIndex) {
+        const { onSort } = this.props;
+        onSort(e, this.state.columns[columnIndex].field);
+        this.setState({ selectedColumn: columnIndex });
```
LeandroTorresSicilia: this make that the table rerender completely and this is not needed, you can select the cell making it focusable

src/components/Table/index.js
```diff
@@ -23,34 +26,68 @@ export default class Table extends Component {
...
         this.setState({ columns: newColumns });
     }
 
-    render() {
-        const { data, className, style } = this.props;
+    resize(event, columnIndex, width) {
         const { columns } = this.state;
+        const resizedColumn = Object.assign({}, columns[columnIndex]);
+        resizedColumn.width = width;
+        columns[columnIndex] = resizedColumn;
```
LeandroTorresSicilia: Here you are mutating the state directly, this is prohibitive.

src/components/Table/index.js
```diff
@@ -23,34 +26,68 @@ export default class Table extends Component {
...
+        this.setState({ columns });
```
LeandroTorresSicilia: Here when you resize the table is completely rerendered, we need to find a way to optimize this. I think you can change only the width of the header column without recreate the table.

src/components/Column/index.js
```diff
@@ -31,9 +31,14 @@ Column.propTypes = {
     * for a column that want to represent names on a collection of people.
     */
     field: PropTypes.string.isRequired,
+    /** Sets whether the column should control the sorting order of the data */
+    sortable: PropTypes.bool,
+    width: PropTypes.string,
```
LeandroTorresSicilia: add description:
```Specifies the width of a column in pixels and makes the column non-resizable.```
and change prop type to number.

File: src/components/Table/body/cell.js [hunk 11]
```diff
@@ -18,9 +18,11 @@ CellValue.defaultProps = {
     value: undefined,
 };
 
-export default function Cell({ component, value }) {
+export default function Cell({ component, value, width }) {
+    const cellStyle = { width };
```
LeandroTorresSicilia: discard all this changes since width is not longer needed here

src/components/Table/body/row.js
```diff
@@ -5,8 +5,8 @@ import Cell from './cell';
 
 export default function Row(props) {
     const { data, columns } = props;
-    const cells = columns.map(({ component, field }) => (
-        <Cell key={uniqueId('cell')} component={component} value={data[field]} />
+    const cells = columns.map(({ component, field, width }) => (
+        <Cell key={uniqueId('cell')} component={component} value={data[field]} width={width} />
```
LeandroTorresSicilia: Same ^^. Discard changes.

src/components/Table/head/index.js
```diff
@@ -1,21 +1,62 @@
...
-    const { columns } = props;
-    if (columns) {
-        return columns.map(column => (
-            <th key={uniqueId('header')} className="rainbow-table_header">{column.header}</th>
-        ));
+export default class Head extends Component {
+    constructor(props) {
+        super(props);
+        this.state = {
+            columnWidths: props.columns.map(column => column.width),
```
LeandroTorresSicilia: what is the point of this?
it is only for change the column width?
if it is, then is better to not send a resize event to Head and instead change the width directly in the Header component with a width state.

src/components/Table/index.js
```diff
@@ -23,34 +24,46 @@ export default class Table extends Component {
...
+        const { data, sortDirection, style } = this.props;
+        const { columns, selectedColumn } = this.state;
 
         return (
-            <table className={className} style={style}>
-                <thead className="rainbow-table_head">
+            <div className={this.getContainerClassNames()} style={style}>
+                <table className="rainbow-table">
+                    <thead className="rainbow-table_head">
                     <tr className="rainbow-table_header-row">
```
LeandroTorresSicilia: indentation

src/components/Table/styles.scss
```diff
@@ -1,24 +1,133 @@
+/* stylelint-disable no-descending-specificity */
 @import "../../styles/colors";
 @import "../../styles/font-sizes";
+@import "../../styles/paddings";
+
```
LeandroTorresSicilia: this css file will become too huge, so we need to put the head specific styles in another css file inside the head folder, and the same for body.

LeandroTorresSicilia:
you forgot to add the input type range
also add the following props to Table:
- resizeColumnDisabled
- minColumnWidth (default to 50)
- maxColumnWidth(default to 1000)

By default, columns are resizable. Users can click and drag the width to a minimum of 50px and a maximum of 1000px, unless the default values are changed. To change the minimum and maximum width column, use the minColumnWidth and maxColumnWidth props.
You can disable column resizing for all columns by setting the datatable's resizeColumnDisabled prop to true. You can disable column resizing for a specific column by setting the width column prop.

src/components/Table/head/header.js
```diff
@@ -13,6 +13,21 @@ export default class Header extends Component {
         this.handleMouseUp = this.handleMouseUp.bind(this);
         this.headerContainer = React.createRef();
         this.resizeBar = React.createRef();
+        this.state = {
+            width: this.props.width,
```
LeandroTorresSicilia: here you can use:
```props.width```

File: src/components/Table/head/header.js [hunk 17]
```diff
@@ -34,30 +49,57 @@ export default class Header extends Component {
...
         this.resizeBar.current.style.transform = `translateX(${this.newXPosition}px)`;
     }
 
     handleMouseDown(dragEvent) {
         dragEvent.preventDefault();
-        this.startXPosition = dragEvent.clientX;
-        document.addEventListener('mousemove', this.handleMouseMove);
-        document.addEventListener('mouseup', this.handleMouseUp);
+        const { resizeColumnDisabled } = this.props;
+        const isResizable = !resizeColumnDisabled || this.state.width === undefined;
```
LeandroTorresSicilia: &&

src/components/Table/head/header.js
```diff
@@ -34,30 +49,57 @@ export default class Header extends Component {
...
+        if (isResizable) {
+            this.startXPosition = dragEvent.clientX;
```
LeandroTorresSicilia: we need to declare:
```this.newXPosition = 0;```
because if I click de resize bar without move it this variable is never declared and then in handleMouseUp when:
```const width = headerContainerWidth + this.newXPosition;```
we got NaN since this.newXPosition is undefined
Also we need to declare it here because whenever we click the resize bar we need to reset this value to 0 since if we move the resize bar 30px to right then this.newXPosition get 30 and when we press the resize bar again without move, it will resize the column with 30 and we don't want this.

File: src/components/Table/head/header.js [hunk 19]
```diff
@@ -34,30 +49,57 @@ export default class Header extends Component {
...
-        const headerStyles = { width };
+        const { width } = this.state;
+        const {
+            content,
+            isSelected,
+            sortDirection,
+            minColumnWidth,
+            maxColumnWidth,
+        } = this.props;
+        const headerStyles = { width: `${width}px` };
```
LeandroTorresSicilia: you can use width directly:
```const headerStyles = { width };```

PR #557, opened 2019-01-22
test: add table integration tests

AlejandroYanes:


AlejandroYanes:
fixes #553

integration/specs/Table/table-2.spec.js
```diff
@@ -0,0 +1,48 @@
...
+    });
+
+    it('should select the first header when clicked', () => {
+        const table = new PageTable(TABLE);
+        const tableHeader = table.getItem(0);
+        tableHeader.click();
+        expect(tableHeader.isSelected()).toBe(true);
+    });
+
+    it('should not select the second header when clicked', () => {
```
LeandroTorresSicilia: should not select the second header when clicked if it is not sortable

integration/specs/Table/table-2.spec.js
```diff
@@ -0,0 +1,48 @@
...
+        const table = new PageTable(TABLE);
+        const tableHeader = table.getItem(1);
+        tableHeader.click();
+        expect(tableHeader.isSelected()).toBe(false);
+    });
+
+    it('should set sortDirection to "asc" when clicked once', () => {
+        const table = new PageTable(TABLE);
```
LeandroTorresSicilia: I think this is not necessary here, with unit test it is ok

integration/specs/Table/table-2.spec.js
```diff
@@ -0,0 +1,48 @@
...
+        const tableHeader = table.getItem(1);
+        tableHeader.click();
+        expect(tableHeader.getSortDirection()).toBe('asc');
+    });
+
+    it('should change the sortDirection when clicked twice', () => {
```
LeandroTorresSicilia: same ^^

integration/specs/Table/table-2.spec.js
```diff
@@ -0,0 +1,48 @@
...
+        tableHeader.click();
+        const secondSortDirection = tableHeader.getSortDirection();
+        expect(firstSortDirection === secondSortDirection).toBe(false);
+    });
+
+    it('should resize the first header when it`s resizeBar is moved', () => {
+        const table = new PageTable(TABLE);
+        const tableHeader = table.getItem(0);
+        tableHeader.dragAndDrop(table.getItem(2).getSelector());
+        expect(tableHeader.isSelected()).toBe(false);
```
LeandroTorresSicilia: we need to find a way to get the header width, then simulate the drag and drop passing an amount of px to left or right and then expect that the header width change to the right value

src/components/Table/pageObject/index.js
```diff
@@ -0,0 +1,31 @@
...
+    constructor(rootElement) {
+        this.rootElement = rootElement;
+    }
+
+    /**
+     * Returns a new Header page object of the element in item position.
+     * @method
+     * @param {number} itemPosition - The base 0 index of the header.
+     */
+    getItem(itemPosition) {
```
LeandroTorresSicilia: we need to be more specific since in the future is likely that we need to get cells so:
```getHeaderItem```

PR #602, opened 2019-02-08
fix: the proptype of the prop value should accept a string in CheckboxGroup

cmcartaya:
fix: #590


src/components/CheckboxGroup/index.js
```diff
@@ -27,6 +27,14 @@ class CheckboxGroup extends Component {
         return undefined;
     }
 
+    getValue() {
+        const { value } = this.props;
+        if (typeof (value) === 'string') {
```
LeandroTorresSicilia: without ():
```if (typeof value === 'string') {}```
also this need to be tested

PR #612, opened 2019-02-17
fix: add ellipsis for Badge component

cmcartaya:


src/components/Badge/__tests__/content.spec.js
```diff
@@ -1,27 +1,58 @@
...
     it('should return the label when children is undefined', () => {
-        const label = 'Label in Content';
-        const children = undefined;
-        expect(Content({ label, children })).toBe(label);
+        const component = mount(
+            <Content label="Label in Content">
+                {undefined}
+            </Content>,
+        );
+        expect(component.find('span').text()).toBe('Label in Content');
```
LeandroTorresSicilia: integrate this test with the above and use an array with all falsy values except 0
also add a test for the 0 case

src/components/Badge/__tests__/content.spec.js
```diff
@@ -1,27 +1,58 @@
...
+                Children in Content
+            </Content>,
+        );
+        expect(component.find('span').prop('className')).toBe('my-custom-class-name-in-badge-when-children');
+    });
+    it('should have the right class names when label is passed and have a custom class', () => {
+        const component = mount(
+            <Content className="my-custom-class-name-in-badge-when-label" label="Label in Content" />,
+        );
+        expect(component.find('span').prop('className')).toBe('my-custom-class-name-in-badge-when-label');
```
LeandroTorresSicilia: since the Content will not receive className prop then remove this two last tests

src/components/Badge/content.js
```diff
@@ -1,6 +1,41 @@
...
+import PropTypes from 'prop-types';
+
+export default function Content(props) {
+    const {
+        className,
+        label,
+        children,
+    } = props;
+
     if (children !== null && children !== undefined) {
```
LeandroTorresSicilia: ```(children || children === 0)```

src/components/Badge/index.js
```diff
@@ -33,7 +33,7 @@ export default function Badge(props) {
 
     return (
         <span className={getClassName()} style={style}>
-            <Content label={label}>
+            <Content className="rainbow-badge_content--truncate" label={label}>
```
LeandroTorresSicilia: you can pass this class directly in the span inside Content

src/components/Badge/__tests__/content.spec.js
```diff
@@ -2,28 +2,39 @@ import React from 'react';
...
+        it(`should return the label when children is ${falsyValue}`, () => {
+            const component = mount(
+                <Content label="Label in Content">
+                    {falsyValue}
+                </Content>,
         );
-        expect(component.find('span').text()).toBe('Label in Content');
+            expect(component.find('span').text()).toBe('Label in Content');
+        });
     });
```
LeandroTorresSicilia: I think is better if you use only one test:
```
it('should return the label when children is a falsy value other than zero', () => {
    falsyValues.forEach((value) => {
        const component = mount(
                <Content label="Label in Content">
                    {value}
                </Content>,
        );
        expect(component.find('span').text()).toBe('Label in Content');
    });
});
```

src/components/Badge/readme.md
```diff
@@ -2,10 +2,16 @@
...
         </div>
         <div className="rainbow-m-horizontal_medium">
-            <Badge label="Outline Badge" variant="outline-brand" />
+            <Badge label={0} />
+        </div>
+        <div className="rainbow-m-horizontal_medium">
+            <Badge>{"0"}</Badge>
+        </div>
+        <div className="rainbow-m-horizontal_medium">
+            <Badge>{0}</Badge>
```
LeandroTorresSicilia: looks like you forgot revert this changes

PR #650, opened 2019-04-02
test: unit testing on tabset component

cmcartaya:


src/components/Tabset/__test__/tabset.spec.js
```diff
@@ -14,6 +31,22 @@ describe('<Tabset />', () => {
         );
         expect(component.find('p').text()).toBe('testing tabset');
     });
+    it('should have the right class names', () => {
+        const component = mount(
+            <Tabset className="my-custom-class-name">
+                <Tab label="Tab-1" name="tab-1" registerTab={registerTabMockFn} />
+            </Tabset>,
+        );
+        expect(component.find('div').prop('className')).toBe('rainbow-tabset my-custom-class-name');
```
LeandroTorresSicilia: this is to fragile since if in the future we add some div inside the tabset or tab then likely it will break, in this case better use:
```expect(component.find('.rainbow-tabset my-custom-class-name').exists()).toBe(true);```

File: src/components/Tabset/__test__/tabset.spec.js [hunk 2]
```diff
@@ -14,6 +31,22 @@ describe('<Tabset />', () => {
...
+    });
+    it('should have the right class names when fullWidth prop is passed', () => {
+        const component = mount(
+            <Tabset fullWidth>
+                <Tab label="Tab-1" name="tab-1" registerTab={registerTabMockFn} />
+            </Tabset>,
+        );
+        expect(component.find('ul').prop('className')).toBe('rainbow-tabset_inner-container rainbow-tabset_inner-container--full-width');
```
LeandroTorresSicilia: in this case it is right because is very difficult that another ul gets added but you can use:
```ul[role="tablist"]```

File: src/components/Tabset/__test__/tabset.spec.js [hunk 3]
```diff
@@ -43,4 +76,95 @@ describe('<Tabset />', () => {
...
+        getLeftButtonDisabledState.mockReset();
+        getLeftButtonDisabledState.mockReturnValue(true);
+        const component = mount(
+            <Tabset activeTabName="tab-1">
+                <Tab label="Tab-1" name="tab-1" registerTab={registerTabMockFn} />
+                <Tab label="Tab-2" name="tab-2" registerTab={registerTabMockFn} />
+                <Tab label="Tab-3" name="tab-3" registerTab={registerTabMockFn} />
+            </Tabset>,
+        );
+        component.instance().tabsetRef.current.scrollTo = jest.fn();
```
LeandroTorresSicilia: review if this is necessary

src/components/Tabset/__test__/tabset.spec.js
```diff
@@ -43,4 +76,95 @@ describe('<Tabset />', () => {
...
+        getRightButtonDisabledState.mockReset();
+        getRightButtonDisabledState.mockReturnValue(true);
+        const component = mount(
+            <Tabset activeTabName="tab-1">
+                <Tab label="Tab-1" name="tab-1" registerTab={registerTabMockFn} />
+                <Tab label="Tab-2" name="tab-2" registerTab={registerTabMockFn} />
+                <Tab label="Tab-3" name="tab-3" registerTab={registerTabMockFn} />
+            </Tabset>,
+        );
+        component.instance().tabsetRef.current.scrollTo = jest.fn();
```
LeandroTorresSicilia: same ^^

src/components/Tabset/__test__/tabset.spec.js
```diff
@@ -43,4 +80,200 @@ describe('<Tabset />', () => {
...
+        component.instance().updateButtonsVisibility = jest.fn();
+        component.instance().isFirstTime = true;
+        component.setState({
+            tabsetChildren: [
+                <Tab label="Tab-1" name="tab-1" registerTab={registerTabMockFn} />,
+                <Tab label="Tab-2" name="tab-2" registerTab={registerTabMockFn} />,
+                <Tab label="Tab-3" name="tab-3" registerTab={registerTabMockFn} />,
+            ],
+        });
+        component.setProps();
```
LeandroTorresSicilia: review:
1. I think you don't need to set ```component.instance().isFirstTime = true;```
since always the component instance initialize this value to true. Then instead of initialize it manually to true test that initially it is true and then that it change to false after call setState.
2 . Maybe ```component.setProps();``` is not necessary since ```component.setState()``` already update the component.

src/components/Tabset/__test__/tabset.spec.js
```diff
@@ -43,4 +80,200 @@ describe('<Tabset />', () => {
...
+        );
+        component.instance().tabsetRef.current = { offsetWidth: 500 };
+        component.setState({
+            tabsetChildren: [
+                <Tab label="Tab-1" name="tab-1" registerTab={registerTabMockFn} />,
+                <Tab label="Tab-2" name="tab-2" registerTab={registerTabMockFn} />,
+                <Tab label="Tab-3" name="tab-3" registerTab={registerTabMockFn} />,
+            ],
+        });
+        component.setProps();
```
LeandroTorresSicilia: same ^^ can be not necessary

src/components/Tabset/__test__/tabset.spec.js
```diff
@@ -43,4 +80,200 @@ describe('<Tabset />', () => {
...
+        expect(component.state().areButtonsVisible).toBe(true);
+    });
+    it('should set the areButtonsVisible to false when children total width is less than tbaset width', () => {
```
LeandroTorresSicilia: tbaset => tabset

src/components/Tabset/__test__/tabset.spec.js
```diff
@@ -43,4 +80,200 @@ describe('<Tabset />', () => {
...
+                <Tab label="Tab-2" name="tab-2" registerTab={registerTabMockFn} />
+                <Tab label="Tab-3" name="tab-3" registerTab={registerTabMockFn} />
+            </Tabset>,
+        );
+        component.instance().tabsetRef.current = {
+            scrollLeft: 600,
+            offsetWidth: 600,
+        };
+        component.instance().tabsetRef.current.scrollTo = jest.fn();
+        const item3 = component.find('Tab[name="tab-3"]').find('a');
```
LeandroTorresSicilia: change to click the second

src/components/Tabset/__test__/utils.spec.js
```diff
@@ -1,11 +1,13 @@
 import {
     insertChildOrderly,
     getChildTabNodes,
-    getActiveTabIndex,
+    getTabIndex,
```
LeandroTorresSicilia: getTabIndexFromName

PR #665, opened 2019-04-08
feat: implement chip UI

TahimiLeonBravo:


src/components/Chip/index.js
```diff
@@ -0,0 +1,71 @@
...
+                    title="Close"
+                    onClick={onDelete}
+                    assistiveText="Remove" />
+            </RenderIf>
+        </span>
+    );
+}
+
+Chip.propTypes = {
+    /** The title that appears in the Chip. */
```
LeandroTorresSicilia: look at badge description, title is the tooltip showed when hover.
also I think that we should add title prop like badge

TahimiLeonBravo:
fix: #665

PR #696, opened 2019-04-24
fix: fix some styles in Table empty

TahimiLeonBravo:


src/components/Table/index.js
```diff
@@ -120,6 +120,14 @@ export default class Table extends Component {
         return classnames('rainbow-table_container', className);
     }
 
+    getScrollableYClassNames() {
+        const { data, isLoading } = this.props;
+        const isEmpty = data.length === 0;
+        return classnames('rainbow-table_container--scrollable-y', {
+            'rainbow-align-content_center': isEmpty && !isLoading,
```
LeandroTorresSicilia: use a specific table class for this

PR #710, opened 2019-05-03
doc: add doc to Lookup props

TahimiLeonBravo:


src/components/Lookup/index.js
```diff
@@ -352,44 +352,62 @@ class Lookup extends Component {
 }
 
 Lookup.propTypes = {
+    /** Text label for the Lookup. */
+    label: PropTypes.oneOfType([
+        PropTypes.string, PropTypes.node,
+    ]).isRequired,
+    /** A boolean to hide the Lookup label. */
+    hideLabel: PropTypes.bool,
     value: PropTypes.shape({
```
LeandroTorresSicilia: Specifies the selected value of the Lookup

src/components/Lookup/index.js
```diff
@@ -352,44 +352,62 @@ class Lookup extends Component {
...
         label: PropTypes.string,
         description: PropTypes.string,
         icon: PropTypes.node,
     }),
+    options: PropTypes.arrayOf(
```
LeandroTorresSicilia: An array of matched options to show in a menu

src/components/Lookup/index.js
```diff
@@ -352,44 +352,62 @@ class Lookup extends Component {
...
+            description: PropTypes.string,
+            icon: PropTypes.node,
+        }),
+    ),
+    /** The name of the Lookup. */
     name: PropTypes.string,
-    label: PropTypes.oneOfType([
-        PropTypes.string, PropTypes.node,
-    ]).isRequired,
+    debounce: PropTypes.bool,
```
LeandroTorresSicilia: If set to true the onSearch event is sent when the customer finish typing

src/components/Lookup/index.js
```diff
@@ -352,44 +352,62 @@ class Lookup extends Component {
...
     disabled: PropTypes.bool,
+    /** Specifies that the Lookup is read-only. This value defaults to false. */
     readOnly: PropTypes.bool,
-    onChange: PropTypes.func,
+    /** Specifies the tab order of an element (when the tab button is used for navigating). */
     tabIndex: PropTypes.oneOfType([
         PropTypes.number,
         PropTypes.string,
     ]),
+    onSearch: PropTypes.func,
```
LeandroTorresSicilia: The action triggered for every key stroke when the customer is typing in the input. It sent the value/query of the input. This value is normally used for filter/search for more options.

PR #720, opened 2019-05-08
chore: setup prettier 

reiniergs:
- it setup prettier in the project
- it formats the entire project using prettier
- it added prettier badge into readme file


TahimiLeonBravo:
fix: #681

PR #739, opened 2019-05-16
docs: add-lookup-modal-redux-example

TahimiLeonBravo:
fix: #743 

Changes proposed in this PR:
- Add Lookup to the Modal redux example.

[x] I have followed (at least) the [PR section of the contributing guide](https://github.com/90milesbridge/react-rainbow/blob/master/CONTRIBUTING.md#submitting-a-pull-request).

@90milesbridge/tigger

### DIFF
--- a/src/components/Lookup/__test__/lookup.spec.js
+++ b/src/components/Lookup/__test__/lookup.spec.js
@@ -322,8 +322,7 @@ describe('<Lookup />', () => {
             { label: 'Miami' },
         ]);
     });
-
-    it('should not call event.stopPropagation when press esc key and the input has value typed', () => {
+    it('should call event.stopPropagation when press esc key and the input has value typed', () => {
         const stopPropagationMockFn = jest.fn();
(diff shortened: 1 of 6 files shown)


src/components/Lookup/__test__/lookup.spec.js
```diff
@@ -322,4 +322,20 @@ describe('<Lookup />', () => {
             { label: 'Miami' },
         ]);
     });
+
+    it('should not call event.stopPropagation when press esc key and the input has value typed', () => {
```
LeandroTorresSicilia: remove not, remove line and add test case for when not call stopPropagation

src/components/Lookup/index.js
```diff
@@ -122,6 +123,7 @@ class Lookup extends Component {
             searchValue: '',
         });
         onChange(value);
+        this.containerRef.current.focus();
```
LeandroTorresSicilia: set focus when change value

PR #763, opened 2019-05-28
fix: lint errors on node12

wildergd:
Changes proposed in this PR:
- fixed lint error on node12

@90milesbridge/tigger


reiniergs:
@wildergd is there a way we don't have to add the eslint disable comments? that solution is not going to scale, there should take a change in the eslint configuration. Also if we need to enforce node12 for development, we can do it by adding the constraint in the package.json file.

ps: I think we should enforce node12 since it's the lates stable node version

PR #765, opened 2019-05-29
feat: implement radiobuttongroup

wildergd:

### DIFF
--- a/src/components/RadioButtonGroup/buttonItems.js
+++ b/src/components/RadioButtonGroup/buttonItems.js
@@ -7,25 +7,39 @@ export default function ButtonItems(props) {
 
     const isChecked = option => option.value === value;
 
-    return (
-        <div className="rainbow-radio-button-group_buttons-list">
-            <div className="rainbow-radio-button-group_indicator" />
-            {options.map((option, index) => {
-                const key = `radio-${index}`;
-                return (
-                    <RadioButton
-                        key={key}
-                        required={required}
-                        onChange={onChange}
-                        isChecked={isChecked(option)}
-                        ariaDescribedby={ariaDescribedby}
-                        name={name}
-                        {...option}
-                    />
-                );
-            })}
-        </div>
-    );
+    return options.map((option, index) => {
+        const key = `radiobutton-${index}`;
+        return (
+            <RadioButton
+                key={key}
+                required={required}
+                onChange={onChange}
+                isChecked={isChecked(option)}
+                ariaDescribedby={ariaDescribedby}
+                name={name}
+                {...option}
+            />
+        );
+    });
+    // return (
+    //     <div className="rainbow-radio-button-group_buttons-list">
+    //         <div className="rainbow-radio-button-group_indicator" />
+    //         {options.map((option, index) => {
+    //             const key = `radiobutton-${index}`;
+    //             return (
+    //                 <RadioButton
+    //                     key={key}
+    //                     required={required}
+    //                     onChange={onChange}
+    //                     isChecked={isChecked(option)}
+    //                     ariaDescribedby={ariaDescribedby}
+    //                     name={name}
+    //                     {...option}
+    //                 />
+    //             );
+    //         })}
+    //     </div>
+    // );
 }
 
 ButtonItems.propTypes = {
--- a/src/components/RadioButtonGroup/buttonItems.js
+++ b/src/components/RadioButtonGroup/buttonItems.js
@@ -1,0 +1,44 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import RadioButton from './radioButton';
+
+export default function ButtonItems(props) {
+    const { options, ariaDescribedby, onChange, value, name, required } = props;
+
+    const isChecked = option => option.value === value;
+
+    return (
+        <div className="rainbow-radio-button-group_buttons-list">
+            <div className="rainbow-radio-button-group_indicator" />
+            {options.map((option, index) => {
+                const key = `radio-${index}`;
+                return (
+                    <RadioButton
+                        key={key}
+                        required={required}
+                        onChange={onChange}
+                        isChecked={isChecked(option)}
+                        ariaDescribedby={ariaDescribedby}
+                        name={name}
+                        {...option}
+                    />
+                );
+            })}
+        </div>
+    );
+}
+
+ButtonItems.propTypes = {
+    ariaDescribedby: PropTypes.string,
+    value: PropTypes.string,
+    options: PropTypes.array.isRequired,
+    onChange: PropTypes.func,
+    name: PropTypes.string.isRequired,
+    required: PropTypes.bool.isRequired,
+};
+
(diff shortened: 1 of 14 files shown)


src/components/RadioButtonGroup/buttonItems.js
```diff
@@ -0,0 +1,58 @@
...
+    //                     onChange={onChange}
+    //                     isChecked={isChecked(option)}
+    //                     ariaDescribedby={ariaDescribedby}
+    //                     name={name}
+    //                     {...option}
+    //                 />
+    //             );
+    //         })}
+    //     </div>
+    // );
```
LeandroTorresSicilia: remove this comments if are not needed

src/components/RadioButtonGroup/radioButton.js
```diff
@@ -0,0 +1,91 @@
...
+        const { isChecked, disabled } = this.props;
+        return classnames(
+            'rainbow-radio-button-group_radio',
+            {
+                'rainbow-radio-button-group_radio--checked': isChecked,
+            },
+            {
+                'rainbow-radio-button-group_radio--disabled': disabled,
+            },
+        );
```
LeandroTorresSicilia: here you can use:
```
return classnames('rainbow-radio-button-group_radio', {
    'rainbow-radio-button-group_radio--checked': isChecked,
    'rainbow-radio-button-group_radio--disabled': disabled,
});
```

src/components/RadioButtonGroup/radioButton.js
```diff
@@ -0,0 +1,91 @@
...
+            disabled,
+            isChecked,
+            name,
+            required,
+        } = this.props;
+
+        return (
+            <span className={this.getContainerClassNames()}>
+                <input
+                    ref={this.ref}
```
LeandroTorresSicilia: I think we don't need this, if so remove it

src/components/RadioButtonGroup/radioButton.js
```diff
@@ -0,0 +1,91 @@
...
+                    aria-describedby={ariaDescribedby}
+                    onChange={onChange}
+                    disabled={disabled}
+                />
+
+                <label
+                    className="rainbow-radio-button-group_radio-label-container"
+                    htmlFor={this.radioId}
+                >
+                    <RenderIf isTrue={!!label}>
```
LeandroTorresSicilia: you can remove the RenderIf since the label is required for accessibility, in RadioGroup we should remove it too

src/components/RadioButtonGroup/styles.scss
```diff
@@ -0,0 +1,146 @@
...
+.rainbow-radio-button-group_label {
+    display: inline-block;
+    color: $color-gray-4;
+    font-size: $font-size-text-medium;
+    margin-right: $margin-small;
+    margin-bottom: $margin-xx-small;
+}
+
+.rainbow-radio-button-group_inner-container {
+    // display: inline-block;
```
LeandroTorresSicilia: remove comment

src/components/RadioButtonGroup/styles.scss
```diff
@@ -0,0 +1,146 @@
...
+    color: $color-error-active;
+}
+
+.rainbow-radio-button-group_radio {
+    display: inline-block;
+}
+
+// .rainbow-radio-button-group_inner-container > .rainbow-radio-button-group_radio:first-child {
+//     border-radius: 100px 0 0 100px;
+// }
```
LeandroTorresSicilia: same ^^

src/components/RadioButtonGroup/styles.scss
```diff
@@ -0,0 +1,146 @@
...
+
+.rainbow-radio-button-group_indicator {
+    @extend %border-margin;
+
+    display: inline-block;
+    border-radius: 1.25rem;
+    // margin-top: -1px;
+    // margin-left: -1px;
+    // width: 0;
+    // height: (100% + 1px);
```
LeandroTorresSicilia: same ^^ but also review since this class is not being used

src/components/RadioButtonGroup/styles.scss
```diff
@@ -0,0 +1,146 @@
...
+//     background: #0000CC;
+// }
+
+// .rainbow-radio-button-group_radio:last-child {
+//     background: #CC0000;
+// }
+// .rainbow-radio-button-group_radio [type='radio'] + .rainbow-radio-button-group_radio-label-container {
+//     padding: $padding-x-small $padding-small;
+//     border:1px solid red;
+// }
```
LeandroTorresSicilia: same ^^

src/components/RadioButtonGroup/styles.scss
```diff
@@ -0,0 +1,146 @@
...
+    margin-right: $margin-small;
+    margin-bottom: $margin-xx-small;
+}
+
+.rainbow-radio-button-group_inner-container {
+    // display: inline-block;
+    border-radius: 1.25rem;
+    border: solid 1px #e3e5ed;
+    background-color: rgba(227, 229, 237, 0.4);
+    padding-top: calc(0.1rem + 1px);
```
TahimiLeonBravo: We should not align the content with `padding-bottom` an  `padding-top`, instead, like we already use `flex`  we should use `align-items: center` and `height: 2.5rem`

src/components/RadioButtonGroup/styles.scss
```diff
@@ -0,0 +1,146 @@
...
+    padding-bottom: 0.1rem;
+    display: -ms-inline-flexbox;
+    display: inline-flex;
+
+    .rainbow-radio-button-group_radio {
+        .rainbow-radio-button-group_radio-label-container {
+            @extend %border-margin;
+
+            border-radius: 1.25rem;
+            padding-top: 0.4rem;
```
TahimiLeonBravo: ^^same

src/components/RadioButtonGroup/styles.scss
```diff
@@ -0,0 +1,146 @@
...
+@mixin user-select() {
+    -webkit-user-select: none;
+    -moz-user-select: none;
+    -ms-user-select: none;
+    user-select: none;
+}
+
+.rainbow-radio-button-group_container {
+    position: relative;
+}
```
TahimiLeonBravo: I think that we don't need to put in `position: relative` to the container

src/components/RadioButtonGroup/styles.scss
```diff
@@ -0,0 +1,146 @@
...
+
+    .rainbow-radio-button-group_radio {
+        .rainbow-radio-button-group_radio-label-container {
+            @extend %border-margin;
+
+            border-radius: 1.25rem;
+            padding-top: 0.4rem;
+            padding-right: $padding-large;
+            padding-bottom: calc(0.5rem + 1px);
+            padding-left: $padding-large;
```
TahimiLeonBravo: it is a best practice use the `padding` shortcut when you want defined various paddings

Issue #775, opened 2019-06-01
feat: implement `icon` prop to the `Lookup` component


TahimiLeonBravo:

TahimiLeonBravo:
fix: #776

PR #776, opened 2019-06-01
feat: add searchIcon prop to lookup component

wildergd:
feat: #775 

Changes proposed in this PR:
- add `searchIcon` prop to make component more customizable

@90milesbridge/tigger

### DIFF
--- a/src/components/Lookup/index.js
+++ b/src/components/Lookup/index.js
@@ -18,6 +18,8 @@ import {
 import { uniqueId } from '../../libs/utils';
 import { UP_KEY, DOWN_KEY, ENTER_KEY, ESCAPE_KEY } from '../../libs/constants';
 import withReduxForm from '../../libs/hocs/withReduxForm';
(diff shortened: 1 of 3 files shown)


src/components/Lookup/index.js
```diff
@@ -477,6 +482,8 @@ Lookup.propTypes = {
     className: PropTypes.string,
     /** An object with custom style applied to the outer element. */
     style: PropTypes.object,
+    /** The icon to show when search is empty. */
+    searchIcon: PropTypes.object,
```
LeandroTorresSicilia: let's call it icon and put node as proptype

src/components/Lookup/index.js
```diff
@@ -500,6 +507,7 @@ Lookup.defaultProps = {
     options: undefined,
     onSearch: () => {},
     debounce: false,
+    searchIcon: undefined,
```
LeandroTorresSicilia: use the defalut icon here in index

Issue #697, opened 2019-04-24
feat: implement GAddressLookup component


TahimiLeonBravo:
### UI/UX Spec
- [x] it should look like the design https://zpl.io/bo84DdZ

### Acceptance criteria
- [x] it should show as the first option the value typing in the input. (view the design)
- [x] it should have `icon`, `label` and `description` in the Google options.
- [x] it should have `icon` and  `label` in the first option.
- [x] it should be compatible with redux-form.
- [x] it should render in the chip of the selected option `the label`,  followed by` the description` showed in the options.
- [x] it should have `the ZIP code`, `the city`, `the country` in the options description
- [x] it should hide `the powered by google image` when the `error` prop is passed

PR #787, opened 2019-06-05
feat: implement GAddressLookup component

wildergd:
feat: #697

Changes proposed in this PR:
- add GAddressLookup component

@90milesbridge/tigger


src/components/GAddressLookup/component.js
```diff
@@ -0,0 +1,271 @@
...
+import SearchValueIcon from './icons/searchValueIcon';
+import { uniqueId } from '../../libs/utils';
+import poweredByGoogleColored from '../../../assets/images/google/powered_by_google_on_white.png';
+import './styles.css';
+
+function formatSuggestionItem(suggestion, highlightMatch, icon) {
+    let formattedLabel;
+
+    if (highlightMatch) {
+        suggestion.structured_formatting.main_text_matched_substrings.forEach(match => {
```
LeandroTorresSicilia: We should use reduce instead of forEach since it is not functional, here you create a formattedLabel variable and then mutate it inside forEach, so if you use reduce there is not mutation. Also try to find another solution that does not use dangerouslySetInnerHTML since it not recommended. Also don't use React.createElement.
Try to create the markup with reduce.

src/components/GAddressLookup/component.js
```diff
@@ -0,0 +1,271 @@
...
+    handleChange(option) {
+        const { onChange } = this.props;
+
+        if (!option) {
+            this.setState({
+                value: null,
+                selectedPlace: option,
+            });
+            onChange(option);
+            return;
```
LeandroTorresSicilia: this states are not needed, only use:
```
if (!option) {
    return onChange(null);
}
```

src/components/GAddressLookup/component.js
```diff
@@ -0,0 +1,271 @@
...
+        const value = places.length > 0 ? places[suggestions.indexOf(option)] : option;
+
+        this.setState({
+            value,
+            selectedPlace: {
+                label: option.label,
+                description: option.description,
+                icon: <SelectedLocationIcon />,
+            },
+        });
```
LeandroTorresSicilia: there is not need to use this states

src/components/GAddressLookup/component.js
```diff
@@ -0,0 +1,271 @@
...
+
+        onChange(value);
```
LeandroTorresSicilia: here you only need to find the option from google that correspond with the option received here as argument and send it trough onChange

src/components/GAddressLookup/component.js
```diff
@@ -0,0 +1,271 @@
...
+    }
+
+    handleSearch(value) {
+        const { initialized } = this.state;
+        if (!initialized) return;
+
+        if (value && this.state.searchValue !== value) {
+            this.setState({ searchValue: value });
```
LeandroTorresSicilia: I think this state is not needed, since this the comparation:
```this.state.searchValue !== value``` is not needed, always you get here this values will be differents and pass value to processSearchResults and handleError as argument and then youn can remove it from state.

File: src/components/GAddressLookup/component.js [hunk 6]
```diff
@@ -0,0 +1,271 @@
...
+            this.searchGoogleMaps(value).then(
```
LeandroTorresSicilia: let's call this searchGooglePlaces

src/components/GAddressLookup/component.js
```diff
@@ -0,0 +1,271 @@
...
+                results => this.processSearchResults(results),
+                () => this.handleError(),
```
LeandroTorresSicilia: if handleError is meant to be used as promise reject then use it in catch clause as:
```
this.searchGoogleMaps(value)
    .then(results => this.processSearchResults(results, searchValue))
    .catch(this.handleError);
```

src/components/GAddressLookup/component.js
```diff
@@ -0,0 +1,271 @@
...
+                    input: needle,
+                },
+                (predictions, status) => {
+                    this.setState({ isSearching: false });
+
+                    if (status !== window.google.maps.places.PlacesServiceStatus.OK) {
+                        reject(status);
+                    }
+
+                    resolve(predictions);
```
LeandroTorresSicilia: use:
```
if (status !== window.google.maps.places.PlacesServiceStatus.OK) {
    return reject(status);
}
return resolve(predictions);
```
or esle in order to not get to resolve when reject

src/components/GAddressLookup/component.js
```diff
@@ -0,0 +1,271 @@
...
+            suggestions: [
+                {
+                    label: (
+                        <span>
+                            Search for: <b>&lsquo;{this.state.searchValue}&rsquo;</b>
+                        </span>
+                    ),
+                    icon: <SearchValueIcon />,
+                },
+            ].concat(results.map(place => formatSuggestionItem(place, true, <LocationItemIcon />))),
```
LeandroTorresSicilia: use this as:
```
this.setState({
    places: results,
    suggestions: getSuggestions(results, searchValue),
});
```
and then put getSuggestions inside helpers folder and it is more easily to unit test:
```
function getSuggestions(results, searchValue) {
    return [
                getSearchForOption(searchValue),
            ].concat(results.map(place => formatSuggestionItem(place, true, <LocationItemIcon />)))
}
```
and use getSearchForOption since it is used twice:
```
function getSearchForOption(searchValue) {
    return {
                    label: (
                        <span>
                            Search for: <b>&lsquo;{this.state.searchValue}&rsquo;</b>
                        </span>
                    ),
                    icon: <SearchValueIcon />,
                };
}
```

src/components/GAddressLookup/component.js
```diff
@@ -0,0 +1,271 @@
...
+            places: [],
+            suggestions: [
+                {
+                    label: (
+                        <span>
+                            Search for: <b>&lsquo;{this.state.searchValue}&rsquo;</b>
+                        </span>
+                    ),
+                    icon: <SearchValueIcon />,
+                },
```
LeandroTorresSicilia: use the getSearchForOption function

src/components/GAddressLookup/component.js
```diff
@@ -0,0 +1,271 @@
...
+            ],
+        });
+    }
+
+    clear() {
+        this.setState({
+            searchValue: '',
```
LeandroTorresSicilia: remove state not used

src/components/GAddressLookup/component.js
```diff
@@ -0,0 +1,271 @@
...
+        return (
+            <div id={id} style={style} className={this.getContainerClassName()}>
+                <Lookup
+                    id={this.lookupId}
+                    name={name}
+                    label={label}
+                    debounce
+                    isLoading={isSearching}
+                    placeholder={placeholder}
+                    options={suggestions.length > 0 ? suggestions : null}
```
LeandroTorresSicilia: not use js inside jsx
put in a variable below state destructuring:
```
const { isSearching, suggestions } = this.state;
const options = suggestions.length > 0 ? suggestions : null;
```

src/components/GAddressLookup/component.js
```diff
@@ -0,0 +1,271 @@
...
+                    hideLabel={hideLabel}
+                    required={required}
+                    tabIndex={tabIndex}
+                    disabled={disabled}
+                    value={this.state.selectedPlace}
```
LeandroTorresSicilia: here use:
```value={getFormattedValue(value)}```
where value comes from props and getFormattedValue return the option with the label, description and icon computed

File: src/components/GAddressLookup/component.js [hunk 14]
```diff
@@ -0,0 +1,271 @@
...
+    isScriptLoaded: PropTypes.bool.isRequired,
+    isScriptLoadSucceed: PropTypes.bool.isRequired,
+    searchOptions: PropTypes.shape({}),
+    label: PropTypes.oneOfType([PropTypes.string, PropTypes.node]).isRequired,
+    hideLabel: PropTypes.bool,
+    value: PropTypes.oneOfType([
+        PropTypes.shape({
+            label: PropTypes.string,
+            description: PropTypes.string,
+            icon: PropTypes.node,
```
LeandroTorresSicilia: change this shape to match with what google returns

src/components/GAddressLookup/index.js
```diff
@@ -0,0 +1,78 @@
...
+    hideLabel: PropTypes.bool,
+    /** Specifies the selected value of the component. */
+    value: PropTypes.oneOfType([
+        PropTypes.shape({
+            label: PropTypes.string,
+            description: PropTypes.string,
+            icon: PropTypes.node,
+        }),
+        PropTypes.string,
+    ]),
```
LeandroTorresSicilia: same ^^

src/components/GAddressLookup/index.js
```diff
@@ -0,0 +1,78 @@
...
+    /** Specifies the tab order of an element (when the tab button is used for navigating). */
+    tabIndex: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
+    /** The id of the outer element. */
+    id: PropTypes.string,
+    /** A CSS class for the outer element, in addition to the component's base classes. */
+    className: PropTypes.string,
+    /** An object with custom style applied to the outer element. */
+    style: PropTypes.object,
+    /**  The action triggered when click/select an option. */
+    onChange: PropTypes.func,
```
LeandroTorresSicilia: we should pass onClick, onFocus and onBlur

src/components/GAddressLookup/index.js
```diff
@@ -0,0 +1,78 @@
...
+
+GAddressLookup.defaultProps = {
+    apiKey: undefined,
+    value: undefined,
+    name: undefined,
+    placeholder: null,
+    required: false,
+    error: null,
+    disabled: false,
+    readOnly: false,
```
LeandroTorresSicilia: remove readOnly

src/components/GAddressLookup/readme.md
```diff
@@ -0,0 +1,32 @@
...
+        margin: 'auto',
+    };
+
+    <div className="rainbow-p-vertical_large rainbow-p-horizontal_xx-large rainbow-m-horizontal_xx-large" style={styles}>
+        <GAddressLookup
+            id="gaddresslookup-2"
+            label="GAddressLookup label"
+            placeholder="Enter location"
+            apiKey={LIBRARY_GOOGLE_MAPS_APIKEY}
+            disabled={true}
```
LeandroTorresSicilia: use disabled

src/components/GAddressLookup/index.js
```diff
@@ -0,0 +1,78 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import scriptLoader from 'react-async-script-loader';
+import PlacesLookupComponent from './component';
+/**
+ * The GAddressLookup component is used to find a location.
```
TahimiLeonBravo: Let's call this component `GoogleAddressLookup` instead `GAddressLookup`

src/components/GAddressLookup/readme.md
```diff
@@ -0,0 +1,32 @@
...
+        />
+    </div>
```
TahimiLeonBravo: I think we should add an example with Redux-Form and an example with an error.

src/components/GAddressLookup/component.js
```diff
@@ -0,0 +1,271 @@
...
+    }
+
+    handleError() {
+        this.setState({
+            places: [],
+            suggestions: [
+                {
+                    label: (
+                        <span>
+                            Search for: <b>&lsquo;{this.state.searchValue}&rsquo;</b>
```
TahimiLeonBravo: The `Search for` text should have a different style than the text inside the quotation marks. Please follow the design.

src/components/GAddressLookup/component.js
```diff
@@ -0,0 +1,271 @@
...
+                    tabIndex={tabIndex}
+                    disabled={disabled}
+                    value={this.state.selectedPlace}
+                    onChange={this.handleChange}
+                    onSearch={this.handleSearch}
+                    error={error}
+                    icon={<LocationIcon />}
+                />
+                <div className="rainbow-google-address-poweredby_container">
+                    <img src={poweredByGoogleColored} alt="powered by Google" />
```
TahimiLeonBravo: The `Powered by Google` image should be aligned to the right. Please review the design. 
Screen Shot 2019-06-06 at 12 44 09 AM

It should be hidden when an error is showed.
Screen Shot 2019-06-06 at 1 57 57 AM

src/components/GoogleAddressLookup/styles.scss
```diff
@@ -0,0 +1,36 @@
...
+        display: -ms-flexbox;
+        display: flex;
+        -ms-flex-pack: end;
+        justify-content: end;
+        -ms-flex-align: end;
+        align-items: flex-end;
+        padding-top: $padding-xx-small;
+        padding-right: $padding-x-small;
+
+        img {
```
TahimiLeonBravo: We should not apply the style to the tag names, we should use classes instead. Let's follow the principle of keeping things decoupled to ensure that our components are scalable and thus prevent any change in the future from breaking them.

src/components/GoogleAddressLookup/styles.scss
```diff
@@ -0,0 +1,36 @@
...
+            height: 0.625rem;
+        }
+    }
+
+    .rainbow-google-address-lookup_search-option-label {
+        color: #a4a7b5;
+
+        b {
```
TahimiLeonBravo: We should not apply the style to the tag names, we should use classes instead. Let's follow the principle of keeping things decoupled to ensure that our components are scalable and thus prevent any change in the future from breaking them.

src/components/GoogleAddressLookup/component.js
```diff
@@ -0,0 +1,293 @@
...
+                    onChange={this.handleChange}
+                    onSearch={this.handleSearch}
+                    onClick={onClick}
+                    onFocus={onFocus}
+                    onBlur={onBlur}
+                    error={error}
+                    icon={<LocationIcon />}
+                />
+                {!error && (
+                    <div className="rainbow-google-address-poweredby_container">
```
TahimiLeonBravo: Let's follow the class name naming convention. Call this `rainbow-google-address-lookup_powered-by-google-container`

src/components/GoogleAddressLookup/__test__/component.spec.js
```diff
@@ -0,0 +1,51 @@
...
+
+global.google = {
+    maps: {
+        places: {
+            AutocompleteService: jest.fn(() => ({
+                getPlacePredictions: jest.fn(),
+            })),
+            // PlacesServiceStatus: jest.{
+            //     OK: 'OK',
+            // },
```
LeandroTorresSicilia: remove this comments if not used, but if you want to mock PlacesServiceStatus then:
```
PlacesServiceStatus: {
    OK: 'OK',
},
```

src/components/GoogleAddressLookup/__test__/format.spec.js
```diff
@@ -0,0 +1,44 @@
+import React from 'react';
```
LeandroTorresSicilia: this tests should be placed inside helpers folder, so inside helpers you create another __test__ folder and then there is where the helpers test should be placed

src/components/GoogleAddressLookup/__test__/format.spec.js
```diff
@@ -0,0 +1,44 @@
...
+        { offset: 46, value: 'Karnataka' },
+        { offset: 57, value: 'India' },
+    ],
+    types: ['park', 'point_of_interest', 'establishment'],
+};
+
+describe('formatSuggestionItem', () => {
+    it('should return the item formated without highligth', () => {
+        expect(getFormattedValue(suggestion, false, <SelectedLocationIcon />)).toEqual({
+            id: 'ChIJL2fQ53MWrjsRuN9D6aalLMY',
```
LeandroTorresSicilia: why use place_id instead of id?

src/components/GoogleAddressLookup/__test__/format.spec.js
```diff
@@ -0,0 +1,44 @@
...
+    });
+
+    // it('should return the item formated with hightligh', () => {
+    //     expect(getFormattedValue(suggestion, true, <SelectedLocationIcon />)).toEqual({
+    //         id: 'ChIJL2fQ53MWrjsRuN9D6aalLMY',
+    //         label: <span><b>Cub</b>bon Park</span>,
+    //         description: 'Sampangi Rama Nagara, Bengaluru, Karnataka, India',
+    //         icon: <SelectedLocationIcon />,
+    //     });
+    // });
```
LeandroTorresSicilia: try to fix it, if not remove it and in another PR we add the tests

src/components/GoogleAddressLookup/component.js
```diff
@@ -0,0 +1,297 @@
...
+class PlacesLookupComponent extends Component {
+    constructor(props) {
+        super(props);
+        this.lookupId = uniqueId('gaddrlookup-input');
+        this.initialized = false;
+        this.handleChange = this.handleChange.bind(this);
+        this.handleSearch = this.handleSearch.bind(this);
+
+        this.state = {
+            initialized: false,
```
LeandroTorresSicilia: remove from here since we will not use it any more

src/components/GoogleAddressLookup/component.js
```diff
@@ -0,0 +1,297 @@
...
+            value = option.data;
+        }
+
+        return onChange(value);
+    }
+
+    handleSearch(value) {
+        if (!this.initialized) return;
+
+        if (value && this.state.searchValue !== value) {
```
LeandroTorresSicilia: only use if (value):
```this.state.searchValue !== value```
this is not needed

File: src/components/GoogleAddressLookup/component.js [hunk 32]
```diff
@@ -0,0 +1,297 @@
...
+            this.searchGooglePlaces(value)
+                .then(results => this.processSearchResults(results, value))
+                .catch(this.handleError(value));
```
LeandroTorresSicilia: use in this way:
```.catch(() => this.handleError(value));```
because here you need to pass a callback that will be called later and here you are calling it right away

File: src/components/GoogleAddressLookup/component.js [hunk 33]
```diff
@@ -0,0 +1,297 @@
...
+        } else if (!value) {
+            this.clear();
+        }
+    }
+
+    searchGooglePlaces(needle) {
+        const { searchOptions } = this.props;
+        let searchParams = {};
+        if (searchOptions) {
```
LeandroTorresSicilia: all this is another helper function `getSearchParams ` and then here only put:
```const searchParams = getSearchParams(searchOptions);```

File: src/components/GoogleAddressLookup/component.js [hunk 34]
```diff
@@ -0,0 +1,297 @@
...
+                <Lookup
+                    id={this.lookupId}
+                    name={name}
+                    label={label}
+                    debounce
+                    isLoading={isSearching}
+                    placeholder={placeholder}
+                    options={options}
+                    hideLabel={hideLabel}
+                    readOnly={false}
```
LeandroTorresSicilia: use:
```readOnly={readOnly}```
and add readOnly to index.js prop types in order to the user can use this prop

File: src/components/GoogleAddressLookup/component.js [hunk 35]
```diff
@@ -0,0 +1,297 @@
...
+                    onFocus={onFocus}
+                    onBlur={onBlur}
+                    error={error}
+                    icon={<LocationIcon />}
+                />
+                {!error && (
+                    <div className="rainbow-google-address-lookup_poweredby-container">
+                        <PoweredByGoogleLogo className="rainbow-google-address-lookup_poweredby-logo" />
+                    </div>
+                )}
```
LeandroTorresSicilia: use RenderIf as:
```
<RenderIf isTrue={!error}>
    <div className="rainbow-google-address-lookup_poweredby-container">
         <PoweredByGoogleLogo className="rainbow-google-address-lookup_poweredby-logo" />
    </div>
</RenderIf>
```

src/components/GoogleAddressLookup/component.js
```diff
@@ -0,0 +1,297 @@
...
+            sw: PropTypes.shape({
+                latitude: PropTypes.number.isRequired,
+                longitude: PropTypes.number.isRequired,
+            }),
+            ne: PropTypes.shape({
+                latitude: PropTypes.number.isRequired,
+                longitude: PropTypes.number.isRequired,
+            }),
+        }),
+        country: PropTypes.string,
```
LeandroTorresSicilia: this can be an array of up to five country code strings too
we need to explain this in proptype description in index.js and mention example of use

src/components/GoogleAddressLookup/component.js
```diff
@@ -0,0 +1,297 @@
...
+            terms: PropTypes.arrayOf(
+                PropTypes.shape({
+                    offset: PropTypes.number,
+                    value: PropTypes.string,
+                }),
+            ),
+            types: PropTypes.arrayOf(PropTypes.string),
+        }),
+        PropTypes.string,
+    ]),
```
LeandroTorresSicilia: we can make a proptypes.js file where we can export value and searchOptions prop types since are to large and used two times, and here use something like:
```
value: valueShape,
searchOptions: searchOptionsShape,

```

src/components/GoogleAddressLookup/index.js
```diff
@@ -0,0 +1,186 @@
...
+            terms: PropTypes.arrayOf(
+                PropTypes.shape({
+                    offset: PropTypes.number,
+                    value: PropTypes.string,
+                }),
+            ),
+            /**
+             * An array of types that the prediction belongs to,
+             * for example 'establishment' or 'geocode'.
+             */
```
LeandroTorresSicilia: very good this sub descriptions
this is what will be stored in proptypes.js
and used here and in component.js

src/components/GoogleAddressLookup/__test__/format.spec.js
wildergd: id property is marked as deprecated in google api.. in favor of place_id... so i decided to use place_id in order to avoid issues in the future...

src/components/GoogleAddressLookup/__test__/format.spec.js
wildergd: this test works ... but as we tried to use jsx and avoid the use dangerousSetInnerHTML... there are several characters that React generates that cause differences between expected result and the result itself... so it is better to remove it...

src/components/GoogleAddressLookup/component.js
wildergd: we need to check this... in the api reference i have the country parameter is just an string... maybe the new API changed

src/components/GoogleAddressLookup/__test__/format.spec.js
LeandroTorresSicilia: @wildergd 👍

PR #804, opened 2019-06-11
test: add test for some helper functions in GoogleAddressLookup component

wildergd:
test: 
Changes proposed in this PR:
- add test for some helper functions in GoogleAddressLookup component

@90milesbridge/tigger


src/components/GoogleAddressLookup/helpers/__test__/getSearchParams.spec.js
```diff
@@ -0,0 +1,61 @@
...
+        country: ['us', 'ca'],
+    },
+];
+
+describe('getSearchParams', () => {
+    it('should convert searchOptions into google.maps.places.AutocompleteRequest search options', () => {
+        expect(getSearchParams(searchOptions[0])).toEqual({
+            bounds: global.google.maps.LatLngBounds(
+                global.google.maps.LatLng(-63.941264, 151.2042969),
+                global.google.maps.LatLng(63.941264, -151.2042969),
```
LeandroTorresSicilia: as you mock this then it is returning an empty object so this is not right, so you need to mock this function to return something based on arg passed and then check that this values are equal to what it returned in the mock

src/components/GoogleAddressLookup/helpers/__test__/getFormattedValue.spec.js
```diff
@@ -0,0 +1,56 @@
...
+        };
+
+        expect(result).toEqual({
+            id: 'ChIJL2fQ53MWrjsRuN9D6aalLMY',
+            label: '<span><b>Cub</b>bon <b>Cub</b>bon Park</span>',
+            description: 'Sampangi Rama Nagara, Bengaluru, Karnataka, India',
+            icon: <SelectedLocationIcon />,
+        });
+    });
+});
```
LeandroTorresSicilia: add tests cases for when:
1. when value is not passed
2. when value is a string

src/components/GoogleAddressLookup/helpers/__test__/getSearchParams.spec.js
LeandroTorresSicilia: as you mock this then it is returning an empty object so this is not right, so you need to mock this function to return something based on arg passed and then check that this values are equal to what it returned in the mock

src/components/GoogleAddressLookup/helpers/__test__/getSearchParams.spec.js
```diff
@@ -0,0 +1,61 @@
...
+            },
+            radius: 150000,
+            types: ['address'],
+        });
+        expect(getSearchParams(searchOptions[2])).toEqual({
+            componentRestrictions: {
+                country: ['us', 'ca'],
+            },
+        });
+    });
```
LeandroTorresSicilia: put each of this expect inside a it define the specific searchOption object in each test like:
```
it('should ...', () => {
    const searchOptions = {
        bounds: {
            sw: {
                latitude: -63.941264,
                longitude: 151.2042969,
            },
            ne: {
                latitude: 63.941264,
                longitude: -151.2042969,
            },
        },
        types: ['address'],
    };
    expect(getSearchParams(searchOptions)).toEqual({
            bounds: global.google.maps.LatLngBounds(
                global.google.maps.LatLng(-63.941264, 151.2042969),
                global.google.maps.LatLng(63.941264, -151.2042969),
            ),
            types: ['address'],
        });
})
```
and so on for each test

src/components/GoogleAddressLookup/helpers/__test__/getSuggestions.spec.js
```diff
@@ -0,0 +1,135 @@
+import React from 'react';
+import ReactDOMServer from 'react-dom/server';
+import getSuggestions from '../getSuggestions';
+import LocationItemIcon from '../../icons/locationItemIcon';
+import SearchValueIcon from '../../icons/searchValueIcon';
+
```
LeandroTorresSicilia: here you need to mock getSearchForOption and getFormattedValue and then test that what it is returned by getSearchForOption is concatenated with what it is returned by the map result of calling getFormattedValue for each result.

src/components/GoogleAddressLookup/__test__/component.spec.js
```diff
@@ -2,28 +2,45 @@ import React from 'react';
...
-            })),
+// global.google = {
+//     maps: {
+//         places: {
+//             AutocompleteService: jest.fn(() => ({
+//                 getPlacePredictions: jest.fn(),
+//             })),
+//         },
+//     },
+// };
```
LeandroTorresSicilia: remove this comments

src/components/GoogleAddressLookup/__test__/component.spec.js
```diff
@@ -32,17 +49,34 @@ describe('<PlacesLookupComponent/>', () => {
...
+        const component = mount(<PlacesLookupComponent />);
+        component.setProps(nextProps);
+        expect(global.google.maps.places.AutocompleteService).toHaveBeenCalled();
+    });
+
+    it('should be intialized after isScriptLoaded and isScriptLoadSucceed are set to true', () => {
+        const component = mount(<PlacesLookupComponent />);
+        expect(component.instance().fieldRef.current.initialized).toBe(false);
+        component.setProps(nextProps);
+        expect(component.instance().fieldRef.current.initialized).toBe(true);
```
LeandroTorresSicilia: here you need to access fieldRef.current because withReduxForm so it is better to export the component itself too in order to test it directly as:
```
export class PlacesLookupComponent extends Component {
     //...
}
export default withReduxForm(PlacesLookupComponent);
```
and the in test import the component as:
```
import { PlacesLookupComponent } from '../component';
```
so the expect should be:
```expect(component.instance().initialized).toBe(true);```

Issue #790, opened 2019-06-07
fix: fix Issues Focus when the Lookup has an option selected


TahimiLeonBravo:
## Acceptance criteria
- [x] it should be accessible when the user press `tab` key 
- [x] it should put focus the input when the user clicking in the label

note: follow the accessibility criteria defined be salesforce to the `lookup` component
https://www.lightningdesignsystem.com/components/lookups/#site-main-content

PR #830, opened 2019-07-09
fix: issues with focus on lookup component

wildergd:
fix: #790

Changes proposed in this PR:
- fix issues with focus on Lookup component when pressing TAB
- fix issues with focus on Lookup component when clicking on Chip label

@90milesbridge/tigger

### DIFF
--- a/src/components/Chip/index.js
+++ b/src/components/Chip/index.js
@@ -29,7 +29,6 @@ export default function Chip(props) {
                 <ButtonIcon
                     className="rainbow-chip_button-container"
                     icon={<CloseIcon />}
-                    tabIndex={-1}
                     size="small"
                     title="Close"
                     onClick={onDelete}
--- a/src/components/Chip/index.js
+++ b/src/components/Chip/index.js
@@ -29,6 +29,7 @@ export default function Chip(props) {
                 <ButtonIcon
                     className="rainbow-chip_button-container"
                     icon={<CloseIcon />}
+                    tabIndex={-1}
                     size="small"
                     title="Close"
                     onClick={onDelete}
--- a/src/components/Lookup/index.js
+++ b/src/components/Lookup/index.js
@@ -3,12 +3,10 @@ import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
 import RenderIf from '../RenderIf';
-import Chip from '../Chip';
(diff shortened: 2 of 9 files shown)


src/components/Chip/index.js
```diff
@@ -29,6 +29,7 @@ export default function Chip(props) {
                 <ButtonIcon
                     className="rainbow-chip_button-container"
                     icon={<CloseIcon />}
+                    tabIndex={-1}
```
LeandroTorresSicilia: Remove this since this make the chip close button not focusable and it must be focusable.
Look at:
https://www.lightningdesignsystem.com/components/lookups/#site-main-content
in order to view the right behavior

src/components/Lookup/index.js
```diff
@@ -69,6 +71,13 @@ class Lookup extends Component {
         };
     }
 
+    componentDidMount() {
+        if (window) {
+            window.addEventListener('click', this.handleWindowClick, false);
+            window.addEventListener('touchstart', this.handleWindowClick, false);
+        }
+    }
```
LeandroTorresSicilia: we should not use window listeners inside compoents
also this will be not needed because we must use an input instead of the chip

src/components/Lookup/index.js
```diff
@@ -81,6 +90,16 @@ class Lookup extends Component {
...
     }
 
+    componentWillUnmount() {
+        window.removeEventListener('click', this.handleWindowClick, false);
+        window.removeEventListener('touchstart', this.handleWindowClick, false);
+    }
+
+    // eslint-disable-next-line class-methods-use-this
+    getChipInputClassNames() {
+        return classnames('rainbow-lookup_input', 'rainbow-lookup_input-hidden-caret');
```
LeandroTorresSicilia: if you use fixed className instead put directly in tag element

src/components/Lookup/readme.md
```diff
@@ -47,7 +47,7 @@
         }
     }
 
-    initialState = { options: null };
+    initialState = { option: { label: 'Paris' }, options: null };
```
LeandroTorresSicilia: the first example has to be the most simplest, this make the lookup to have an option selected by default, then make a new example with this and name it :
Lookup with initial option selected
Also put this new example at fourth position because now there is integrations tests based on the first 3 examples and this break tests

PR #836, opened 2019-07-15
fix: lookup accessible

LeandroTorresSicilia:
fix: #790 

Changes proposed in this PR:
-
-
-

@90milesbridge/tigger


src/components/Lookup/icons/closeIconDark.js
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export default function CloseIcon(props) {
+    const { className, style } = props;
+    return (
+        <svg
+            className={className}
+            style={style}
+            fill="#576574"
```
LeandroTorresSicilia: as this is exact the same icon let's use only one icon (closeIcon) and pass a color prop with '#01b6f5' as default then when a value is selected use:
```<CloseIcon color="#576574" />

File: src/components/Lookup/index.js [hunk 2]
```diff
@@ -92,16 +93,20 @@ class Lookup extends Component {
     }
 
     getInputClassNames() {
-        const { isLoading } = this.props;
+        const { value, isLoading } = this.props;
         return classnames('rainbow-lookup_input', {
             'rainbow-lookup_input--loading': isLoading,
+            'rainbow-lookup_input--value': !!value,
+            'rainbow-lookup_input--value-w-icon': !!value && !!value.icon,
```
LeandroTorresSicilia: use:
```'rainbow-lookup_input--value-with-icon'```

File: src/components/Lookup/styles.scss [hunk 3]
```diff
@@ -59,6 +59,31 @@
         }
     }
 
+    .rainbow-lookup_input-left-icon {
+        fill: $color-gray-3;
+        height: 100%;
+        // width: 22px;
```
LeandroTorresSicilia: remove comment

Issue #828, opened 2019-07-08
styles: add animation to the `RadioButtonGroup` component


TahimiLeonBravo:
it should have animation when the option selected is changed. (use this animation https://www.appointy.com/pricing/)

PR #839, opened 2019-07-18
fix: animate radiobuttongroup

wildergd:
fix: #828

Changes proposed in this PR:
- add animation when selecting an item

@90milesbridge/tigger

### DIFF
--- a/src/components/RadioButtonGroup/index.js
+++ b/src/components/RadioButtonGroup/index.js
@@ -30,7 +30,9 @@ class RadioButtonGroup extends Component {
     }
 
     componentDidMount() {
-        this.updateMarker();
+        setTimeout(() => {
+            this.updateMarker();
+        }, 0);
     }
 
     getVariantClassNames() {
@@ -71,12 +73,16 @@ class RadioButtonGroup extends Component {
     }
 
     addRefsToOptions(options) {
-        return options.reduce((list, option, index) => {
-            const enhancedOption = Object.assign(option, {
-                optionRef: this.optionsRefs[index],
-            });
-            return [...list, enhancedOption];
-        }, []);
+        return options.reduce(
+            (list, option, index) => [
+                ...list,
+                {
+                    optionRef: this.optionsRefs[index],
+                    ...option,
+                },
+            ],
+            [],
+        );
     }
 
     isMarkerActive() {
@@ -99,7 +105,10 @@ class RadioButtonGroup extends Component {
         if (activeOptionRef !== null) {
             this.setState({
                 markerLeft: activeOptionRef.current.offsetLeft,
-                markerWidth: activeOptionRef.current.clientWidth,
+                markerWidth: Math.max(
+                    activeOptionRef.current.offsetWidth,
+                    activeOptionRef.current.clientWidth,
+                ),
             });
         }
     }
@@ -118,15 +127,13 @@ class RadioButtonGroup extends Component {
                     </legend>
                 </RenderIf>
                 <div className="rainbow-radio-button-group_inner-container">
-                    <div className="rainbow-radio-button-group_marker">
-                        <Marker
-                            isVisible={this.isMarkerActive()}
-                            style={{
-                                left: markerLeft,
-                                width: markerWidth,
-                            }}
-                        />
-                    </div>
+                    <Marker
+                        isVisible={this.isMarkerActive()}
+                        style={{
+                            left: markerLeft,
+                            width: markerWidth,
+                        }}
+                    />
                     <div className="rainbow-radio-button-group_items-container">
                         <ButtonItems
                             value={value}
--- a/src/components/RadioButtonGroup/index.js
+++ b/src/components/RadioButtonGroup/index.js
@@ -78,13 +78,10 @@ class RadioButtonGroup extends Component {
     }
 
     addRefsToOptions(options) {
-        return options.map(
-            (option, index) => ({
-                optionRef: this.optionsRefs[index],
-                ...option,
-            }),
-            this,
-        );
+        return options.map((option, index) => ({
+            optionRef: this.optionsRefs[index],
+            ...option,
+        }));
     }
 
     isMarkerActive() {
--- a/src/components/RadioButtonGroup/index.js
+++ b/src/components/RadioButtonGroup/index.js
@@ -79,8 +79,8 @@ class RadioButtonGroup extends Component {
 
     addRefsToOptions(options) {
         return options.map((option, index) => ({
-            optionRef: this.optionsRefs[index],
             ...option,
+            optionRef: this.optionsRefs[index],
         }));
     }
 
--- a/src/components/RadioButtonGroup/index.js
+++ b/src/components/RadioButtonGroup/index.js
@@ -7,6 +7,8 @@ import RenderIf from '../RenderIf';
 import RequiredAsterisk from '../RequiredAsterisk';
 import { uniqueId } from '../../libs/utils';
 import './styles.css';
+import Marker from './marker';
+import isOptionChecked from './helpers/isOptionChecked';
 
 /**
  * A button list that can have a single entry checked at any one time.
@@ -17,6 +19,18 @@ class RadioButtonGroup extends Component {
         super(props);
         this.errorId = uniqueId('error-message');
         this.groupNameId = props.name || uniqueId('options');
+        this.handleChange = this.handleChange.bind(this);
+        this.isMarkerActive = this.isMarkerActive.bind(this);
+        this.optionsRefs = this.generateRefsForOptions();
+
+        this.state = {
+            markerLeft: 0,
+            markerWidth: 0,
+        };
+    }
+
+    componentDidMount() {
+        this.updateMarker();
     }
 
     getVariantClassNames() {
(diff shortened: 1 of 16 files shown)


src/components/RadioButtonGroup/index.js
```diff
@@ -17,6 +19,20 @@ class RadioButtonGroup extends Component {
         super(props);
         this.errorId = uniqueId('error-message');
         this.groupNameId = props.name || uniqueId('options');
+        this.handleChange = this.handleChange.bind(this);
+        this.isMarkerActive = this.isMarkerActive.bind(this);
```
LeandroTorresSicilia: there is not need to bind this.isMarkerActive since it not called later as callback

src/components/RadioButtonGroup/index.js
```diff
@@ -17,6 +19,20 @@ class RadioButtonGroup extends Component {
...
+        this.state = {
+            markerLeft: 0,
+            markerWidth: 0,
+        };
+    }
+
+    componentDidMount() {
+        setTimeout(() => {
+            this.updateMarker();
+        }, 0);
```
LeandroTorresSicilia: this is not needed:
you can simply use: ```this.updateMarker();```
remember we need to avoid setTimeout

src/components/RadioButtonGroup/index.js
```diff
@@ -45,8 +61,62 @@ class RadioButtonGroup extends Component {
         return undefined;
     }
 
+    getCheckedOptionRef() {
+        const { value, options } = this.props;
+        const currentOptionIndex = options.findIndex(option => isOptionChecked(option, value));
+        return currentOptionIndex !== -1 ? this.optionsRefs[currentOptionIndex] : null;
+    }
```
LeandroTorresSicilia: move this to a helper and add test to it

src/components/RadioButtonGroup/index.js
```diff
@@ -45,8 +61,62 @@ class RadioButtonGroup extends Component {
...
+
+    generateRefsForOptions() {
+        const { options } = this.props;
+        return options.reduce(list => [...list, React.createRef()], []);
```
LeandroTorresSicilia: you can simply use:
```return options.map(() => React.createRef());```

File: src/components/RadioButtonGroup/index.js [hunk 5]
```diff
@@ -45,8 +61,62 @@ class RadioButtonGroup extends Component {
...
+            (list, option, index) => [
+                ...list,
+                {
+                    optionRef: this.optionsRefs[index],
+                    ...option,
+                },
+            ],
+            [],
+        );
+    }
```
LeandroTorresSicilia: here you can use:
```
return options.map(option => ({
            ...option,
            optionRef: React.createRef(),
        }));
```

src/components/RadioButtonGroup/index.js
```diff
@@ -45,8 +61,62 @@ class RadioButtonGroup extends Component {
...
+
+    isMarkerActive() {
+        const { value, options } = this.props;
+        const reducer = (accumulator, option) => accumulator || isOptionChecked(option, value);
+
+        return options.reduce(reducer, false);
```
LeandroTorresSicilia: what decide if the marker is visible is if there is a valid value passed so you can:
```
isMarkerActive() {
        const { value, options } = this.props;
        return options.some(option => option.value === value);
    }
```

src/components/RadioButtonGroup/index.js
```diff
@@ -45,8 +61,62 @@ class RadioButtonGroup extends Component {
...
+    }
+
+    handleChange(event) {
+        const { onChange } = this.props;
+        onChange(event);
+        setTimeout(() => {
+            this.updateMarker();
+        }, 0);
```
LeandroTorresSicilia: handleChange is not needed, pass onChange direct to ButtonItems,
instead of update the marker here let's update it in componentDidUpdate when value change:
```
        if (prevProps.value !== this.props.value) {
            this.updateMarker();
        }
    }
```

src/components/RadioButtonGroup/index.js
```diff
@@ -45,8 +61,62 @@ class RadioButtonGroup extends Component {
...
+                    activeOptionRef.current.clientWidth,
+                ),
+            });
+        }
+    }
+
     render() {
-        const { style, label, required, error, onChange, options, value, id } = this.props;
+        const { style, label, required, options, error, value, id } = this.props;
+        const optionsWithRefs = this.addRefsToOptions(options);
```
LeandroTorresSicilia: this function make an iteration and is called every time the component rerender, so this need to be computed only one time and when the options are changed dynamically(check and recompute in componentDidUpdate when prev options are different from current options).

src/components/RadioButtonGroup/index.js
```diff
@@ -57,14 +127,23 @@ class RadioButtonGroup extends Component {
...
-                        options={options}
-                        name={this.groupNameId}
-                        required={required}
-                        ariaDescribedby={this.getErrorMessageId()}
+                    <Marker
+                        isVisible={this.isMarkerActive()}
+                        style={{
+                            left: markerLeft,
+                            width: markerWidth,
+                        }}
```
LeandroTorresSicilia: avoid to mix js inside jsx, instead use a const markerStyles:
```
<Marker
                        isVisible={this.isMarkerActive()}
                        style={markerStyles}
/>
```

src/components/RadioButtonGroup/marker.js
```diff
@@ -0,0 +1,22 @@
...
+import PropTypes from 'prop-types';
+
+export default function Marker(props) {
+    const { style, isVisible } = props;
+
+    return isVisible ? (
+        <div className="rainbow-radio-button-group_marker-container">
+            <span className="rainbow-radio-button-group_marker" style={style} />
+        </div>
+    ) : null;
```
LeandroTorresSicilia: use a RenderIf

src/components/RadioButtonGroup/radioButton.js
```diff
@@ -70,6 +71,7 @@ RadioButton.propTypes = {
     isChecked: PropTypes.bool,
     name: PropTypes.string,
     required: PropTypes.bool.isRequired,
+    itemRef: PropTypes.any,
```
LeandroTorresSicilia: use:
```itemRef: PropTypes.object.isRequired,```

File: src/components/RadioButtonGroup/index.js (hunk 2 above)
Comment (wildergd) [reply src/components/RadioButtonGroup/index.js]: Had to set the settimeout cos atthis point dom has not set the styles entirely and the values of left and witdh are wrong.. it happends on chrome and safari

File: src/components/RadioButtonGroup/index.js (hunk 6 above)
Comment (wildergd) [reply src/components/RadioButtonGroup/index.js]: !option.disabled validation is also required cos a radiobuttongroup where all items are disabled should not display marker

File: src/components/RadioButtonGroup/index.js (hunk 3 above)
Comment (wildergd) [reply src/components/RadioButtonGroup/index.js]: this depends on component props and internal attributes...

File: src/components/RadioButtonGroup/index.js (hunk 5 above)
Comment (wildergd) [reply src/components/RadioButtonGroup/index.js]: the point of this function is to assign a previusly created reference to an option... createRef here will just create a new reference that we do not know.... the idea of this is that you have to different lists one with the options and one with the references...  remember that when value change we have no control over it.... and the value component receive is a value without ref... so we have to search for that value on the original options list and then use it index to get the ref on the refs list... so if we use this code you suggest... we are just reassigning a new useless ref to the options list ..

Comment (LeandroTorresSicilia) [reply src/components/RadioButtonGroup/index.js]: You are right but I mean not to use reduce

File: src/components/RadioButtonGroup/index.js [hunk 12]
```diff
@@ -45,8 +66,49 @@ class RadioButtonGroup extends Component {
...
+
+    generateRefsForOptions() {
+        const { options } = this.props;
+        return options.map(() => React.createRef());
+    }
+
+    addRefsToOptions(options) {
+        return options.map((option, index) => ({
+            optionRef: this.optionsRefs[index],
+            ...option,
```
LeandroTorresSicilia: this need to be in this order:
```
...option,
optionRef: this.optionsRefs[index],
```
if you put optionRef first, then the consumer can pass an optionRef ref prop that will cancel the one we want inside and then will break the component

src/components/RadioButtonGroup/radioButton.js
```diff
@@ -70,6 +71,7 @@ RadioButton.propTypes = {
     isChecked: PropTypes.bool,
     name: PropTypes.string,
     required: PropTypes.bool.isRequired,
+    itemRef: PropTypes.any.isRequired,
```
LeandroTorresSicilia: use object as proptype not any

src/components/RadioButtonGroup/radioButton.js
```diff
@@ -79,4 +81,5 @@ RadioButton.defaultProps = {
     disabled: false,
     isChecked: false,
     name: undefined,
+    itemRef: React.createRef(),
```
LeandroTorresSicilia: remove this since itemRef is required

Issue #842, opened 2019-07-21
fix: fix focus Lookup when `readOnly`


TahimiLeonBravo:
## Requirements
- [x] The `Lookup` should not have the focus when is `readOnly`
- [x] The `Lookup` should not have the focus when is `disabled`
- [x] Add a new interactive example 
<img width="836" alt="Screen Shot 2019-07-21 at 8 06 19 PM" src="https://user-images.githubusercontent.com/32083556/61603743-607a8680-ac04-11e9-883e-0e6fa6771b94.png">
https://zpl.io/aw4Algj

### Tests
- [x] should set the right class names when readOnly is passed
- [x] should not render the close button when readOnly is passed
- [x] should not render the close button when disabled is passed
- [x] should fire an event when the close button is clicked

PR #843, opened 2019-07-21
fix: fix lookup focus when `readOnly`

wildergd:
fix: #842 

Changes proposed in this PR:
- fix lookup styles when 'readOnly'

@90milesbridge/tigger

### DIFF
--- a/src/components/Lookup/__test__/selectedValue.spec.js
+++ b/src/components/Lookup/__test__/selectedValue.spec.js
@@ -35,16 +35,14 @@ describe('<SelectedValue />', () => {
     });
     it('should set the right class names when readOnly is passed', () => {
         const component = mount(<SelectedValue readOnly />);
-        expect(component.find('div.rainbow-lookup_selected-value--readonly').exists()).toEqual(
-            true,
-        );
+        expect(component.find('div.rainbow-lookup_selected-value--readonly').exists()).toBe(true);
     });
     it('should not render the close button when readOnly is passed', () => {
         const component = mount(<SelectedValue readOnly />);
-        expect(component.find('ButtonIcon').exists()).toEqual(false);
+        expect(component.find('ButtonIcon').exists()).toBe(false);
     });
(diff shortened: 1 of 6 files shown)


src/components/Lookup/styles.scss
```diff
@@ -317,3 +317,180 @@
...
+            &:active {
+                outline: 0;
+                box-shadow: none;
+                border: 2px transparent solid;
+            }
+        }
+    }
+
+    .rainbow-lookup_selected-value-input-with-icon {
+        &[readonly] {
```
TahimiLeonBravo: You don't need to put the styles to the [readonly] selector because this input always is readOnly

src/components/Lookup/styles.scss
```diff
@@ -317,3 +317,180 @@
...
+
+            &:focus,
+            &:active {
+                padding-left: 1.75rem;
+            }
+        }
+    }
+}
+
+.rainbow-lookup_selected-value--disabled {
```
TahimiLeonBravo: I think that this className is not necessary, you only need use the `[disabled]` selector in the `rainbow-lookup_selected-value` className

src/components/Lookup/__test__/selectedValue.spec.js
```diff
@@ -27,10 +27,24 @@ describe('<SelectedValue />', () => {
...
+            true,
+        );
+    });
+    it('should not render the close button when readOnly is passed', () => {
+        const component = mount(<SelectedValue readOnly />);
+        expect(component.find('ButtonIcon').exists()).toEqual(false);
+    });
+    it('should not render the close button when disabled is passed', () => {
+        const component = mount(<SelectedValue disabled />);
+        expect(component.find('ButtonIcon').exists()).toEqual(false);
```
LeandroTorresSicilia: use toBe instead of toEqual
toEqual is for array and objects equality check
in order to compare strings, numbers and booleans use toBe

Issue #584, opened 2019-01-29
feat: implement VisualPicker component


TahimiLeonBravo:
### Requirements
- use context API to connect parent with children. see other components like the primitive menu, picklist, vertical navigation.
- when clicking option children it will send the name to VisualPicker parent component and then VisualPicker will send the name in onChange event

`<VisualPicker
      label="string|node"
      onChange="function"
      value="string">`
      `<VisualPickerOption
            disabled="bool"
            name="string"
            footer="node">
            picker content here
      </VisualPickerOption>`
`</VisualPicker>;`

### Design
https://zpl.io/VKyjYKq

Note: follow these accessibilities criteria
https://www.lightningdesignsystem.com/components/visual-picker/#site-main-content

PR #860, opened 2019-07-30
feat: implement VisualPicker component

TahimiLeonBravo:
fix: #584 

Changes proposed in this PR:
- Implement VisualPicker
- Implement VisualPickerOption
- Implement VisualPickerOptionFooter

[ x ] I have followed (at least) the [PR section of the contributing guide](https://github.com/90milesbridge/react-rainbow/blob/master/CONTRIBUTING.md#submitting-a-pull-request).

@90milesbridge/tigger


src/components/VisualPickerOption/index.js
```diff
@@ -1,55 +1,57 @@
+/* eslint-disable react/prop-types */
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
+import { isArray, isString } from 'util';
```
LeandroTorresSicilia: don't use this instead for check array use native Array.isArray and for string use typeof === 'string'

src/components/VisualPickerOption/index.js
```diff
@@ -1,55 +1,57 @@
...
-    //     const { multiple } = this.props;
-    //     if (multiple) {
-    //         return 'checkbox';
-    //     }
-    //     return 'radio';
-    // }
+    getType() {
+        const { multiple } = this.props;
+        return multiple ? 'checkbox' : 'radio';
+    }
```
LeandroTorresSicilia: use with if because it is more readable

src/components/VisualPickerOption/index.js
```diff
@@ -1,55 +1,57 @@
...
 
     isChecked() {
-        // here goes logic to determine if it is checked
-        return false;
+        const { multiple, name, value } = this.props;
+        return multiple
+            ? isArray(value) && value.includes(name)
+            : isString(value) && value !== '' && name === value;
```
LeandroTorresSicilia: same use if instead, this is more difficult to understand
in the case of not multiple the ```value !== ''``` it is not necessary because when make the comparation: ```name === value``` it will fail.

src/components/VisualPickerOption/index.js
```diff
@@ -71,9 +72,21 @@ export default class VisualPickerOption extends Component {
...
+export default function VisualPickerOption(props) {
+    return (
+        <VisualPickerConsumer>
+            {context => <PickerOption {...props} {...context} />}
+        </VisualPickerConsumer>
+    );
+}
+
 VisualPickerOption.propTypes = {
     /** It is a unitque value the identifies the picker option. */
```
LeandroTorresSicilia: It is a unique value that identifies the picker option.

PR #875, opened 2019-08-15
feat: implement picklist

wildergd:
fix: #837 

Changes proposed in this PR:
- implement picklist

@90milesbridge/tigger

### DIFF
--- a/src/components/Picklist/__test__/utils.spec.js
+++ b/src/components/Picklist/__test__/utils.spec.js
@@ -1,0 +1,35 @@
+import { insertChildOrderly, getChildMenuItemNodes } from '../utils';
+
+describe('<ButtonMenu/> utils', () => {
+    describe('insertChildOrderly', () => {
+        it('should insert the child in the right order if the item in the nodes is at middle', () => {
+            const childrenRefs = ['item1', 'item2', 'item4', 'item5'];
+            const nodes = ['item1', 'item2', 'item3', 'item4', 'item5'];
+            expect(insertChildOrderly(childrenRefs, 'item3', nodes)).toEqual(nodes);
+        });
+        it('should insert the child in the right order if the item in the nodes is at end', () => {
+            const childrenRefs = ['item1', 'item2', 'item3'];
+            const nodes = ['item1', 'item2', 'item3', 'item4'];
+            expect(insertChildOrderly(childrenRefs, 'item4', nodes)).toEqual(nodes);
+        });
+    });
+    describe('getChildMenuItemNodes', () => {
+        const elements = [
+            { element: 'a', innerText: 'Julio' },
+            { element: 'a', innerText: 'Pepe' },
+        ];
+        const ref = {
+            querySelectorAll: jest.fn(() => elements),
+        };
+        it('should call querySelectorAll with a[role="menuitem"]', () => {
+            getChildMenuItemNodes(ref);
+            expect(ref.querySelectorAll).toHaveBeenCalledWith('a[role="menuitem"]');
+        });
+        it('should return the elements resolved by querySelectorAll', () => {
+            expect(getChildMenuItemNodes(ref)).toEqual(elements);
+        });
+        it('should return an empty array if anything is passed', () => {
+            expect(getChildMenuItemNodes()).toEqual([]);
+        });
+    });
+});
--- a/src/components/Picklist/helpers/get-children-names.js
+++ b/src/components/Picklist/helpers/get-children-names.js
@@ -1,0 +1,14 @@
+import React from 'react';
+
+export default function getChildrenNames(children) {
+    return React.Children.map(children, child => {
+        if (
+            child &&
+            child.type.displayName === 'PicklistOption' &&
+            typeof child.props.name === 'string'
+        ) {
+            return child.props.name;
+        }
+        return null;
+    });
+}
--- a/src/components/Picklist/helpers/isEmpty.js
+++ b/src/components/Picklist/helpers/isEmpty.js
@@ -1,0 +1,3 @@
+export default function isEmpty(value) {
+    return [null, undefined, '', {}, []].includes(value);
+}
--- a/src/components/Picklist/index.js
+++ b/src/components/Picklist/index.js
@@ -1,0 +1,401 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
+import withReduxForm from '../../libs/hocs/withReduxForm';
+import RenderIf from '../RenderIf';
+import { UP_KEY, DOWN_KEY, ESCAPE_KEY, TAB_KEY, ENTER_KEY } from '../../libs/constants';
+import { Provider } from './context';
+import MenuContent from './menuContent';
+import { insertChildOrderly, getChildMenuItemNodes } from './utils';
+import './styles.css';
+import isEmpty from './helpers/isEmpty';
+
+class Picklist extends Component {
+    constructor(props) {
+        super(props);
+        this.containerRef = React.createRef();
+        this.triggerRef = React.createRef();
+        this.handleInputClick = this.handleInputClick.bind(this);
+        this.handleBlur = this.handleBlur.bind(this);
+        this.handleKeyPressed = this.handleKeyPressed.bind(this);
+        this.hoverChild = this.hoverChild.bind(this);
+        this.handleOptionClick = this.handleOptionClick.bind(this);
+        this.selectOption = this.selectOption.bind(this);
+
+        this.registerChild = this.registerChild.bind(this);
+        this.unregisterChild = this.unregisterChild.bind(this);
+
+        this.currentOption = null;
+        this.activeChildren = [];
+        this.state = {
+            isOpen: false,
+            childrenRefs: [],
+            activeOptionIndex: -1,
+            activeOptionName: null,
+            context: {
+                privateOnClick: this.handleOptionClick,
+                privateRegisterChild: this.registerChild,
+                privateUnregisterChild: this.unregisterChild,
+                privateOnHover: this.hoverChild,
+            },
+        };
+
+        this.keyHandlerMap = {
+            [UP_KEY]: this.handleKeyUpPressed.bind(this),
+            [DOWN_KEY]: this.handleKeyDownPressed.bind(this),
+            [ESCAPE_KEY]: this.closeMenu.bind(this),
+            [TAB_KEY]: this.closeMenu.bind(this),
+            [ENTER_KEY]: this.handleKeyEnterPressed.bind(this),
+        };
+    }
+
+    componentDidUpdate(prevProps) {
+        const { value: prevValue } = prevProps;
+        const { value } = this.props;
+        if (prevValue !== value) {
+            this.selectOption(value);
+        }
+    }
+
+    getContainerClassNames() {
+        const { isOpen } = this.state;
+        const { className } = this.props;
+
+        return classnames(
+            'rainbow-picklist',
+            {
+                'rainbow-picklist--open': isOpen,
+            },
+            className,
+        );
+    }
+
+    getInputClassNames() {
+        const { icon } = this.getValue();
+        return classnames('rainbow-picklist_input', {
+            'rainbow-picklist_input--icon': !!icon,
+        });
+    }
+
+    getDropdownClassNames() {
+        const { isLoading } = this.props;
+        return classnames('rainbow-picklist_dropdown', {
+            'rainbow-picklist_dropdown--loading-box': isLoading,
+        });
+    }
+
+    getValue() {
+        const { value } = this.props;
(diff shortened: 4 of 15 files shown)


src/components/Picklist/helpers/get-children-names.js
```diff
@@ -0,0 +1,14 @@
...
+        if (
+            child &&
+            child.type.displayName === 'PicklistOption' &&
+            typeof child.props.name === 'string'
+        ) {
+            return child.props.name;
+        }
+        return null;
+    });
+}
```
LeandroTorresSicilia: remove this since it is not used

src/components/Picklist/helpers/isEmpty.js
```diff
@@ -0,0 +1,3 @@
+export default function isEmpty(value) {
+    return [null, undefined, '', {}, []].includes(value);
```
LeandroTorresSicilia: when value is {} or [] this always will return false since objects and arrays in javascript are compared by reference not by its value

src/components/Picklist/index.js
```diff
@@ -0,0 +1,401 @@
+import React, { Component } from 'react';
```
LeandroTorresSicilia: Instead of maintain activeChildren and currentOption what you can do is register and unregister when in child when is deselected and selected in componentDidUpdate in PicklistOption component. Try this since it will simply all logic related to activeChildren

src/components/Picklist/index.js
```diff
@@ -0,0 +1,401 @@
...
+            isOpen: false,
+            childrenRefs: [],
+            activeOptionIndex: -1,
+            activeOptionName: null,
+            context: {
+                privateOnClick: this.handleOptionClick,
+                privateRegisterChild: this.registerChild,
+                privateUnregisterChild: this.unregisterChild,
+                privateOnHover: this.hoverChild,
+            },
```
LeandroTorresSicilia: since context it is not reactive we can put this in a class variable as:
```this.context```

File: src/components/Picklist/index.js [hunk 5]
```diff
@@ -0,0 +1,401 @@
...
+        const [...nodes] = getChildMenuItemNodes(this.containerRef.current);
+        const newChildrenRefs = insertChildOrderly(childrenRefs, childRef, nodes);
+        const refIndex = newChildrenRefs.indexOf(childRef);
+        const extendedChildProps = {
+            ...childProps,
+            ref: childRef,
+        };
+        const { value } = this.props;
+        if (!isEmpty(value)) {
+            if (isEmpty(this.currentOption) && childProps.name === value.name) {
```
LeandroTorresSicilia: I think you don't need to maintain this currentOption state since  it is much the same as value prop, instead make a helper function to compute what you need based on value prop, let's stick with functional model.

src/components/Picklist/index.js
```diff
@@ -0,0 +1,401 @@
...
+                this.currentOption = {
+                    index: currentOptionIndex,
+                    optionProps: this.activeChildren[currentOptionIndex],
+                };
+                this.activeChildren.splice(currentOptionIndex, 1);
+            } else {
+                this.activeChildren.splice(refIndex, 0, extendedChildProps);
+            }
+        } else {
+            this.activeChildren.splice(refIndex, 0, extendedChildProps);
```
LeandroTorresSicilia: here you are mutating this.activeChildren, don't use splice, remember we need to avoid mutations,
for example this code achieve the same without mutation:
```
this.activeChildren = this.activeChildren.concat(extendedChildProps);
```

src/components/Picklist/styles.scss
```diff
@@ -0,0 +1,224 @@
...
+}
+
+.rainbow-picklist_dropdown {
+    position: absolute;
+    z-index: 2000;
+    left: 50%;
+    float: left;
+    width: 100%;
+    // min-width: 6rem;
+    // max-width: 20rem;
```
LeandroTorresSicilia: remove comments

src/components/PicklistOption/index.js
```diff
@@ -0,0 +1,179 @@
...
+        if (disabled || isHeader || typeof name !== 'string') {
+            return null;
+        }
+        return setTimeout(
+            () =>
+                privateRegisterChild(this.itemRef.current, {
+                    label,
+                    name,
+                    icon,
+                    value,
```
LeandroTorresSicilia: here only send what you need, I believe you only need the name

wildergd: it is necessary to send all because in PickList.handleKeyEnterPressed we have to fire an onChange event

LeandroTorresSicilia: 👍

src/components/Picklist/index.js
```diff
@@ -215,10 +169,11 @@ class Picklist extends Component {
     }
 
     closeMenu() {
+        // const { activeChildren } = this.state;
```
LeandroTorresSicilia: remove comment

src/components/Picklist/index.js
```diff
@@ -303,11 +235,14 @@ class Picklist extends Component {
...
-        const { context, activeOptionName } = this.state;
+        const { activeOptionName } = this.state;
         const ariaLabel = title || assistiveText;
         const { label, icon, name: currentValueName } = this.getValue();
         const providerContext = {
-            ...context,
+            privateOnClick: this.handleOptionClick,
+            privateRegisterChild: this.registerChild,
+            privateUnregisterChild: this.unregisterChild,
+            privateOnHover: this.hoverChild,
```
LeandroTorresSicilia: what I meant (we can put this in a class variable like this.context, see above) is that it is not necessary in state but you can put it in constructor like:
```
this.context = {
                privateOnClick: this.handleOptionClick,
                privateRegisterChild: this.registerChild,
                privateUnregisterChild: this.unregisterChild,
                privateOnHover: this.hoverChild,
            }
```

src/components/Picklist/__test__/utils.spec.js
```diff
@@ -0,0 +1,35 @@
...
+            expect(ref.querySelectorAll).toHaveBeenCalledWith('a[role="menuitem"]');
+        });
+        it('should return the elements resolved by querySelectorAll', () => {
+            expect(getChildMenuItemNodes(ref)).toEqual(elements);
+        });
+        it('should return an empty array if anything is passed', () => {
+            expect(getChildMenuItemNodes()).toEqual([]);
+        });
+    });
+});
```
LeandroTorresSicilia: since you change this helpers fix this tests in order to close correctly this PR
the remaining tests will be in another PR

src/components/Picklist/index.js
wildergd: this is just a waste of memory.... as in the render we include activeOptionName and  currentValueName...  that are dynamic variables... there is no need to have this twice... another option is to create a function buildContext() and pass it directly to Provider value.. i think its less confusing

TahimiLeonBravo:
We need to implement arrows when having many options in the `Picklist`, please follow the design
picklist

TahimiLeonBravo:
Fix this visual error
Screen Shot 2019-08-21 at 2 44 35 AM

src/components/Picklist/helpers/isOptionVisible.js
```diff
@@ -0,0 +1,8 @@
+export default function isOptionVisible(el, container) {
+    const { top: elTop, bottom: elBottom } = el.getBoundingClientRect();
+    const { top: containerTop, bottom: containerBottom } = container.getBoundingClientRect();
+    return (
+        Math.floor(elTop) >= Math.floor(containerTop) &&
+        Math.ceil(elBottom) <= Math.ceil(containerBottom)
```
LeandroTorresSicilia: Why use Math.floor and Math.ceil?

src/components/Picklist/index.js
```diff
@@ -182,8 +207,18 @@ class Picklist extends Component {
     }
 
     openMenu() {
+        const { activeChildren } = this.state;
+        const firstOptionIndex = activeChildren.length > 0 ? 0 : -1;
+        const firstOptionName = activeChildren.length > 0 ? activeChildren[0].name : null;
+
+        setTimeout(() => {
+            this.scrollTo(0);
+        }, 0);
```
LeandroTorresSicilia: let's try to avoid setTimeout, instead use didUpdate to check when open (!prevIsOpen && isOpen) the menu and then here is where you scroll without timeout

src/components/Picklist/index.js
```diff
@@ -223,6 +258,38 @@ class Picklist extends Component {
...
 
+    // isScrollUpArrowVisible() {
+    //     return this.menuRef.current && this.menuRef.current.scrollTop > 0;
+    // }
+
+    // isScrollDownArrowVisible() {
+    //     const menu = this.menuRef.current;
+    //     if (!menu) return false;
+    //     return menu.scrollHeight - menu.scrollTop !== menu.clientHeight;
+    // }
```
LeandroTorresSicilia: remove this comments if not needed

src/components/Picklist/index.js
```diff
@@ -265,6 +335,12 @@ class Picklist extends Component {
         const { label: valueLabel, icon } = this.getValue();
         const value = valueLabel || '';
 
+        const menuContainerStyles = {
+            // height: 48 * items.length + 17,
```
LeandroTorresSicilia: clean all comments

src/components/Picklist/index.js
```diff
@@ -338,6 +434,13 @@ Picklist.propTypes = {
     style: PropTypes.object,
     /** The id of the outer element. */
     id: PropTypes.string,
+    /** Specifies that the Lookup must be filled out before submitting the form.
```
LeandroTorresSicilia: Specifies that an option must be selected before submitting the form.

src/components/Picklist/menuArrowButton.js
```diff
@@ -0,0 +1,18 @@
...
+        'rainbow-picklist_dropdown-arrow-button',
+        `rainbow-picklist_dropdown-arrow-${arrow}`,
+    );
+
+    return <div className={className} />;
+}
+
+MenuArrowButton.propTypes = {
+    arrow: PropTypes.oneOf(['up', 'down']).isRequired,
+    // onHover: PropTypes.func.isRequired,
```
LeandroTorresSicilia: comment

LeandroTorresSicilia:
Other issues:
1. When press space to open the menu we need to prevent default behavior because space make scroll to move down.
2. Set label prop as required.
3. Right now readOnly prop does not make any visual change and disabled is visual broken.
4. When hover arrows is expected that work same as select.
5. See @TahimiLeonBravo to make the right example, only add the redux form example and fix styles to it.

src/components/Picklist/helpers/isOptionVisible.js
wildergd: because getboundingclientrect returns floats and sometimes an small fraction casues failure

wildergd:
> Other issues:
>     1. When press space to open the menu we need to prevent default behavior because space make scroll to move down.
 i think this issue is related to mouse position on the screen... put mouse pointer outside the component...probably on the top.. and use tab  to move onto it... and press space.. ... what i think is happening .. is that when you open the menu using space... the mouse pointer is located over an arrow button and fires the event....
check if that is what is happening

wildergd:
> Other issues:
>     1. When press space to open the menu we need to prevent default behavior because space make scroll to move down.
 i think this issue is related to mouse position on the screen... put mouse pointer outside the component...probably on the top.. and use tab  to move onto it... and press space.. ... what i think is happening .. is that when you open the menu using space... the mouse pointer is located over an arrow button and fires the event....
check if that is what is happening..

LeandroTorresSicilia:
> > Other issues:
> > 1. When press space to open the menu we need to prevent default behavior because space make scroll to move down.
> > i think this issue is related to mouse position on the screen... put mouse pointer outside the component...probably on the top.. and use tab  to move onto it... and press space.. ... what i think is happening .. is that when you open the menu using space... the mouse pointer is located over an arrow button and fires the event....
> > check if that is what is happening

When a space key is pressed the browser default behavior is scroll down, so when keyDown handler is fired you need to check if the key pressed is space and then call event.preventDefault():
```
if (keyCode === ESCAPE_KEY) {
    event.preventDefault();
}
```
import ESCAPE_KEY from libs/constants.js
this is a file where we store key codes values that are reused in whatever component needs it

TahimiLeonBravo:
Change the arrow color to gray when the Picklist is disabled
disabled

TahimiLeonBravo:
Fix the focus styles, it should looks like the Input
focus

TahimiLeonBravo:
change the styles the options when hover to `$color-gray-1`
Screen Shot 2019-08-23 at 2 41 22 PM

TahimiLeonBravo:
fix the keyboard scroll behavior, it should work like the Lookup

PR #899, opened 2019-09-03
test: add unit test to helpers and ampmSelect on TimePicker component

cmcartaya:
Changes proposed in this PR:
- unit test to helpers and ampmSelect on TimePicker component

@90milesbridge/tigger

### DIFF
--- a/src/components/TimePicker/__test__/ampmSelect.spec.js
+++ b/src/components/TimePicker/__test__/ampmSelect.spec.js
@@ -1,0 +1,92 @@
+import React from 'react';
+import { mount } from 'enzyme';
+import AmPmSelect from './../ampmSelect';
+
+describe('<AmPmSelect/>', () => {
+    it('should render the right markup initially when component is not focused', () => {
+        const component = mount(<AmPmSelect tabIndex="-1" />);
+        component.setState({ isFocused: false });
+        expect(
+            component.find('input[className="rainbow-time-picker_time-select-value"]').exists(),
+        ).toBe(true);
+    });
+    it('should render the right markup when component is focused', () => {
+        const component = mount(<AmPmSelect />);
+        component.setState({ isFocused: true });
+        expect(
+            component
+                .find(
+                    'fieldset[className="rainbow-time-picker_time-select-value rainbow-time-picker_select-ampm"]',
+                )
+                .exists(),
+        ).toBe(true);
+        expect(component.find('input[value="AM"]').exists()).toBe(true);
+        expect(component.find('input[value="PM"]').exists()).toBe(true);
+    });
+    it('should fire the onChange event with "AM" when the component get focus and value and defaultValue are undefined', () => {
+        const onChangeMckFn = jest.fn();
+        const component = mount(<AmPmSelect onChange={onChangeMckFn} />);
+        component.setState({ isFocused: false });
+        component
+            .find('input[className="rainbow-time-picker_time-select-value"]')
+            .simulate('focus');
+        expect(component.prop('onChange')).toHaveBeenCalledWith('AM');
+        expect(component.state('isFocused')).toBe(true);
+    });
+    it('should fire the onChange event with the defaultValue passed when the component get focus and the prop value is undefined', () => {
+        const onChangeMckFn = jest.fn();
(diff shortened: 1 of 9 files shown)


src/components/TimePicker/__test__/ampmSelect.spec.js
```diff
@@ -0,0 +1,92 @@
+import React from 'react';
+import { mount } from 'enzyme';
+import AmPmSelect from './../ampmSelect';
+
+describe('<AmPmSelect/>', () => {
+    it('should render the right markup initially when component is not focused', () => {
+        const component = mount(<AmPmSelect tabIndex="-1" />);
+        component.setState({ isFocused: false });
```
LeandroTorresSicilia: we should not set isFocused to false here since the test is for check what is render initially and therefore this implies that we are also testing that initially the isFocused is false, but if you manually set isFocused to false then this test no make sense

src/components/TimePicker/__test__/ampmSelect.spec.js
```diff
@@ -0,0 +1,92 @@
...
+        expect(
+            component.find('input[className="rainbow-time-picker_time-select-value"]').exists(),
+        ).toBe(true);
+    });
+    it('should render the right markup when component is focused', () => {
+        const component = mount(<AmPmSelect />);
+        component.setState({ isFocused: true });
```
LeandroTorresSicilia: I think it is better to simulate focus instead of set state

src/components/TimePicker/__test__/ampmSelect.spec.js
```diff
@@ -0,0 +1,92 @@
...
+                )
+                .exists(),
+        ).toBe(true);
+        expect(component.find('input[value="AM"]').exists()).toBe(true);
+        expect(component.find('input[value="PM"]').exists()).toBe(true);
+    });
+    it('should fire the onChange event with "AM" when the component get focus and value and defaultValue are undefined', () => {
+        const onChangeMckFn = jest.fn();
+        const component = mount(<AmPmSelect onChange={onChangeMckFn} />);
+        component.setState({ isFocused: false });
```
LeandroTorresSicilia: why set isFocused to false if initially it has this value?

src/components/TimePicker/__test__/ampmSelect.spec.js
```diff
@@ -0,0 +1,92 @@
...
+        component
+            .find('input[className="rainbow-time-picker_time-select-value"]')
+            .simulate('focus');
+        expect(component.prop('onChange')).toHaveBeenCalledWith('AM');
+        expect(component.state('isFocused')).toBe(true);
```
LeandroTorresSicilia: this are the stuff we should avoid to test, also in a previous test we already test this but checking what is important: the UI rendered

src/components/TimePicker/__test__/ampmSelect.spec.js
```diff
@@ -0,0 +1,92 @@
...
+    });
+    it('should fire the onChange event with the defaultValue passed when the component get focus and the prop value is undefined', () => {
+        const onChangeMckFn = jest.fn();
+        const component = mount(<AmPmSelect defaultValue="PM" onChange={onChangeMckFn} />);
+        component.setState({ isFocused: false });
```
LeandroTorresSicilia: same ^^ this is not needed

src/components/TimePicker/__test__/ampmSelect.spec.js
```diff
@@ -0,0 +1,92 @@
...
+        component
+            .find('input[className="rainbow-time-picker_time-select-value"]')
+            .simulate('focus');
+        expect(component.prop('onChange')).toHaveBeenCalledWith('PM');
+        expect(component.state('isFocused')).toBe(true);
```
LeandroTorresSicilia: same ^^

src/components/TimePicker/helpers/__test__/isChecked.spec.js
```diff
@@ -0,0 +1,74 @@
+import isChecked from '../isChecked';
+
+describe('isChecked', () => {
+    it('should return true when value is equal to inputValue and value is truety', () => {
```
LeandroTorresSicilia: truety => truthy

PR #900, opened 2019-09-07
test: add unit test on timeSelect

cmcartaya:
Changes proposed in this PR:
- unit test on timeSelect of the TimePicker component

@90milesbridge/tigger

### DIFF
--- a/src/components/TimePicker/__test__/timeSelect.spec.js
+++ b/src/components/TimePicker/__test__/timeSelect.spec.js
@@ -350,26 +350,33 @@ describe('<TimeSelect/>', () => {
         const focusedElementDataId = document.activeElement.getAttribute('data-id');
         expect(focusedElementDataId).toBe('input-element');
     });
-    it('should focus PM input when AM input is focused and press up key', () => {
+    it('should pass the right defaultValue to AmPmSelect when up or down key is not pressed and does not have value initially', () => {
         const component = mount(<TimeSelect />);
-        const container = component.find('div[role="presentation"]');
-        container.simulate('keyDown', { keyCode: 39 });
-        container.simulate('keyDown', { keyCode: 39 });
-        container.simulate('keyDown', { keyCode: 38 });
-        const focusedElementDataId = document.activeElement.getAttribute('value');
-        expect(focusedElementDataId).toBe('PM');
+        const values = ['12', '11', '10', '5', '0', '1'];
+        const expects = ['PM', 'AM', 'AM', 'AM', 'AM', 'AM'];
+        values.forEach((value, index) => {
+            const container = component.find('div[role="presentation"]');
+            const hourInput = component.find('input').at(0);
+            hourInput.simulate('change', { target: { value } });
+            container.simulate('keyDown', { keyCode: 39 });
+            container.simulate('keyDown', { keyCode: 39 });
+            expect(component.find('AmPmSelect').prop('defaultValue')).toBe(expects[index]);
+        });
     });
-    it('should focus AM input when PM input is focused and press down key', () => {
+    it('should pass the right value to AmPmSelect when it is focused and up or down key is pressed', () => {
         const component = mount(<TimeSelect />);
-        const container = component.find('div[role="presentation"]');
-        container.simulate('keyDown', { keyCode: 39 });
-        container.simulate('keyDown', { keyCode: 39 });
-        container.simulate('keyDown', { keyCode: 38 });
-        container.simulate('keyDown', { keyCode: 40 });
-        const focusedElementDataId = document.activeElement.getAttribute('value');
-        expect(focusedElementDataId).toBe('AM');
+        const values = [38, 40, 38, 38, 40, 40, 38];
+        const expects = ['PM', 'AM', 'PM', 'AM', 'PM', 'AM', 'PM'];
+
+        values.forEach((value, index) => {
+            const container = component.find('div[role="presentation"]');
+            container.simulate('keyDown', { keyCode: 39 });
+            container.simulate('keyDown', { keyCode: 39 });
+            container.simulate('keyDown', { keyCode: value });
+            expect(component.find('AmPmSelect').prop('value')).toBe(expects[index]);
+        });
     });
-    it('should reset hour input when it is focused, show the 7 hour and press delete key', () => {
+    it('should reset hour input when has a value and press delete key while it is focused', () => {
         const component = mount(<TimeSelect />);
         const hourInput = component.find('input').at(0);
         const container = component.find('div[role="presentation"]');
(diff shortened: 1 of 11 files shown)


src/components/TimePicker/__test__/timeSelect.spec.js
```diff
@@ -3,21 +3,29 @@ import { mount } from 'enzyme';
 import TimeSelect from '../timeSelect';
 
 describe('<TimeSelect/>', () => {
-    it('should set hour state to "01" when press up key', () => {
+    it('should set hour value to "01" when hour input is focused and press up key', () => {
         const component = mount(<TimeSelect />);
-        const hourInput = component.find('input').at(0);
-        hourInput.simulate('focus');
         component.find('div[role="presentation"]').simulate('keyDown', { keyCode: 38 });
```
LeandroTorresSicilia: use constants defined in libs instead of keyCode directly

src/components/TimePicker/__test__/timeSelect.spec.js
```diff
@@ -3,21 +3,29 @@ import { mount } from 'enzyme';
...
+                .at(0)
+                .prop('value'),
+        ).toBe('01');
     });
-    it('should set hour state to "12" when press down key', () => {
+    it('should set hour value to "12" when hour input is focused and press down key', () => {
         const component = mount(<TimeSelect />);
         const hourInput = component.find('input').at(0);
         hourInput.simulate('focus');
         component.find('div[role="presentation"]').simulate('keyDown', { keyCode: 40 });
```
LeandroTorresSicilia: same ^^ change in all places

src/components/TimePicker/__test__/timeSelect.spec.js
```diff
@@ -79,33 +92,392 @@ describe('<TimeSelect/>', () => {
...
+                .prop('value'),
+        ).toBe('02');
+    });
+    it('should set the right minutes value and focus ampm input', () => {
+        const component = mount(<TimeSelect />);
+        const values = [
+            '022',
+            '000',
+            '001',
+            '022',
```
LeandroTorresSicilia: 022 case is twice
my advice is to follow a order to not make this mistakes

src/components/TimePicker/__test__/timeSelect.spec.js
```diff
@@ -79,33 +92,392 @@ describe('<TimeSelect/>', () => {
...
+        container.simulate('keyDown', { keyCode: 39 });
+        container.simulate('keyDown', { keyCode: 39 });
+        const focusedElementDataId = document.activeElement.getAttribute('data-id');
+        expect(focusedElementDataId).toBe('input-element');
+    });
+    it('should pass the right defaultValue to AmPmSelect when up or down key is not pressed and does not have value initially', () => {
+        const component = mount(<TimeSelect />);
+        const values = ['12', '11', '10', '5', '0', '1'];
+        const expects = ['PM', 'AM', 'AM', 'AM', 'AM', 'AM'];
+        values.forEach((value, index) => {
```
LeandroTorresSicilia: this is not needed since what computes defaultValue is a external function then you only need to test one case:
that what is passed as defaultValue to AmPmSelect is what is returned by this function, then when test the function in isolation there will be where you test all possible cases

src/components/TimePicker/__test__/timeSelect.spec.js
```diff
@@ -79,33 +92,392 @@ describe('<TimeSelect/>', () => {
...
+            const container = component.find('div[role="presentation"]');
+            const hourInput = component.find('input').at(0);
+            hourInput.simulate('change', { target: { value } });
+            container.simulate('keyDown', { keyCode: 39 });
+            container.simulate('keyDown', { keyCode: 39 });
```
LeandroTorresSicilia: this is not needed for the purpose of this test

src/components/TimePicker/__test__/timeSelect.spec.js
```diff
@@ -79,33 +92,392 @@ describe('<TimeSelect/>', () => {
...
+        container.simulate('keyDown', { keyCode: 13 });
+        expect(onChangeMockFn).toHaveBeenCalledWith('12:14');
+    });
+    it('should not call onChange when press enter key while hour, minutes and ampm has not value', () => {
+        const onChangeMockFn = jest.fn();
+        const component = mount(<TimeSelect onChange={onChangeMockFn} />);
+        const container = component.find('div[role="presentation"]');
+        container.simulate('keyDown', { keyCode: 13 });
+        expect(onChangeMockFn).toHaveBeenCalledTimes(0);
     });
```
LeandroTorresSicilia: more tests to add:
- should call onCloseModal when click the cancel button
and add the cases when press ok button

src/components/TimePicker/helpers/__test__/getDefaultAmPm.spec.js
```diff
@@ -0,0 +1,12 @@
+import getDefaultAmPm from '../getDefaultAmPm';
+
+describe('getDefaultAmPm', () => {
+    it('should return "PM" when value passed is a number and it is greater than 11 and less than 20', () => {
+        const values = [12, 13, 14, 15, 16, 17, 18, 19];
+        values.forEach(value => expect(getDefaultAmPm(value)).toBe('PM'));
+    });
+    it('should return "AM" when value passed is a number and it is less than 11 and greater than 20', () => {
+        const values = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 20, 21, 22, 23, 34];
```
LeandroTorresSicilia: there is not need to put so much values

src/components/TimePicker/helpers/__test__/getHour.spec.js
```diff
@@ -0,0 +1,13 @@
+import getHour from '../getHour';
+
+describe('getHour', () => {
+    it('should return the right hour when value passed is a valid string', () => {
```
LeandroTorresSicilia: ```when value passed is a valid string``` is not exact here you can only put:
```should return the right hour value```
or:
```should return the right hour when value passed is a valid formatted time```
same for getAmPm helper

File: src/components/TimePicker/helpers/__test__/getMinutes.spec.js [hunk 9]
```diff
@@ -0,0 +1,13 @@
+import getMinutes from '../getMinutes';
+
+describe('getMinutes', () => {
+    it('should return the right minutes when value passed is a valid string', () => {
```
LeandroTorresSicilia: same ^^

src/components/TimePicker/helpers/__test__/isNumber.spec.js
```diff
@@ -0,0 +1,12 @@
+import isNumber from '../isNumber';
+
+describe('isNumber', () => {
+    it('should return true when the value paseed is an entery number, it is not an empty string and it is not an array', () => {
```
LeandroTorresSicilia: what you mean is integer? entery does not exists and change paseed => passed
also change this names to:
- should return true when the value passed is a positive integer number
- should return false when the value passed is not a positive integer number

Issue #795, opened 2019-06-09
test: add UnitTest to the TimePicker component


TahimiLeonBravo:
#### Helpers
**getNextAmPmValue**
- [x] should return the string "PM" when value passed is equal to "AM"
- [x] should return the string "PM" when value passed is undefined
- [x] should return the string "AM" when value passed is different to "AM" or undefined
**getNextHour**
- [x] should return the string "1" when any value is passed
- [x] should return the string "1" when value passed converted to Number is equal to 12
- [x] should return the string of the value passed increased by 1 when value passed converter to Number is different to 12
**getNextMinute**
- [x] should return the string "0" when any value is passed
- [x] should return the string "0" when value passed converted to Number is equal to 59
- [x] should return the string of the value passed increased by 1 when value passed converter to Number is different to 59
**getPrevHour**
- [x] should return the string "12" when any value is passed
- [x] should return the string "12" when value passed converted to Number is equal to 1
- [x] should return the string of the value passed decreased by 1 when value passed converter to Number is different to 1
**getPrevMinute**
- [x] should return the string "59" when any value is passed
- [x] should return the string "59" when value passed converted to Number is equal to 0
- [x] should return the string of the value passed decreased by 1 when value passed converter to Number is different to 0
**AmPmSelect**
- [x] should render the right markup initially when component is not focused
- [x] should render the right markup when component is focused
- [x] should fire the onChange event with "AM" when the component get focus and value and defaultValue are undefined
- [x] should fire the onChange event with the defaultValue passed when the component get focus and the prop value is undefined
- [x] should render the initial input when component lost focus
- [x] should focus the component when the focus method is called and the component is focused
- [x] should focus the component when the focus method is called and the component is not focused
- [x] should call event.stopPropagation when component is focused and the input with value "AM" is blurred
- [x] should call event.stopPropagation when component is focused and the input with value "PM" is blurred
**isChecked**
- [x] should return true when value is equal to inputValue
- [x] should return true when defaultValue is equal to inputValue and value is undefined
- [x] should return true when inputValue is equal to "AM" if defaultValue and value are undefined 
**TimeSelect**
- [x] should set hour value to "01" when hour input is focused and press up key
- [x] should set hour value to "12" when hour input is focused and press down key
- [x] should set hour value to "05" when hour input is focused and press up key for 5 times
- [x] should set the right hour value and focus minutes input
- [x] should set hour value to "01" when hour input is focused and click up button
[...]

PR #920, opened 2019-09-13
test: add unit test on timePicker component

cmcartaya:
fix: #795 

## - add unit test on TimePicker Component

@nexxtway/tigger

### DIFF
--- a/src/components/TimePicker/__test__/timePicker.spec.js
+++ b/src/components/TimePicker/__test__/timePicker.spec.js
@@ -1,34 +1,139 @@
 import React from 'react';
 import { mount } from 'enzyme';
 import TimePicker from '../';
+import { SPACE_KEY, ENTER_KEY } from '../../../libs/constants';
 
 describe('<TimePicker/>', () => {
     it('should fire onBlur with undefined when there is not value', () => {
         const onBlurMockFn = jest.fn();
-        const component = mount(<TimePicker onBlur={onBlurMockFn} />);
+        const component = mount(
+            <TimePicker label="unit-testing-timePicker" onBlur={onBlurMockFn} />,
+        );
         component.find('input').simulate('focus');
         component.find('input').simulate('blur');
         expect(onBlurMockFn).toHaveBeenCalledWith(undefined);
     });
     it('should fire onBlur with the value passed', () => {
         const onBlurMockFn = jest.fn();
-        const component = mount(<TimePicker value="18:35" onBlur={onBlurMockFn} />);
+        const component = mount(
+            <TimePicker label="unit-testing-timePicker" value="18:35" onBlur={onBlurMockFn} />,
+        );
         component.find('input').simulate('focus');
         component.find('input').simulate('blur');
         expect(onBlurMockFn).toHaveBeenCalledWith('18:35');
     });
     it('should fire onFocus with undefined when there is not value', () => {
         const onFocusMockFn = jest.fn();
-        const component = mount(<TimePicker onFocus={onFocusMockFn} />);
+        const component = mount(
+            <TimePicker label="unit-testing-timePicker" onFocus={onFocusMockFn} />,
+        );
         component.find('input').simulate('focus');
         component.find('input').simulate('blur');
         expect(onFocusMockFn).toHaveBeenCalledWith(undefined);
     });
     it('should fire onFocus with the value passed', () => {
         const onFocusMockFn = jest.fn();
-        const component = mount(<TimePicker value="18:35" onFocus={onFocusMockFn} />);
+        const component = mount(
+            <TimePicker label="unit-testing-timePicker" value="18:35" onFocus={onFocusMockFn} />,
+        );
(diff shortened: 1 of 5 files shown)


src/components/TimePicker/__test__/timePicker.spec.js
```diff
@@ -1,34 +1,139 @@
...
         expect(onBlurMockFn).toHaveBeenCalledWith('18:35');
     });
     it('should fire onFocus with undefined when there is not value', () => {
         const onFocusMockFn = jest.fn();
-        const component = mount(<TimePicker onFocus={onFocusMockFn} />);
+        const component = mount(
+            <TimePicker label="unit-testing-timePicker" onFocus={onFocusMockFn} />,
+        );
         component.find('input').simulate('focus');
         component.find('input').simulate('blur');
```
LeandroTorresSicilia: clean this since what it is tested is focus then blur is not needed

src/components/TimePicker/__test__/timePicker.spec.js
```diff
@@ -1,34 +1,139 @@
...
+            <TimePicker label="unit-testing-timePicker" placeholder="testing-timePicker" />,
+        );
+        const instance = component.instance();
+        const focusedElementPlaceholder = document.activeElement.getAttribute('placeholder');
+
+        instance.focus();
+        expect(document.activeElement.getAttribute('placeholder')).toBe(focusedElementPlaceholder);
+        instance.blur();
+        expect(document.activeElement.getAttribute('placeholder')).toBeNull();
+    });
```
LeandroTorresSicilia: there is a toBeFocusable custom jest matcher for not repeat this in all focusable components so remove this two test
make this test in a new file called ```timePicker.focusable.spec.js```
look at button or input component to see how it is used
this matcher check that onClick, onFocus, and onBlur events are called and that the component have the imperative methods (click, focus, blur).
in the case of the onBlur and onFocus already testes we need to keep them since what is tested is that they are called with the value and not with the native event

src/components/TimePicker/__test__/timePicker.spec.js
```diff
@@ -1,34 +1,139 @@
...
+    it('should set isOpen to true when enter key or space key is pressed while input is focused and readOnly is not passed', () => {
```
LeandroTorresSicilia: i think it is better if you say:
```should open the modal when ...``` and ```should not open the modal when ...```

File: src/components/TimePicker/__test__/timePicker.spec.js [hunk 4]
```diff
@@ -1,34 +1,139 @@
...
+            const timePickerInput = component.find('input');
+            timePickerInput.simulate('focus');
+            timePickerInput.simulate('keyDown', { keyCode: value });
+            expect(component.find('Modal').prop('isOpen')).toBe(false);
+        });
+    });
+    it('should call to focusHourInput when the modal is open', () => {
+        const component = mount(<TimePicker label="unit-testing-timePicker" />);
+        component.find('input').simulate('click');
+        component.update();
```
LeandroTorresSicilia: is this necessary?

src/components/TimePicker/__test__/timePicker.spec.js
```diff
@@ -1,34 +1,139 @@
...
+        component.instance().fieldRef.current.timeSelectRef.current.focusHourInput = jest.fn();
+        setTimeout(() => {
+            expect(
+                component.instance().fieldRef.current.timeSelectRef.current.focusHourInput,
+            ).toHaveBeenCalledTimes(1);
+        }, 1000);
```
LeandroTorresSicilia: why wait for 1000 ms?
if the problem is with the next tick as we talk then with 0 should work

src/components/TimePicker/__test__/timePicker.spec.js
```diff
@@ -1,34 +1,139 @@
...
+    });
+    it('should call updateValue when value change dynamically', () => {
+        const component = mount(<TimePicker label="unit-testing-timePicker" value="22:59" />);
+        component.instance().fieldRef.current.updateValue = jest.fn();
+        component.setProps({ value: '23:01' });
+        component.update();
+        expect(component.instance().fieldRef.current.updateValue).toHaveBeenCalledTimes(1);
+    });
+    it('should set the right value when value change dynamically', () => {
```
LeandroTorresSicilia: should set the right input value when value change dynamically

src/components/TimePicker/__test__/timePicker.spec.js
```diff
@@ -1,34 +1,139 @@
...
+        const component = mount(<TimePicker label="unit-testing-timePicker" value="22:59" />);
+        component.setProps({ value: '23:01' });
+        component.update();
+        expect(component.find('input').prop('value')).toBe('11:01 PM');
+    });
```
LeandroTorresSicilia: this both above are one test
remember that this.updateValue is internal and we should not test that this.updateValue was called, but that the input get the new value like this last test do

Issue #789, opened 2019-06-07
feat: make controllable the selected options in the Lookup component


TahimiLeonBravo:
- As a customer, I want to control the option selected initially.

 ### Spec
- Add prop `preferredSelectedOption`
- The description of the prop should say "The index of the option that is visual-focus initially"

### Acceptance Criteria
- [ ] it should set the initial visual-focus to the option that is passed as `prefereedSelectedOption`
- [ ] it should fallback to 0 if the index passed as 
prefereedSelectedOption does not exist.

### Test
- [ ] should set the initial visual-focus to the option that is passed as `prefereedSelectedOption`
- [ ] should fallback to 0 if the index passed as prefereedSelectedOption does not exist.

PR #930, opened 2019-09-13
feat: make controllable the selected options in the Lookup component

wildergd:
fix: #789 

- add preferredSelectedOption prop to component
- add tests  

@nexxtway/tigger

### DIFF
--- a/src/components/Lookup/helpers/__test__/getInitialFocusedIndex.spec.js
+++ b/src/components/Lookup/helpers/__test__/getInitialFocusedIndex.spec.js
@@ -17,9 +17,16 @@ describe('getInitialFocusedIndex', () => {
         expect(getInitialFocusedIndex(options)).toBe(1);
     });
     it('should return 0 when preferredSelectedOption is an invalid index', () => {
-        const options = [{ label: 'Paris' }, { label: 'New York' }];
+        const options = [{ label: 'New York' }, { label: 'San Fransisco' }];
         expect(getInitialFocusedIndex(options, -1)).toBe(0);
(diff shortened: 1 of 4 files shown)


src/components/Lookup/index.js
```diff
@@ -68,12 +71,15 @@ class Lookup extends Component {
 
     componentDidUpdate(prevProps) {
         const { options: prevOptions } = prevProps;
-        const { options } = this.props;
+        const { options, preferredSelectedOption } = this.props;
         if (prevOptions !== options) {
```
LeandroTorresSicilia: we also need to check whether preferredSelectedOption change to recompute focusedItemIndex again, make it in other if condition since in this case we don't need to recompute normalizedOptions3

src/components/Lookup/helpers/__test__/getInitialFocusedIndex.spec.js
```diff
@@ -16,4 +16,29 @@ describe('getInitialFocusedIndex', () => {
         ];
         expect(getInitialFocusedIndex(options)).toBe(1);
     });
+    it('should return 0 when preferredSelectedOption is an invalid index', () => {
+        const options = [{ label: 'Paris' }, { label: 'New York' }];
+        expect(getInitialFocusedIndex(options, -1)).toBe(0);
+        expect(getInitialFocusedIndex(options, 3)).toBe(0);
+    });
```
LeandroTorresSicilia: add case for when preferredSelectedOption is greater than nonHeaderItems length

Issue #829, opened 2019-07-08
feat: implement `copy and paste` in the value Lookup 


TahimiLeonBravo:
As a customer, I want to copy the value selected in the lookup without the need the select the text manually 

### Acceptance Criteria
- [x] should copy to clipboard the label of the selected option when the input gets focus (clicking the input or the input label, focus the input through tab keys)

## Test
- [x] should copy to clipboard the label of the selected option when the input gets focus

PR #935, opened 2019-09-15
feat: add copy to clipboard functionality when focused

wildergd:
fix: #829

 - add copy to clipboard functionality when lookup is focused

@nexxtway/tigger


src/components/Lookup/__test__/lookup.spec.js
```diff
@@ -377,4 +380,16 @@ describe('<Lookup />', () => {
         component.find('input').simulate('focus');
         expect(component.find('Options').prop('focusedItemIndex')).toBe(0);
     });
+    it('should copy value label to clipboard when input gets focus', () => {
+        jest.mock('clipboard-copy');
```
LeandroTorresSicilia: you don't need to repeat this here

src/components/Lookup/__test__/lookup.spec.js
```diff
@@ -377,4 +380,16 @@ describe('<Lookup />', () => {
...
+
+        const value = { label: 'San Francisco' };
+        const options = [
+            { label: 'San Francisco' },
+            { label: 'New York', description: 'awesome city' },
+        ];
+        const component = mount(<Lookup label="custom label" options={options} value={value} />);
+        component.find('input').simulate('focus');
+        expect(copyFn).toHaveBeenCalledWith(value.label);
+    });
```
LeandroTorresSicilia: this test belongs to selected value component so move to selectedValue.spec.js

Issue #547, opened 2019-01-20
feat: implement ProgressCircular component


TahimiLeonBravo:
https://zpl.io/VKKrpAX

### API
`<ProgressCircular
	value={35}
	variant=”brand | success”
	assistiveText=””
/>`

https://docs.google.com/document/d/1mR6qYoDQN4eqU5lup-lywwOWH3suK599c_tZzEvpR8w/edit

### Test

Review: https://www.lightningdesignsystem.com/components/progress-ring/#site-main-content

PR #936, opened 2019-09-15
feat: implement progress circular component

wildergd:
fix: #547

- add ProgressCircular component

@nexxtway/tigger


src/components/ProgressCircular/__test__/progressRing.spec.js
```diff
@@ -0,0 +1,20 @@
...
+import ProgressRing from './../progressRing';
+
+describe('<ProgressRing/>', () => {
+    it('should have the right classnames', () => {
+        const component = shallow(<ProgressRing percent={25} />);
+        const container = component.find('div.rainbow-progress-circular_ring');
+        const svgRingPath = component.find('svg').find('path.rainbow-progress-circular_ring-path');
+        expect(container.exists()).toBe(true);
+        expect(svgRingPath.exists()).toBe(true);
+    });
```
LeandroTorresSicilia: in this case there is no need to test class names since they are fixed

src/components/ProgressCircular/index.js
```diff
@@ -0,0 +1,42 @@
...
+import classnames from 'classnames';
+import AssistiveText from './../AssistiveText';
+import normalizeValue from './normalizeValue';
+import ProgressRing from './progressRing';
+import './styles.css';
+
+/**
+ * ProgressCircular component communicates to the user the progress of a particular process.
+ */
+export default function ProgressCircular({ value, variant, assistiveText }) {
```
LeandroTorresSicilia: all components must have a className and style props
also when there is more than two props passed instead use props as argument and then make destructuring in first line inside the function as:
```
export default function ProgressCircular(props) {
    const { value, variant, assistiveText } = props;
```

src/components/ProgressCircular/index.js
```diff
@@ -0,0 +1,42 @@
...
+        classnames('rainbow-progress-circular', {
+            'rainbow-progress-circular--success': variant === 'success',
+        });
+
+    const normalizedValue = normalizeValue(value);
+
+    return (
+        <div className={getContainerClassNames()}>
+            <ProgressRing percent={normalizedValue} />
+            <span className="rainbow-progress-circular_percent-text">{`${value}%`}</span>
```
LeandroTorresSicilia: we should use normalizedValue here

src/components/ProgressCircular/normalizeValue.js
```diff
@@ -0,0 +1,8 @@
+export default function normalizeValue(value) {
```
LeandroTorresSicilia: add test for this function

Issue #967, opened 2019-09-18
Column header does not get updated


ighotouch:
In the table->column component, passing in a dynamic header does not get updated on component re-render.

-   [x ] I have searched the [issues](https://github.com/nexxtway/react-rainbow/issues) of this repository and believe that this is not a duplicate.

## Describe the bug

`          
const {counter} = this.state;
     <Column
                    header={counter}
                    field="name"
                  />`

## To Reproduce 🕹

1. Create a table component with one or two columns having header set from state;
2. Try updating the state value

## Expected Behavior 🤔

Header should be updated when state value changes

## Current Behavior 😯
Header is not updated

TahimiLeonBravo:
@ighotouch Thanks, we have fixed the issue, it will be available in the next release

PR #971, opened 2019-09-19
fix: fix column header does not get updated

wildergd:
fix: #967 

- fix column header does not get updated issue

@nexxtway/react-rainbow


src/components/Table/helpers/columns/isNotSameColumns.js
```diff
@@ -2,5 +2,9 @@ export default function isNotSameColumns(prevColumns, currentColumns) {
     if (prevColumns.length !== currentColumns.length) {
         return true;
     }
-    return prevColumns.some((column, index) => column.field !== currentColumns[index].field);
+    return prevColumns.some(
+        (column, index) =>
+            column.field !== currentColumns[index].field ||
+            column.header !== currentColumns[index].header,
```
LeandroTorresSicilia: we should add all other column props to make them reactive too

src/components/Table/helpers/columns/isNotSameColumns.js
```diff
@@ -2,5 +2,13 @@ export default function isNotSameColumns(prevColumns, currentColumns) {
...
     }
-    return prevColumns.some((column, index) => column.field !== currentColumns[index].field);
+    return prevColumns.some(
+        (column, index) =>
+            column.field !== currentColumns[index].field ||
+            column.type !== currentColumns[index].type ||
+            column.component !== currentColumns[index].component ||
+            column.defaultWidth !== currentColumns[index].defaultWidth ||
+            column.component !== currentColumns[index].component ||
+            column.width !== currentColumns[index].width,
```
LeandroTorresSicilia: use:
```
const {
    field,
    type,
    ...
} = currentColumns[index];
```
and add test in isNotSameColumns.spec.js for all cases added

src/components/Table/helpers/columns/__test__/isNotSameColumns.spec.js
```diff
@@ -7,18 +7,93 @@ describe('isNotSameColumns', () => {
...
+                defaultWidth: 100,
+                width: 'auto',
+            },
+        ];
+        const currentColumns = [
+            { field: 'phone', header: 'Name' },
+            { field: 'email', header: 'Email' },
+        ];
+        expect(isNotSameColumns(prevColumns, currentColumns)).toBe(true);
+    });
```
LeandroTorresSicilia: this test is the same below so remove this

src/components/Table/helpers/columns/__test__/isNotSameColumns.spec.js
```diff
@@ -7,18 +7,93 @@ describe('isNotSameColumns', () => {
...
+        const currentColumns = [
+            { field: 'name', header: 'Name', defaultWidth: 80 },
+            { field: 'email', header: 'Email', defaultWidth: 100 },
+        ];
+        expect(isNotSameColumns(prevColumns, currentColumns)).toBe(true);
+    });
+    it('should return true when column type property changed', () => {
+        const prevColumns = [
+            { field: 'name', header: 'Name', type: 'string' },
+            { field: 'email', header: 'Email', type: 'string' },
```
LeandroTorresSicilia: better use real values action and text

src/components/Table/helpers/columns/__test__/isNotSameColumns.spec.js
```diff
@@ -7,18 +7,93 @@ describe('isNotSameColumns', () => {
...
+        ];
         const currentColumns = [
-            { field: 'phone', label: 'Name' },
-            { field: 'email', label: 'Email' },
+            { field: 'name', header: 'Name', type: 'text' },
+            { field: 'email', header: 'Email', type: 'string' },
         ];
         expect(isNotSameColumns(prevColumns, currentColumns)).toBe(true);
     });
```
LeandroTorresSicilia: you forget test case for width and sortable

src/components/Table/helpers/columns/isNotSameColumns.js
```diff
@@ -2,5 +2,15 @@ export default function isNotSameColumns(prevColumns, currentColumns) {
...
-    return prevColumns.some((column, index) => column.field !== currentColumns[index].field);
+    return prevColumns.some((column, index) => {
+        const { field, type, header, component, defaultWidth, width } = currentColumns[index];
+        return (
+            column.field !== field ||
+            column.type !== type ||
+            column.header !== header ||
+            column.defaultWidth !== defaultWidth ||
+            column.component !== component ||
+            column.width !== width
```
LeandroTorresSicilia: we need add all column props, you forget sortable

src/components/Table/helpers/columns/__test__/isNotSameColumns.spec.js
```diff
@@ -7,18 +7,104 @@ describe('isNotSameColumns', () => {
...
+                type: 'string',
+                defaultWidth: 100,
+                width: 'auto',
+            },
+        ];
+        const currentColumns = [
+            { field: 'phone', header: 'Name' },
+            { field: 'email', header: 'Email' },
+        ];
+        expect(isNotSameColumns(prevColumns, currentColumns)).toBe(true);
```
LeandroTorresSicilia: I said that remove this because it have other props prevColumns not present in currentColumns then there are others props that will cause the function to return true so it is not exact
Then remove this and leave the one removed that is similar to below tests with:
```
const prevColumns = [
            { field: 'name', header: 'Name' },
            { field: 'email', header: 'Email' },
        ];
        const currentColumns = [
            { field: 'phone', header: 'Name' },
            { field: 'email', header: 'Email' },
        ];
```

Issue #950, opened 2019-09-17
feat: migrate styled-components


gustavopch:
I think this lib would benefit from using styled-components instead of SCSS.

-   [x] I have searched the [issues](https://github.com/nexxtway/react-rainbow/issues) of this repository and believe that this is not a duplicate.

**Is your feature request related to a problem? Please describe.**
No, not related to a problem. It's an improvement.

## Expected Behavior 🤔
Not about behavior but architecture.

## Context 🔦
By using styled-components, this lib would be able to:
- Stop having duplicated CSS and SCSS files everywhere.
- Remove className helper functions from components (`getClassNames`, `getVariantClassNames` etc) as you can inject props into styled-components.
- Stop worrying about vendor prefixes.
- Stop worrying about naming conflicts (I know BEM solves that problem, but it's even better when a problem doesn't even exist in the first place).
- Better theming support as styled-components generates CSS dynamically and theming is supported out-of-the-box with Context API.
- By the way, no need to build anything.
- Basically, a lot less lines of code which generally means less room for bugs and more productivity.

**Describe the solution you'd like**
Basically replacing CSS/SCSS by styled-components.

**Describe alternatives you've considered**
I haven't considered other alternatives.

reiniergs:
styled-components are pretty cool, that's something we should consider, at least new components should use it, replacing CSS/SCSS by styled-components is a major refactor in our codebase at this point.

LeandroTorresSicilia:
We will refactor all components to use styled components here the first one:
https://github.com/nexxtway/react-rainbow/pull/981

TahimiLeonBravo:
#### Components list to migrate
- [ ] Accordion
- [ ] AccordionSection
- [ ] ActivityTimeline
- [ ] Breadcrumb
- [ ] Breadcrumbs
- [ ] Card
- [ ] Modal
- [ ] Pagination
- [ ] Sidebar
- [ ] SidebarItem
- [ ] Tab
- [ ] Tabset
- [ ] TimelineMarker
- [ ] VerticalItem
- [ ] VerticalNavigation
- [ ] VerticalSection
- [ ] VerticalSectionOverflow
- [ ] CheckboxGroup
- [ ] CheckboxToogle
- [ ] DatePicker
- [ ] GoogleAddressLookup
- [ ] Input
- [ ] Lookup
- [ ] Picklist
- [ ] PicklistOption
- [ ] RadioButtonGroup
- [ ] RadioGroup
[...]

PR #981, opened 2019-09-21
refactor: use styled components in Badge

LeandroTorresSicilia:
- use styled components in Badge

@nexxtway/react-rainbow


src/components/Badge/truncate.js
```diff
@@ -0,0 +1,9 @@
+import styled from 'styled-components';
+
+const Truncate = styled.span`
```
reiniergs: TruncatedText??

gustavopch:
In case you don't know, you don't need to import those theme variables. Instead, you could do something like:
```
color: ${props => props.theme.COLOR_GRAY_4}
```
Of course you'd have to change the Application component to have a `ThemeProvider`. Then, when lib is fully refactored with styled-components, it'll be a matter of changing Application to take a `theme` prop in order to let each consumer customize their themes or implement a dark mode.

LeandroTorresSicilia:
@gustavopch 
That's right, but I think first we should move forward refactoring all components to use styled-components and removing all scss files.
For theming we need to discuss its structure and what will be customizable so it is early to use:
```props.theme.COLOR_GRAY_4 ```
although I know that we will need to touch all components again

TahimiLeonBravo:
#950

Issue #942, opened 2019-09-16
feat: TypeScript support


itskibo:
-   [x] I have searched the [issues](https://github.com/nexxtway/react-rainbow/issues) of this repository and believe that this is not a duplicate.

**Is your feature request related to a problem? Please describe.**
TypeScript is in more demand these days, and rapidly being adopted. Especially in professional environments, type safety is a large issue that's being thought of more these days (e.g. PHP looking at it with 7.x).
This library seems interesting to me, and I want to ensure that I can properly use it with personal projects.

## Expected Behavior 🤔
Correct typings when importing components.

## Context 🔦
I feel like the context is pretty clear: type safety for those who want it.

One could argue to add the definitions to DefinitelyTyped, but then we're maintaining both this repository AND the types over at DT.

**Describe the solution you'd like**
TypeScript support.

**Describe alternatives you've considered**
None.

gustavopch:
TS support is a make-or-break deal for a lot of projects.

rpmonteiro:
Exactly my situation... this library looks amazing, but until TS support is here, it can't be used.

gustavopch:
Refactoring to TS should probably also come with:
- Use named modules wherever possible (see https://basarat.gitbooks.io/typescript/docs/tips/defaultIsBad.html).
- Use functional components and Hooks wherever possible as they are easier to type (not to mention a bunch of other benefits).

reiniergs:
add TS is a high priority now since I will increase adoption :)

gustavopch:
> add TS is a high priority now since I will increase adoption :)

Would you accept a PR to refactor the whole code base to TS? I may give it a try.

reiniergs:
that's the whole purpose of OSS right? please go for it!

gustavopch:
@reiniergs Great! Just a problem: I cloned the repo and ran `yarn test`, but I'm getting a lot (hundreds I guess) of failed tests. Is that right?

reiniergs:
CI is passing, Jest test should be all passing locally ... I have no idea.

gustavopch:
Installed Node.js v11.8.0 to match the CI version, but the tests still fail. Mostly related to prop types, like: ``Warning: Failed prop type: The prop `label` is marked as required in `Lookup`, but its value is `undefined`.``
No idea what to do. I'm using Arch Linux BTW.

LeandroTorresSicilia:
@gustavopch 
[...]

PR #1012, opened 2019-09-25
feat: define typescript components interface

TahimiLeonBravo:
#942 

- define typescript components interface

@nexxtway/react-rainbow


src/components/Badge/index.d.ts
```diff
@@ -1,9 +1,7 @@
-export interface BadgeProps {
+export interface BadgeProps extends BaseProps {
```
LeandroTorresSicilia: BadgeProps?

src/components/Badge/index.d.ts
```diff
@@ -1,9 +1,7 @@
...
     label?: string | JSX.ElementChildrenAttribute;
     title?: string;
     variant?: 'default' | 'inverse' | 'lightest' | 'outline-brand' | 'brand';
-    className?: string;
-    style?: CSS.style;
 }
 
 export default function(props: BadgeProps): JSX.Element | null;
```
LeandroTorresSicilia: same ^^

stephen-zhao:
I think `export default` should be avoided. See: https://basarat.gitbooks.io/typescript/docs/tips/defaultIsBad.html

What do you think?

src/components/Button/index.d.ts
```diff
@@ -0,0 +1,34 @@
+import React, { Component } from 'react';
+import { BaseProps } from './../types';
+
+export interface ButtonProps extends BaseProps {
+    label?: string | JSX.ElementChildrenAttribute;
+    variant?:
+        | 'base'
```
stephen-zhao: that's cool, didn't know leading operators were a thing 😁

src/components/ButtonIcon/index.d.ts
```diff
@@ -0,0 +1,34 @@
...
+    disabled?: boolean;
+    form?: string;
+    icon: JSX.ElementChildrenAttribute;
+    id?: string;
+    // onBlur?: ;
+    // onClick?: ;
+    // onFocus?: ;
+    // onKeyDown?: ;
+    shaded?: boolean;
+    size?: 'xx-small' | 'x-small' | 'small' | 'medium' | 'large';
```
stephen-zhao: Would there be any benefit of type-aliasing these union types to share across files? I noticed there is a lot of repeat. I can see maintainability issues if we need to rename or add a new size in the future...

Type aliases are resolved before showing up in IDEs like VS Code, so they won't affect clients' workflow.

reiniergs:
@stephen-zhao agree with you about `export default`, but at this point, the library expects the components exported as default in `index.js` files and the declaration files need to mock the component implementation, basically there is not much we can do :)

Issue #962, opened 2019-09-18
fix: improve accessibility test in components


TahimiLeonBravo:
We need to review all components

### Components list
- [x] Accordion
- [x] AccordionSection
- [x] ActivityTimeline
- [x] Breadcrumb
- [x] Breadcrumbs
- [x] Card
- [x] Modal
- [x] Pagination
- [x] Sidebar
- [x] SidebarItem
- [x] Tab
- [x] Tabset
- [x] TimelineMarker
- [x] VerticalItem
- [x] VerticalNavigation
- [x] VerticalSection
- [x] VerticalSectionOverflow
- [x] CheckboxGroup
- [x] CheckboxToogle
- [x] DatePicker
- [x] GoogleAddressLookup
- [x] Input
- [x] Lookup
- [x] Picklist
- [x] PicklistOption
- [x] RadioButtonGroup
- [x] RadioGroup
- [x] Rating
- [x] Select
- [x] Slider
- [x] Textarea
- [x] TimePicker
- [x] VisualPicker
- [x] VisualPickerOption
- [x] VisualPickerOptionFooter
- [x] Avatar
- [x] AvatarMenu
- [x] Badge
- [x] Button
- [x] ButtonGroup
- [x] ButtonIcon
- [x] ButtonMenu
- [x] Calendar
- [x] Carousel
- [x] CarouselImage
- [x] Chip
- [x] GMap
- [x] MapMarker
- [x] MenuDivider
- [x] MenuItem
- [x] Notification
- [x] ProgressBar
- [x] ProgressCircular
- [x] ProgressIndicator
- [x] ProgressStep
- [x] ReCaptcha
- [x] Spinner
- [x] Chart
- [x] Column
- [x] Dataset
- [x] Table
- [x] TableWithBrowserPagination
- [x] ImportRecordsFlow

PR #1039, opened 2019-09-28
fix: fix accessibility on Picklist component

cmcartaya:
#962 

- Fix accessibility on Picklist component.

@nexxtway/react-rainbow

### DIFF
--- a/src/components/Picklist/index.js
+++ b/src/components/Picklist/index.js
@@ -1,4 +1,3 @@
-/* eslint-disable jsx-a11y/role-has-required-aria-props */
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
@@ -363,8 +362,6 @@ class Picklist extends Component {
             hideLabel,
             style,
             error,
-            title,
-            assistiveText,
             isLoading,
             disabled,
             readOnly,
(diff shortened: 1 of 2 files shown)


src/components/Picklist/index.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable jsx-a11y/role-has-required-aria-props */
```
LeandroTorresSicilia: use in the line instead because this will not alert us if there is another error in the file

src/components/Picklist/index.js
```diff
@@ -403,14 +405,20 @@ class Picklist extends Component {
...
                     <RenderIf isTrue={!!icon}>
                         <span className="rainbow-picklist_icon">{icon}</span>
                     </RenderIf>
                     <RenderIf isTrue={!readOnly}>
                         <span className={this.getIndicatorClassNames()} />
                     </RenderIf>
                     <input
+                        aria-controls={this.listboxId}
                         className={this.getInputClassNames()}
                         id={this.inputId}
```
LeandroTorresSicilia: we should use role="textbox"?

src/components/Picklist/index.js
```diff
@@ -428,7 +436,11 @@ class Picklist extends Component {
                         autoComplete="off"
                         ref={this.triggerRef}
                     />
-                    <div role="listbox" className={this.getDropdownClassNames()}>
+                    <div
+                        id={this.listboxId}
+                        role="listbox"
+                        className={this.getDropdownClassNames()}
+                    >
```
LeandroTorresSicilia: review whether is necessary ```aria-label={ariaLabel}``` below in ul element

PR #1049, opened 2019-10-02
fix: improve accessibility on Picklist component

cmcartaya:
#962 

- Improve accessibility on Picklist component adding some aria attr

@nexxtway/react-rainbow


src/components/Picklist/index.js
```diff
@@ -176,6 +183,9 @@ class Picklist extends Component {
         const { onChange } = this.props;
         const { activeOptionIndex } = this.state;
         const { label, name, icon, value } = this.activeChildren[activeOptionIndex];
+        this.setState({
+            selectedOptionName: name,
+        });
```
LeandroTorresSicilia: we can't do this since there is two source of true, the only one that we need to use is value prop so there is not need to add other state for this

src/components/Picklist/index.js
```diff
@@ -142,6 +144,11 @@ class Picklist extends Component {
         return undefined;
     }
 
+    getAriaActivedescendant() {
+        const { selectedOptionName, activeOptionName } = this.state;
+        return activeOptionName || selectedOptionName;
```
LeandroTorresSicilia: here you can check whether there is not activeOptionName and then get the AriaActivedescendant form the value prop.
Also you need to figure out what take precedence in case when there is an option selected an an item has visual focus in this case you are giving priority to the option with visual focus

Issue #970, opened 2019-09-19
feat: implement datetime component.


jerlyn06:
Hello! 

Looking to adapt react-rainbow on an enterprise project, and it's almost perfect except for a datetime picker component, that would essentially merge the individual date picker and the time picker.   

Is this something that is already available but I seem to have missed or something in the works?

reiniergs:
having two fields (data & time) do not work for your use case? although, it will be a good addition to the library :)

TahimiLeonBravo:
Hi,
@jerlyn06 We have designed a proposal for the DateTimePicker component.
Something like this is what you are looking for?
![Components DateTimePicker modal](https://user-images.githubusercontent.com/32083556/65275303-ff2f4080-daea-11e9-9eae-21ddad703ba7.png)
![Components DateTimePicker modal in mobile](https://user-images.githubusercontent.com/32083556/65275519-78c72e80-daeb-11e9-91e2-54f12588dc2d.png)

jerlyn06:
> having two fields (data & time) do not work for your use case? although, it will be a good addition to the library :)

Unfortunately not very well. Thanks for taking this into consideration!

jerlyn06:
> Hi,
> @jerlyn06 We have designed a proposal for the DateTimePicker component.
> Something like this is you looking for?
> 
> ![Components DateTimePicker prop](https://user-images.githubusercontent.com/32083556/65275621-a9a76380-daeb-11e9-87d0-a76840902c21.png)
> 
> This is one of the next components to implement in our roadmap

This is EXACTLY what I'm looking for, thank you. I would offer to contribute but I'm still learning React.

TahimiLeonBravo:
@jerlyn06 
Thank you for the suggestion
We should be ready a beta version for the component on the next week.
Anyway you are welcome to contribute 😉

TahimiLeonBravo:
@wildergd 
Links to designs in Zeplin 
https://zpl.io/bzL4koA
https://zpl.io/aRXZrdp
https://zpl.io/2pP5yA8

TahimiLeonBravo:
@wildergd when we add the integration test for the component we can close the issue

PR #1114, opened 2019-10-25
test: add integration tests to DateTimePicker component

wildergd:
fix: #970 

- add integration tests to DateTimePicker  component

@nexxtway/react-rainbow

### DIFF
--- a/integration/specs/DateTimePicker/datetimepicker-1.spec.js
+++ b/integration/specs/DateTimePicker/datetimepicker-1.spec.js
@@ -15,8 +15,8 @@ describe('DateTimePicker', () => {
         const picker = new PageDateTimePicker(DATETIMEPICKER);
         picker.click();
         picker.waitUntilModalIsOpen();
-        expect(picker.getModalHeaderText()).toBe('10/25/19, 2:44 PM');
+        expect(picker.getModalHeaderText()).toBe('10/25/19, 10:44 PM');
         picker.clickDay(10);
-        expect(picker.getModalHeaderText()).toBe('10/10/19, 2:44 PM');
+        expect(picker.getModalHeaderText()).toBe('10/10/19, 10:44 PM');
     });
 });
--- a/integration/specs/DateTimePicker/datetimepicker-1.spec.js
+++ b/integration/specs/DateTimePicker/datetimepicker-1.spec.js
@@ -17,6 +17,6 @@ describe('DateTimePicker', () => {
         picker.waitUntilModalIsOpen();
         expect(picker.getModalHeaderText()).toBe('10/25/19, 2:44 PM');
         picker.clickDay(10);
-        expect(picker.getTimeValue()).toBe('10/10/19, 2:44 PM');
+        expect(picker.getModalHeaderText()).toBe('10/10/19, 2:44 PM');
     });
 });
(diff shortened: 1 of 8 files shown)


integration/specs/DateTimePicker/datetimepicker-1.spec.js
```diff
@@ -0,0 +1,22 @@
...
+    });
+    beforeEach(() => {
+        browser.refresh();
+        const component = $(DATETIMEPICKER);
+        component.waitForExist();
+    });
+    it('should set correct date on modal header when day is clicked', () => {
+        const picker = new PageDateTimePicker(DATETIMEPICKER);
+        picker.click();
+        picker.waitUntilModalIsOpen();
```
LeandroTorresSicilia: let's be consistent like we did in TimePicker page object so call this method waitUntilOpen

integration/specs/DateTimePicker/datetimepicker-1.spec.js
```diff
@@ -0,0 +1,22 @@
...
+        expect(picker.getModalHeaderText()).toBe('10/25/19, 10:44 AM');
+        picker.clickDay(10);
+        expect(picker.getModalHeaderText()).toBe('10/10/19, 10:44 AM');
```
LeandroTorresSicilia: Remove this test and the getModalHeaderText method in the page object since the time is not changed we don't know if we will maintain this.
Then add a test for:
```should set the right value when select a date```
create a method getValue in the page object that will return the input value so you can check that the right date time is selected, for now we will not implement a method for set the time, so left it as 12:00

File: integration/specs/DateTimePicker/datetimepicker-1.spec.js [hunk 3]
```diff
@@ -0,0 +1,22 @@
...
+    });
+});
```
LeandroTorresSicilia: also add test cases for:
- should open the DateTimePicker when click on input element
- should open the DateTimePicker when click on the label element (for this create a clickLabel method in page object)
- should close the DateTimePicker when click the ok button (for close test cases you need to add waitUntilClose method in page object)
- should close the DateTimePicker when click the cancel button

we don't need to repeat the modal tests but this above does not belong to modal but to DateTimePicker itself

src/components/DateTimePicker/index.js
```diff
@@ -38,6 +38,7 @@ function PickerComponent(props, ref) {
     } = props;
 
     const inputRef = useRef();
+    const modalId = (id && `${id}_modal`) || undefined;
```
LeandroTorresSicilia: here you only need:
```const modalId = id && `${id}_modal`;```
when id is undefined it will return the id value (undefined)
otherwise will return  `${id}_modal`

File: src/components/DateTimePicker/pageObject/index.js [hunk 5]
```diff
@@ -0,0 +1,84 @@
...
+     * @constructor
+     * @param {string} rootElement - The selector of the PageDateTimePicker root element.
+     */
+    constructor(rootElement) {
+        this.rootElement = rootElement;
+        this.modalRootEl = `${rootElement}_modal`;
+    }
+
+    /**
+     * Returns the time selected element.
```
LeandroTorresSicilia: Returns the selected time value

src/components/DateTimePicker/pageObject/index.js
```diff
@@ -0,0 +1,84 @@
...
+    /**
+     * Returns the text of the modal header element.
+     * @method
+     * @returns {string}
+     */
+    getModalHeaderText() {
+        return $(this.modalRootEl)
+            .$('header > h2')
+            .getText();
+    }
```
LeandroTorresSicilia: remove this method since it will not be used right now

Issue #1108, opened 2019-10-22
feat: Modal close button


FouMez:
Hello ! 
Can you please add the possibility to hide the close button of a modal, i'm doing it manually by overriding the CSS class but it's not very effective to do so. 

Thank you for this great project !

TahimiLeonBravo:
Hi @FouMez 
Thanks for the suggestion,
Please, we would like to know a little more about the use case you have to understand why you need to hide the close button on the Modal.

FouMez:
Hi @TahimiLeonBravo 
sometimes i use modals to block the access to some content of my apps , for example if the user have an unverified email or have some payment issues, i will block all app features with a modal that cannot be closed.

TahimiLeonBravo:
Hi @FouMez
There is a new version 1.8.0 where we add a new prop `hideCloseButton` to hide the close button of a modal.
This prop was added in the pr #1115.
Thank for the feedback.

FouMez:
Thank you so much for being reactive. 
Keep going !

PR #1115, opened 2019-10-25
feat: add hideCloseButton prop to modal component

mycodeself:
feat: add hideCloseButton prop to modal component

- Added to modal component a new optional prop called hideCloseButton, this allows to control whether the component for close the modal should render or not
- The necessary tests have been added

Issue related #1108 

@nexxtway/react-rainbow

### DIFF
--- a/src/components/Modal/__test__/modal.spec.js
+++ b/src/components/Modal/__test__/modal.spec.js
@@ -187,4 +187,24 @@ describe('<Modal/>', () => {
         component.unmount();
         expect(CounterManager.decrement).not.toHaveBeenCalled();
     });
+    it('should show the cross for close the modal by default', () => {
+        const component = mount(
+            <Modal isOpen>
+                <p />
+            </Modal>,
+        );
+
+        expect(component.find('button[data-id="button-icon-element"]').prop('id')).toBe(
+            'modal-close-button',
+        );
+    });
+    it('should not render the cross for close the modal when the showClose prop is equal to false', () => {
+        const component = mount(
+            <Modal isOpen showClose={false}>
+                <p />
+            </Modal>,
+        );
+
+        expect(component.exists('button[data-id="button-icon-element"]')).toBeFalsy();
+    });
 });
--- a/src/components/Modal/__test__/modal.spec.js
+++ b/src/components/Modal/__test__/modal.spec.js
@@ -195,7 +195,7 @@ describe('<Modal/>', () => {
             </Modal>,
         );
 
-        expect(component.find(StyledCloseButton).exists).toBeTruthy();
+        expect(component.find(StyledCloseButton).exists()).toBe(true);
     });
     it('should not render the close button when the hideCloseButton prop is equal to true', () => {
         const component = mount(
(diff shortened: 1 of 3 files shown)


src/components/Modal/index.js
```diff
@@ -188,6 +200,8 @@ Modal.propTypes = {
      * @ignore
      */
     children: PropTypes.node,
+    /** Determines if the cross for close is showed or not */
+    showClose: PropTypes.bool,
```
LeandroTorresSicilia: the most use cases will use the close button so instead of use showClose I think it is better to use hideCloseButton to be more explicit, also this way is more consistent with what we do in Notification component. Then default value would be:
```hideCloseButton: false,```
So let's change the description to:
```If true, hide the close button in the modal```
as we did in Notification

File: src/components/Modal/index.js [hunk 2]
```diff
@@ -133,13 +143,15 @@ export default class Modal extends Component {
...
                         size={size}
                     >
-                        <StyledCloseButton
-                            id="modal-close-button"
-                            icon={<CloseIcon />}
-                            title="Close"
-                            onClick={this.closeModal}
-                            ref={this.buttonRef}
-                        />
+                        {showClose && (
```
LeandroTorresSicilia: we have an internal component, RenderIf for this cases, then instead use:
```
<RenderIf isTrue={!hideCloseButton}>
    <StyledCloseButton ... />
</RenderIf>
```

src/components/Modal/__test__/modal.spec.js
```diff
@@ -187,4 +187,24 @@ describe('<Modal/>', () => {
...
+    it('should show the cross for close the modal by default', () => {
+        const component = mount(
+            <Modal isOpen>
+                <p />
+            </Modal>,
+        );
+
+        expect(component.find('button[data-id="button-icon-element"]').prop('id')).toBe(
+            'modal-close-button',
+        );
```
LeandroTorresSicilia: you can simply:
```
import StyledCloseButton from '../styled/closeButton';
```
and then:
```expect(component.find(StyledCloseButton).exists()).toBe(true);```
same for the case below since ```button[data-id="button-icon-element"]``` is an internal implementation of ButtonIcon and this implies an introspection that we need to avoid.

Comment (mycodeself) [reply src/components/Modal/__test__/modal.spec.js]: I have added this suggestion, thank you very much

File: src/components/Modal/index.js (hunk 2 above)
Comment (mycodeself) [reply src/components/Modal/index.js]: I have added this suggestion, thank you very much

File: src/components/Modal/index.js (hunk 1 above)
Comment (mycodeself) [reply src/components/Modal/index.js]: I have added this suggestion, thank you very much!

File: src/components/Modal/__test__/modal.spec.js [hunk 4]
```diff
@@ -187,4 +188,22 @@ describe('<Modal/>', () => {
...
         expect(CounterManager.decrement).not.toHaveBeenCalled();
     });
+    it('should render the close button by default', () => {
+        const component = mount(
+            <Modal isOpen>
+                <p />
+            </Modal>,
+        );
+
+        expect(component.find(StyledCloseButton).exists()).toBe(true);
```
LeandroTorresSicilia: I changed this because when you said:
```expect(component.find(StyledCloseButton).exists).toBeTruthy();```
exists is a method in the component.find(StyledCloseButton) object and then you are accessing to the method declaration that is truthy instead of calling it

LeandroTorresSicilia:
great job due!!!
we are happy for your contribution

mycodeself:
Thank you for all the tips

Issue #848, opened 2019-07-22
chore: add an eslint rule to complains about defaultProps that are not defined in prop types.


TahimiLeonBravo:
```
MyComponent.propTypes = {
    style: PropTypes.object,
}
```
and
```
MyComponent.defaultProps = {
    style: undefined,
    isOpen: false,
}
```
### Current behavior: 
In this example, isOpen is not defined in propTypes and currently, eslint does not complain about this.
### Expected behavior:
eslint should complain when a prop is defined in defaultProps and it is not in propTypes.

PR #1116, opened 2019-10-25
chore: add eslint defaultProps not defined

mycodeself:
chore: add eslint rule for throw error for optional props that are not defined in defaultProps

- Enforce a defaultProps definition for every prop that is not a required prop (https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/require-default-props.md)

Related issue #848 

@nexxtway/react-rainbow

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -11,6 +11,7 @@
         "react/jsx-first-prop-new-line": "off",
         "react/jsx-closing-bracket-location": "off",
         "react/forbid-prop-types": "off",
+        "react/require-default-props": 2,
         "no-mixed-operators": "off",
         "arrow-parens": ["error", "as-needed"]
     },


LeandroTorresSicilia:
Doesn't work this for you by default?
For me it work without explicit setting this rule.

mycodeself:
Yep, you are right, it works without explicit set the rule.

Im sorry, so i dont understand the issue #848 

Regards!

LeandroTorresSicilia:
Sorry, the issue is unclear.
It is about props that are defined in defaultProps and not in propTypes like:

```
MyComponent.propTypes = {
    style: PropTypes.object,
}
```
and
```
MyComponent.defaultProps = {
    style: undefined,
    isOpen: false,
}
```
as you see in this example isOpen is not defined in propTypes and currently, eslint does not complain about this

Issue #1111, opened 2019-10-22
fix: library for support components wrapped with React.forwardRef


TahimiLeonBravo:
it should be fixed if we update React Styleguide

PR #1118, opened 2019-10-28
fix: library for support components wrapped with React.forwardRef

cmcartaya:
fix: #1111 

- Fix library for support components wrapped with React.forwardRef.

@nexxtway/react-rainbow


library/styleguideComponents/SectionRenderer/index.js
```diff
@@ -1,20 +1,11 @@
 /* eslint-disable import/no-extraneous-dependencies */
 import React from 'react';
 import PropTypes from 'prop-types';
-import Markdown from 'react-styleguidist/lib/rsg-components/Markdown/Markdown';
+// import Markdown from 'react-styleguidist/lib/rsg-components/Markdown/Markdown';
```
LeandroTorresSicilia: remove if not used

library/styleguideComponents/SectionRenderer/index.js
```diff
@@ -24,12 +15,14 @@ export default function SectionRenderer(props) {
...
                     pagePerSection={pagePerSection}
-                    slotProps={props}>
+                    slotProps={props}
+                >
                     {name}
                 </SectionHeading>
             )}
             <div>
-                {description && <Markdown text={description} />}
+                {/* {description && <Markdown text={description} />} */}
```
LeandroTorresSicilia: same ^^ remove the comment

src/components/AccordionSection/readme.md
```diff
@@ -1,6 +1,11 @@
 ##### accordion base
 
 ```js
+import React from 'react';
+import Card from 'react-rainbow-components/components/Card';
+import Accordion from 'react-rainbow-components/components/Accordion';
+import AccordionSection from 'react-rainbow-components/components/AccordionSection';
```
LeandroTorresSicilia: this is needed? I think that what is needed to import was components like GlobalHeader

src/components/AccordionSection/readme.md
```diff
@@ -21,5 +26,5 @@
             </AccordionSection>
         </Accordion>
     </Card>
-</div>
+</div>;
```
LeandroTorresSicilia: remove ;

src/components/AccordionSection/readme.md
cmcartaya: So sorry, I was forgetting to communicate with you for the new feature and breaking change of the react-styleguidist 9 version. For now no more global component, only the current component is accessible in the example context, or you can explicitly import everything, to make examples easier to copy into your app code.

styleguide.config.js
```diff
@@ -19,6 +19,9 @@ if (env.parsed && !env.error) {
 }
 
 module.exports = {
+    moduleAliases: {
+        'react-rainbow-components': path.resolve(__dirname, './src'),
```
LeandroTorresSicilia: I think we should use:
```path.resolve(__dirname, './src/components'),```
since in the examples right now you are using:
```import { Accordion } from 'react-rainbow-components/components';```
and this is confuse since the real usage is:
```import { Accordion } from 'react-rainbow-components';```

File: src/components/Modal/readme.md [hunk 6]
```diff
@@ -468,7 +468,17 @@ class ModalWSize extends React.Component {
...
+    Button,
+    Input,
+    TimePicker,
+    DatePicker,
+    Lookup,
+    Textarea,
+} from 'react-rainbow-components/components';
 
 const { Field, reduxForm } = require('redux-form');
 const { FontAwesomeIcon } = require('@fortawesome/react-fontawesome');
```
LeandroTorresSicilia: this need to be changed to use ES imports since can't be import and require together
change it in all examples

Issue #646, opened 2019-04-02
feat: implement keyboard-navigation in Calendar component


TahimiLeonBravo:
### Spec
We should follow the w3 specifications
https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/datepicker-dialog.html

#### Note:
I think that also should review that:
https://developer.salesforce.com/docs/component-library/bundle/lightning:input/example#lightningcomponentdemo:exampleInputDate
https://www.lightningdesignsystem.com/components/datepickers/#site-main-content

TahimiLeonBravo:
fix: #1141

PR #1122, opened 2019-11-03
feat: implement keyboard-navigation in Calendar component

wildergd:
fix: #646 

- add arrow keys navigation to Calendar component

@nexxtway/react-rainbow


LeandroTorresSicilia:
@TahimiLeonBravo  there is an issue that needs to be addressed:
Tabbing through the Calendar component there is no visual change when the selected day gets focus due to current styles.

src/components/Calendar/helpers/compareDates.js
```diff
@@ -0,0 +1,5 @@
+export default function compareDates(value1, value2) {
+    const date1 = new Date(value1).setHours(0, 0, 0, 0);
+    const date2 = new Date(value2).setHours(0, 0, 0, 0);
+    return Math.sign(date1 - date2);
```
LeandroTorresSicilia: this is very clean but IE doesn't support Math.sign, since I saw that this is used again in order to not refactor too much I think you should create our own sign function like (getSign), don't forget test it.

src/components/Calendar/index.js
```diff
@@ -29,38 +45,136 @@ class CalendarComponent extends Component {
...
+    }
+
+    moveFocusedDay(increment) {
+        const { minDate, maxDate } = this.props;
+        const { currentMonth, focusedDate } = this.state;
+        let nextFocusedDate = addDays(focusedDate, increment);
+        let nextFocusedMonth = currentMonth;
+
+        if (nextFocusedDate.getMonth() !== currentMonth.getMonth()) {
+            nextFocusedMonth = getFirstDayMonth(addMonths(currentMonth, Math.sign(increment)));
```
LeandroTorresSicilia: same ^^ sign doesn't work in IE

src/components/Calendar/index.js
```diff
@@ -29,38 +45,136 @@ class CalendarComponent extends Component {
...
+        }
+
+        if (minDate && compareDates(nextFocusedDate, minDate) < 0) {
+            nextFocusedDate = minDate;
+            nextFocusedMonth = getFirstDayMonth(minDate);
+        } else if (maxDate && compareDates(nextFocusedDate, maxDate) > 0) {
```
LeandroTorresSicilia: here the check for minDate and maxDate are preventing to stop when you get to max and min boundaries when they are undefined (not passed minDate and maxDate props)

src/components/Calendar/index.js
```diff
@@ -29,38 +45,136 @@ class CalendarComponent extends Component {
...
+            currentMonth: nextFocusedMonth,
+        });
+    }
+
+    moveFocusedMonth(increment) {
+        const { minDate, maxDate } = this.props;
+        const { focusedDate } = this.state;
+        let nextFocusedDate = addMonths(focusedDate, increment);
+
+        if (minDate && compareDates(nextFocusedDate, minDate) < 0) {
```
LeandroTorresSicilia: let's create helpers with tests for this since they are used twice, also this will improve readability (isMinBound and isMaxBound or something similar)

src/components/Calendar/index.js
```diff
@@ -29,38 +45,136 @@ class CalendarComponent extends Component {
...
+            nextFocusedDate = minDate;
+        } else if (maxDate && compareDates(nextFocusedDate, maxDate) > 0) {
```
LeandroTorresSicilia: same ^^

src/components/Calendar/index.js
```diff
@@ -69,13 +183,94 @@ class CalendarComponent extends Component {
...
+    }
+
+    handleKeyHomePressed() {
+        const { focusedDate } = this.state;
+        this.moveFocusedDay(-focusedDate.getDay());
+    }
+
+    handleKeyEndPressed() {
+        const { focusedDate } = this.state;
+        this.moveFocusedDay(6 - focusedDate.getDay());
```
LeandroTorresSicilia: the combination fn + alt + right or left should do the same as fn + right or left, in other words the first combination should call this or the above function

src/components/Calendar/index.js
```diff
@@ -69,13 +183,94 @@ class CalendarComponent extends Component {
...
+    }
+
+    handleKeyPageUpPressed() {
+        this.moveFocusedMonth(-1);
```
LeandroTorresSicilia: there is a problem when try to change month or year when minDate and maxDate are not passed

src/components/Calendar/index.js
```diff
@@ -122,16 +317,26 @@ class CalendarComponent extends Component {
                         onChange={this.handleYearChange}
                     />
                 </StyledControlsContainer>
-                <table role="grid" aria-labelledby={this.monthLabelId}>
+                <StyledTable
+                    role="grid"
+                    ref={this.tableRef}
+                    aria-labelledby={this.monthLabelId}
+                    tabIndex="0"
```
LeandroTorresSicilia: The approach followed by the links showed in the issue don't do this, instead they set tabIndex="0" only to the day that will get the focus and all other are tabIndex="-1", then when the use the keyboard to move focus to another day then the new focused day gets tabIndex="0"  and previous get tabIndex="-1". This should solve the problem with the onClick. I think it is better to follow this approach because it is not accessible to make the table focusable since what is focusable is the day to select.

src/components/Calendar/index.js
```diff
@@ -29,38 +45,136 @@ class CalendarComponent extends Component {
...
         const normalizedDate = normalizeDate(value);
         if (formatDate(normalizeDate(prevValue)) !== formatDate(normalizedDate)) {
             this.updateCurrentMonth(normalizedDate);
+            this.updateFocusedDate(normalizedDate);
+        }
+
+        const { showFocusedDate: prevShowFocusedDate } = prevState;
+        const { showFocusedDate } = this.state;
+        if (showFocusedDate && !prevShowFocusedDate) {
+            this.tableRef.current.focus();
```
LeandroTorresSicilia: this should not be necessary removing tabIndex="0"  from the table

src/components/Calendar/index.js
```diff
@@ -69,13 +183,94 @@ class CalendarComponent extends Component {
...
+            showFocusedDate: true,
+        });
+    }
+
+    handleBlurDaysContainer() {
+        if (!this.state.showFocusedDate) return;
+        this.setState({
+            showFocusedDate: false,
+        });
+    }
```
LeandroTorresSicilia: this both (this method and the above) should not be needed when removing the tabindex in table, and maybe the showFocusedDate state also is needed, really I don't know this but review carefully

src/components/Calendar/index.js
```diff
@@ -69,13 +183,94 @@ class CalendarComponent extends Component {
...
+
+    handleChange(date) {
+        const { onChange } = this.props;
+        onChange(date);
+        setTimeout(() => {
+            this.tableRef.current.focus();
+        }, 0);
```
LeandroTorresSicilia: By removing tabIndex="0"  from the table this should not be needed since the focus will get in the button itself when select a day by clicking.

src/components/Calendar/styled/table.js
```diff
@@ -0,0 +1,9 @@
+import styled from 'styled-components';
+
+const StyledTable = styled.table`
+    &:focus {
+        outline: none;
```
LeandroTorresSicilia: this component should not be needed when removing tabindex from table

src/components/Calendar/helpers/__test__/getCalendarBounds.spec.js
```diff
@@ -0,0 +1,39 @@
+import getCalendarBounds from '../getCalendarBounds';
+
+const minDate = new Date(2018, 0, 1);
+const maxDate = new Date(2020, 7, 3);
+const today = new Date();
+const date1 = new Date(today.getFullYear() - 100, 0, 1);
+const date2 = new Date(today.getFullYear() + 100, 11, 31);
```
LeandroTorresSicilia: let's make this deterministic using fixed dates

src/components/Calendar/index.js
```diff
@@ -52,30 +171,119 @@ class CalendarComponent extends Component {
...
     handleYearChange(event) {
         const year = +event.target.value;
         const newMonth = new Date(this.state.currentMonth);
         newMonth.setFullYear(year);
+
+        const { value } = this.props;
+        const focusedDate =
+            isSameMonth(value, newMonth) && isSameYear(value, newMonth)
+                ? value
+                : getFirstDayMonth(newMonth);
```
LeandroTorresSicilia: this is repeated three times so use a helper for this (getNextFocusedDate) and test it

PR #1129, opened 2019-11-07
fix: fix insert child orderly on Picklist component

cmcartaya:
- Fix insert child orderly on Picklist component

@nexxtway/react-rainbow


src/components/Picklist/utils.js
```diff
@@ -18,7 +18,7 @@ export function insertChildOrderly(children, child, nodes) {
 
 export function getChildMenuItemNodes(ref) {
     if (ref) {
-        return ref.querySelectorAll('a[role="menuitem"]');
+        return ref.querySelectorAll('div[role="option"]');
```
LeandroTorresSicilia: Since we don't get aware of this bug we need to find somehow test this scenario.

integration/specs/Picklist/picklist-9.spec.js
```diff
@@ -0,0 +1,50 @@
...
+        const picklist = new PagePicklist(PICKLIST);
+        picklist.clickInput();
+        picklist.waitUntilOpen();
+        expect(picklist.getOptionsLength()).toBe(2);
+        browser.keys(ESCAPE_KEY);
+        addNewBuildings();
+        picklist.clickInput();
+        picklist.waitUntilOpen();
+        expect(picklist.getOptionsLength()).toBe(5);
+    });
```
LeandroTorresSicilia: this test does not test any component functionality, instead, it check the example code, and we don't want to add unnecessary tests,  so remove it

integration/specs/Picklist/picklist-9.spec.js
```diff
@@ -0,0 +1,50 @@
...
+        const picklist = new PagePicklist(PICKLIST);
+        addNewBuildings();
+        picklist.clickInput();
+        picklist.waitUntilOpen();
+        const option = picklist.getOption(4);
+        option.hover();
+        expect(option.isActive()).toBe(true);
+        option.click();
+        expect(picklist.getSelectedOptionLabel()).toBe('Empire State');
+    });
```
LeandroTorresSicilia: only make one test for this, remove the two above this one, for simplicity, I think it is better to add only one item above Empire state and then test the position of empire state before and after the item is added, you can get the option at specific index and check its label so the only one test needed is:
```
1. open picklist
2. check position 1 is Empire state
3. add new option
4. check position 1 is the new added item
5. check position 2 is Empire state
```

PR #1138, opened 2019-11-10
refactor: migrate Textarea to styled-components

TahimiLeonBravo:
#950 

- migrate Textarea to styled-components

@nexxtway/react-rainbow


src/components/Textarea/index.js
```diff
@@ -1,12 +1,14 @@
...
 import autosize from 'autosize';
 import withReduxForm from './../../libs/hocs/withReduxForm';
 import { uniqueId } from './../../libs/utils';
 import RenderIf from '../RenderIf';
 import Label from './label';
-import './styles.css';
+import StyledContainer from './styled/container';
+import StyledTextarea from './styled/textarea';
+import StyledBottomHelp from './styled/bottomHelp';
+import StyledError from './styled/error';
```
LeandroTorresSicilia: bottomHelp and error are already defined in input component so use these instead:
```input/styled/helpText```
```input/styled/errorText```
same for label, use this one:
```input/label```
also remove label and his test file and all other styeld files

File: src/components/Textarea/index.js [hunk 2]
```diff
@@ -113,9 +107,9 @@ class Textarea extends Component {
                     id={this.getInlineTextLabelId()}
```
LeandroTorresSicilia: here note that you must use inputId instead of textareaId

src/components/Textarea/index.js
```diff
@@ -103,7 +97,7 @@ class Textarea extends Component {
         } = this.props;
 
         return (
-            <div className={this.getContainerClassNames()} style={style} id={id}>
+            <StyledContainer className={className} error={error} style={style} id={id}>
```
LeandroTorresSicilia: if the error prop is not used in StyledContainer then don't pass it

Issue #1155, opened 2019-11-19
fix: add arrows to the Lookup when displaying many options


TahimiLeonBravo:
### Requirements
- [ ] Show arrow up/down functionality when are many options to show in the dropdown menu
- [ ] Add unit test for this functionality
- [ ] Add integration test

### Spec
- It should render an arrow down/up when are many options to show
- It should scrolling when the user hovers over the arrows

Please follow to the Picklist behavior for more details
https://react-rainbow.web.app/#!/Picklist/3

### Unit Test Spec
- [ ] should render scroll down arrow when the number of children is more than five
- [ ] should not render scroll down arrow when the number of children is less than five
- [ ] should not render scroll up arrow when menu scroll position is zero
- [ ] should render scroll up arrow when popup menu scroll position is greater than zero

### Design
https://zpl.io/aw4Algj

TahimiLeonBravo:
@rafaelmg83

rafaelmg83:
Ok, perfect, i started to work on it already.

PR #1169, opened 2019-11-27
feat: adding scroll arrows

rafaelmg83:
fix: #1155 

## Added scroll arrows to the Lookup component
-

@nexxtway/react-rainbow


src/components/Lookup/index.js
```diff
@@ -72,6 +75,9 @@ class Lookup extends Component {
             [DOWN_KEY]: this.handleKeyDownPressed,
             [ENTER_KEY]: this.handleKeyEnterPressed,
         };
+        this.handleScrollDownArrowHover = this.handleScrollDownArrowHover.bind(this);
+        this.handleScrollUpArrowHover = this.handleScrollUpArrowHover.bind(this);
+        this.stopArrowScoll = this.stopArrowScoll.bind(this);
```
reiniergs: @LeandroTorresSicilia we should have util here that does all the bindings, we have exactly the same code every where in-class components, something like this

```js
bind({
    methods: [
        'handleScrollDownArrowHover',
       'handleScrollUpArrowHover',
    ],
    to: this,
});
```
it will be more semantic and reduce litte the component code, agree??

LeandroTorresSicilia: Yes, I agree with you, I will create an issue for this

src/components/Lookup/index.js
```diff
@@ -236,6 +243,63 @@ class Lookup extends Component {
...
+        this.stopArrowScoll();
+
+        const menu = this.menuRef.current.getRef();
+        this.scrollingTimer = setInterval(() => {
+            if (menu.scrollTop > 0) {
+                menu.scrollBy(0, -1);
+            } else {
+                this.stopArrowScoll();
+            }
+        }, 5);
```
reiniergs: not a fan on `setInterval`, prefer to implement same behavior using only `setTimeout`, setInterval can cause serious memory leaks

src/components/Lookup/index.js
```diff
@@ -236,6 +243,63 @@ class Lookup extends Component {
...
+        const menu = this.menuRef.current.getRef();
+        const showScrollUpArrow = menu.scrollTop > 0;
+        const showScrollDownArrow = !isScrollPositionAtMenuBottom(menu);
+        this.setState({
+            showScrollUpArrow,
+            showScrollDownArrow,
+        });
+    }
+
+    // - end of mine
```
reiniergs: since we are using the same logic in other components or it can potentially be used then we can move it out of the component into a library, this behavior doesn't have to be tied to react, it's pure DOM manipulation of the scrollable container.

src/components/Lookup/index.js
rafaelmg83: i changed `setInterval` by `setTimeout` already.

Issue #1157, opened 2019-11-20
test: add integration-test to `reorder-child` functionality in all components


TahimiLeonBravo:
- [x] Accordion (https://zpl.io/VYR5ELd)
- [ ] CarouselCard (https://zpl.io/bJvwYAD)
- [ ] PrimitiveMenu (https://zpl.io/adkNDyL)
- [ ] Tabset (https://zpl.io/VODL45r)
- [ ] Add test-case for when unregister child in `Tabset`, `Accordion`, `CarouselCard`, `PrimitiveMenu` and `Picklist` components

yvmunayev:
@TahimiLeonBravo I check the Picklist component tests and start

PR #1173, opened 2019-11-29
test: integration test to reorder child functionality in Accordion

yvmunayev:
#1157

-add integration-test to `reorder-child` functionality in Accordion with AccordionOption changed dynamically

@nexxtway/react-rainbow


integration/specs/Accordion/accordion-9.spec.js
```diff
@@ -0,0 +1,25 @@
...
+
+const addAdvancedSettings = () => $('#button-icon_add-new-advanced-settings').click();
+
+describe('Accordion with AccordionOption changed dynamically', () => {
+    it('should select the new option with keyboard after it is added dynamically', () => {
+        browser.url('/#!/Accordion/9');
+        const component = $(ACCORDION);
+        component.waitForExist();
+        const accordion = new PageAccordion(ACCORDION);
+        accordion.clickFirstButton();
```
LeandroTorresSicilia: The accordion page object has a getItem method that returns a page object for each section child. Since we don't know how much sections will have an Accordion and we would need to click each one, instead of creating a clickFirstButton method you should do:
```
const firstSection = accordion.getItem(0);
firstSection.clickButtonIcon();
```
with this approach, you will be able to click whatever section you want

integration/specs/Accordion/accordion-9.spec.js
```diff
@@ -0,0 +1,25 @@
...
+        browser.keys(ARROW_DOWN_KEY);
+        browser.keys(ENTER_KEY);
+        expect(accordion.getOpenSectionLabel()).toBe('Personal Profile');
+        browser.refresh();
+        addAdvancedSettings();
+        accordion.clickFirstButton();
```
LeandroTorresSicilia: same ^^

package.json
```diff
@@ -142,7 +143,7 @@
         "sass-loader": "^7.1.0",
         "shelljs": "^0.8.3",
         "style-loader": "^0.22.1",
-        "stylelint": "^9.5.0",
+        "stylelint": "^12.0.0",
```
LeandroTorresSicilia: For now, don't change anything in package.json since there packages with breaking change and we need to be careful when update them.

src/components/Accordion/pageObject/index.js
```diff
@@ -30,6 +30,29 @@ class PageAccordion {
...
+    /**
+     * Returns the label of the open AccordionSection
+     * @method
+     * @returns {string}
+     */
+    getOpenSectionLabel() {
+        return $(this.rootElement)
+            .$('div[aria-hidden="false"]')
+            .$('..')
+            .$('div h3 > span:nth-child(2)')
```
LeandroTorresSicilia: This is a fragile selector since we can change the markup and then break it. Here you can set:
```data-expanded={isExpanded}```
in the ```StyledHeading ``` component and then select it as:
```
$(this.rootElement)
            .$('[data-expanded="true"]')
            .$('[title="Accordion Label"]')
```
also in this way is more readable

File: src/components/Accordion/readme.md [hunk 5]
```diff
@@ -159,3 +159,117 @@ import { Card, Accordion, AccordionSection } from 'react-rainbow-components';
...
+    Card, 
+    Accordion, 
+    AccordionSection,
+    ButtonGroup,
+    ButtonIcon,
+} from 'react-rainbow-components';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import {
+    faPlus,
+  	faEllipsisV,
```
LeandroTorresSicilia: fix indentation

src/components/Accordion/readme.md
```diff
@@ -159,3 +159,117 @@ import { Card, Accordion, AccordionSection } from 'react-rainbow-components';
...
+                        variant="border"
+                        disabled
+                        icon={<FontAwesomeIcon icon={faPlus} />}
+                    />
+                    <ButtonIcon
+                        variant="border"
+                        disabled
+                        icon={<FontAwesomeIcon icon={faEllipsisV} />}
+                    />
+                </ButtonGroup>
```
LeandroTorresSicilia: fix indentation for closing ButtonGroup  and both ButtonIcon

PR #1178, opened 2019-12-04
test: add integration-test to `reorder-child` functionality in CarouselCard

yvmunayev:
fix: #1157 

- test: add integration-test to `reorder-child` functionality in CarouselCard with CarouselImagen changed dynamically
- fix: `reorder-child` functionality in CarouselCard component
- fix: add `reorder-child` functionality to Indicator component 
- fix: isSelected function of PageCarouselCardIndicator class

@nexxtway/react-rainbow


integration/specs/CarouselCard/carouselCard-3.spec.js
```diff
@@ -0,0 +1,28 @@
...
+const CAROUSEL = '#carousel-3';
+
+const addNewCard = () => $('#button-icon_add-new-card').click();
+
+describe('CarouselCard with CarouselImagen changed dynamically', () => {
+    it('should select the new option with keyboard after it is added dynamically', () => {
+        browser.url('/#!/CarouselCard/3');
+        const component = $(CAROUSEL);
+        component.waitForExist();
+        var carousel = new PageCarouselCard(CAROUSEL);
```
LeandroTorresSicilia: Let's use const instead, we never use var, when the variable is not reassigned use const otherwise use let.

src/components/CarouselCard/indicators/utils.js
```diff
@@ -0,0 +1,23 @@
...
+    newChildrenRefs.sort(
+        (refA, refB) => getRefIndex(nodes, refA.ref) - getRefIndex(nodes, refB.ref),
+    );
+    return newChildrenRefs;
+}
+
+export function insertChildOrderly(childrenRefs, childRef, nodes) {
+    const newChildrenRefs = childrenRefs.concat([childRef]);
+    return sortChildren(newChildrenRefs, nodes);
+}
```
LeandroTorresSicilia: These last three methods are the same that are in the other utils file, then let's remove these and use the ones in the other utils and only keep the first one that is different.

src/components/CarouselImage/pageObject/index.js
```diff
@@ -21,6 +21,17 @@ class PageCarouselImage {
             .$('a')
             .click();
     }
+
+    /**
+     * Returns the header of the CarouselImage.
+     * @method
+     * @returns {string}
+     */
+    getHeader() {
```
LeandroTorresSicilia: since we are getting a text use `getHeaderText`

src/components/CarouselImage/pageObject/index.js
```diff
@@ -21,6 +21,17 @@ class PageCarouselImage {
...
+        return $(this.rootElement)
+            .$('[title="Imagen Header"]')
+            .getHTML(false);
```
LeandroTorresSicilia: here you can use getText

yvmunayev: The problem with getText () is that if the element is not visible the function returns empty.

src/components/CarouselCard/readme.md
```diff
@@ -48,3 +48,98 @@ const carouselContainerStyles = {
...
+                        onClick={() => this.addNewCard()}
+                        variant="border"
+                        icon={<FontAwesomeIcon icon={faPlus} />}
+                    />
+                </GlobalHeader>
+                <CarouselCard
+                    style={carouselContainerStyles}
+                    className="rainbow-align-content_center rainbow-m_auto"
+                    id="carousel-3"
+                    disableAutoScroll={true}
```
LeandroTorresSicilia: we usually only use:
```disableAutoScroll```
when passing true values as props
and when pass false:
```disableAutoScroll={false}```
here there is no problem because this is a markdown file that eslint doesn't inspect but in js files, the linter doesn't allow you to do this.

Issue #1191, opened 2019-12-08
feat: add support Theme for Badge component


TahimiLeonBravo:
https://github.com/nexxtway/react-rainbow/projects/1#card-29896923

yvmunayev:
@TahimiLeonBravo

PR #1194, opened 2019-12-09
feat: add support theme for badge component

yvmunayev:
fix: #1191

- feat: add support theme for badge component

@nexxtway/react-rainbow


src/components/Badge/styled/container.js
```diff
@@ -32,14 +36,14 @@ const StyledContainer = styled.span`
         props.variant === 'brand' &&
         `
             color: ${COLOR_WHITE};
```
LeandroTorresSicilia: Here we need to get the contrast text computed as we did in the Button component. This is necessary because if the customer sets a light brand color we need to present a text color with the right contrast text in order to make it visually possible. One clear example of this if when the user sets #fff color as a brand, keeping text color to #fff as well will be visually impossible since all we get is a white box.

Issue #757, opened 2019-05-24
feat: add `hour24` to the TimePicker component


TahimiLeonBravo:

alonsosfera:
Hey! Can I work on this?

LeandroTorresSicilia:
@alonsosfera  Yeah, you are welcome to contribute

TahimiLeonBravo:
Hi @alonsosfera 
We released a new version 1.12.0 where the `hour24` prop is available.
Thanks for your contribution :)

PR #1230, opened 2019-12-19
feat: add `hour24` to the TimePicker component

alonsosfera:
feat: #757

-Add `hour24` prop to the TimePicker component to set its format to 24 hrs.

@nexxtway/react-rainbow


src/components/TimePicker/timeSelect.js
```diff
@@ -342,7 +351,7 @@ export default class TimeSelect extends Component {
             ampm,
         });
 
-        if (currentHour && currentMinutes && ampm) {
+        if (currentHour && currentMinutes) {
```
LeandroTorresSicilia: Here we need a check because in the case of 12 hours if the user doesn't select the am pm value then we should not send the onChange event.

LeandroTorresSicilia:
There is a warning in console when pressing the right arrow key twice.
Screen Shot 2019-12-19 at 10 11 47 AM
Seems that we need another check on keyboard navigation.
The same happens when the minute's input is focused and type a number greater than 6, because it will try to focus the third input (am pm select) and since it doesn't exist then fails.

alonsosfera:
@LeandroTorresSicilia Thanks for your feedback, it's greatly appreciated! I just pushed the changes to take care of those little things. If there is anything else, please feel free to let me know

PR #1243, opened 2019-12-23
feat: poc proposal for background API in theme

yvmunayev:
fix: #1239

- add "background" attr to palette
- add "text" attr to palette
- add "divider" attr to palette
- add "action" attr to palette
- add normalization to new palette's attrs using "background" attr
- add support background theme for Input 
- add support background theme for MenuButtton  
- add support background theme for Modal
- add support background theme for Card
- add support background theme for Calendar
- add getBrightness function to color helpers
- add isDark function to color helpers
- add support shadows themes
- add attachThemeAttrs function

@nexxtway/react-rainbow


library/styleguideComponents/Wrapper/index.js
```diff
@@ -62,10 +62,7 @@ const cyanDarkTheme = {
             brand: '#80deea',
             success: '#b9f6ca',
             error: '#ff5252',
-            background: {
-                primary: '#242424',
-                secondary: '#333333',
-            },
+            background: '#303030',
```
LeandroTorresSicilia: let's use mainBackground

src/styles/helpers/color/getBrightness.js
```diff
@@ -0,0 +1,17 @@
+import isValidColor from './isValidColor';
+import decomposeColor from './decomposeColor';
+import hslToRgb from './hslToRgb';
+
+/* http://www.w3.org/TR/AERT#color-contrast */
+
+/* eslint-disable no-param-reassign */
+export default function getBrightness(color) {
```
LeandroTorresSicilia: we must add tests to this helper, but can be done in another PR

library/styleguideComponents/Wrapper/index.js
yvmunayev: I used background to not break the continuity of de brand, success, and error, for the API it is background and then when it normalizes then it is:

background: {
        main: COLOR_WHITE,
        secondary: darken(COLOR_WHITE, 0.2),
},

maxxgreene: from the consumer perspective, `mainBackground` makes more sense, because I could understand it's the predominant background in my app or design, using background it will prompt the question on my head ... background??    background of what?? 

the generated background object is an implementation detail, and to be honest I don't think there will be many people trying to do more complex customization ... if we decide to move forward with the more complex customization of the backgrounds we could always allow to pass background object that it will take precedence over `mainBackground`

src/styles/helpers/color/getBrightness.js
```diff
@@ -0,0 +1,17 @@
+import isValidColor from './isValidColor';
+import decomposeColor from './decomposeColor';
+import hslToRgb from './hslToRgb';
+
+/* http://www.w3.org/TR/AERT#color-contrast */
+
+/* eslint-disable no-param-reassign */
```
maxxgreene: do not disable `eslint` here, just assign the value to a different var

src/styles/helpers/color/getBrightness.js
```diff
@@ -0,0 +1,17 @@
+import isValidColor from './isValidColor';
+import decomposeColor from './decomposeColor';
+import hslToRgb from './hslToRgb';
+
+/* http://www.w3.org/TR/AERT#color-contrast */
```
maxxgreene: remove this comment, does not add any value

src/styles/helpers/color/getBrightness.js
```diff
@@ -0,0 +1,17 @@
...
+
+/* eslint-disable no-param-reassign */
+export default function getBrightness(color) {
+    if (isValidColor(color)) {
```
maxxgreene: the validation of the color should be made only one time at the top level normalization, at this moment the color will be always valid, it will also avoid  returning undefined .. which is weird :)

src/styles/helpers/normalizeTheme.js
```diff
@@ -32,6 +32,33 @@ function normalizeColors(colors) {
...
+            action: {
+                active: dark ? lighten(background, 0.54) : darken(background, 0.54),
+                hover: dark ? lighten(background, 0.1) : darken(background, 0.1),
+                selected: dark ? lighten(background, 0.2) : darken(background, 0.2),
+                disabled: dark ? lighten(background, 0.3) : darken(background, 0.3),
+            },
+        };
+        return theme;
+    }
+    return {};
```
maxxgreene: should we return default values here?? we must not have visual breaking changes

src/styles/helpers/normalizeTheme.js
```diff
@@ -32,6 +32,33 @@ function normalizeColors(colors) {
     }, {});
 }
 
+function normalizeBackground(background) {
+    if (background) {
```
maxxgreene:
```
if (background and isValidColor(background)) {
}
```

src/styles/defaultTheme.js
```diff
@@ -22,8 +29,20 @@ export default {
...
+        text: {
+            main: COLOR_DARK_1,
+            secondary: darken(COLOR_DARK_1, 0.3),
+            disabled: darken(COLOR_DARK_1, 0.5),
+        },
+        action: {
+            active: darken(COLOR_WHITE, 0.54),
+            hover: darken(COLOR_WHITE, 0.1),
+            selected: darken(COLOR_WHITE, 0.2),
+            disabled: darken(COLOR_WHITE, 0.3),
```
maxxgreene: default values shouldn't be computed because it will be a visual breaking change

src/styles/helpers/normalizeTheme.js
yvmunayev: Since the return of normalizeTheme has the default values, if normalizeBackground returns {} that values are not changed.

src/styles/helpers/color/getBrightness.js
yvmunayev: ok

src/components/MenuItem/styled/li.js
```diff
@@ -1,8 +1,17 @@
...
+const StyledLi = styled.li.attrs(props => {
+    const palette = getTheme(props).palette;
+    const { main: mainText, disabled: disabledText } = palette.text;
+    const hover = palette.action.hover;
+    return {
+        mainText,
+        hover,
+        disabledText,
+    };
+})`
```
maxxgreene: since we are going to be repeating this in every styled component, can we create kinda of hoc that dynamically does the work??

maxxgreene:
```
const SyliedLi = attachThemeAttr(styled.li)`
    color: ${props => props.palette.text.main};
`;
```

```
function attachThemeAttrs(styledElement) {
    return atyledElement.attrs((props) => {
        const theme = getTheme(props);
        return {
             ...props,
             ...theme
        }
    })
}
```

src/styles/helpers/normalizeTheme.js
maxxgreene: sure, I understand ... but there is a semantic problem when you read the code, you are normalizing which means it needs to return  the normalized values or a fallback ... and you are returning `{}` ... we would change the name to `resolveCustomBackground` ... which reads better according to your implementation

library/exampleComponents/GlobalHeader/styled.js
```diff
@@ -4,16 +4,14 @@ import getTheme from '../../../src/styles/helpers/getTheme';
...
+    const secondaryBackground = getTheme(props).palette.background.secondary;
     const divider = darken(secondaryBackground, 0.1);
     return {
-        primaryBackground,
+        secondaryBackground,
         divider,
     };
 })`
-    background-color: ${props => props.primaryBackground};
+    background-color: ${props => props.secondaryBackground};
```
TahimiLeonBravo: In the defaultTheme,  the background-color on the GlobalHeader we should maintain the color white defined on the design

TahimiLeonBravo:
We should keep the library user interface as designed.
### Expected:
Untitled

### Current:
Untitled 2

src/components/Card/styled/container.js
```diff
@@ -1,20 +1,29 @@
...
+        shadow4,
+        shadow6,
+    };
+})`
     position: relative;
-    background-color: ${COLOR_WHITE};
-    border: 0.0625rem solid ${COLOR_GRAY_2};
+    background-color: ${props => props.palette.background.main};
+    color: ${props => props.palette.text.main};
+    border: 0.0625rem solid ${props => props.palette.divider};
```
TahimiLeonBravo: - I think that is not semantic, why we are passing to the border `palette.divider`?
In order to, we have different borders color in various components
e.g the `Input`componet use a different `border-color` than the `Card` component
- Shouldn't we create an entry with the name border?

src/components/Card/styled/container.js
```diff
@@ -1,20 +1,29 @@
...
 import { BORDER_RADIUS_1 } from '../../../styles/borderRadius';
-import { SHADOW_4, SHADOW_6 } from '../../../styles/shadows';
 import { MARGIN_LARGE } from '../../../styles/margins';
+import getTheme from '../../../styles/helpers/getTheme';
 
-const StyledContainer = styled.article`
+const StyledContainer = styled.article.attrs(props => {
+    const palette = getTheme(props).palette;
+    const shadow4 = `0 1px 2px 0 ${palette.divider}`;
+    const shadow6 = `0 2px 12px 0 ${palette.divider}`;
```
TahimiLeonBravo: I think that is not semantic, why we are passing to the shadow `palette.divider`?
Shouldn't we create an entry with the name shadow?

TahimiLeonBravo:
@yvmunayev 
We have using the incorrect color for the title of the card
### Current
card

### Expected
card expected

TahimiLeonBravo:
@yvmunayev 
Always the `shadow-color` should be darker

#### Current
card-dark

#### Expected
card-dark expected

src/components/Card/styled/container.js
```diff
@@ -1,20 +1,20 @@
...
     position: relative;
-    background-color: ${COLOR_WHITE};
-    border: 0.0625rem solid ${COLOR_GRAY_2};
+    background-color: ${props => props.palette.background.main};
+    color: ${props => props.palette.text.main};
+    border: 0.0625rem solid ${props => props.palette.border.divider};
     border-radius: ${BORDER_RADIUS_1};
     background-clip: padding-box;
-    box-shadow: ${SHADOW_4};
+    box-shadow: ${props => props.palette.shadows.shadow_4};
```
LeandroTorresSicilia: I think we should put shadows at the same level of palette, not under it.

src/components/Input/checkboxRadioLabel/labelText.js
```diff
@@ -1,14 +1,22 @@
...
+    };
+})`
     display: inline;
     vertical-align: middle;
-    color: ${COLOR_DARK_1};
+    color: ${props => props.mainText};
     font-size: ${FONT_SIZE_TEXT_LARGE};
     line-height: 1.65rem;
-    ${props => props.disabled && `color: ${COLOR_GRAY_3};`};
+    ${props => props.disabled && `color: ${props.disabledText};`};
```
LeandroTorresSicilia: @TahimiLeonBravo  check this

src/components/Input/styled/iconContainer.js
```diff
@@ -1,12 +1,16 @@
...
+    const palette = getTheme(props).palette;
+    const errorMainColor = palette.error.main;
+    const headerText = palette.text.header;
+    return {
+        errorMainColor,
+        headerText,
+    };
 })`
-    fill: ${COLOR_GRAY_3};
+    color: ${props => props.headerText};
```
LeandroTorresSicilia: @TahimiLeonBravo check this maybe we can define color for icons or other since header text color is not semantic to use in icon fill color

src/components/MenuItem/styled/li.js
```diff
@@ -17,24 +26,24 @@ const StyledLi = styled.li`
...
 
     &:focus {
         outline: 0;
-        background-color: ${COLOR_GRAY_1};
+        background-color: ${props => props.hover};
     }
 
     &:active {
-        background-color: ${COLOR_GRAY_1};
+        background-color: ${props => props.hover};
```
LeandroTorresSicilia: Maybe we can use other action for this like active and use it here and above when focus since hover is not semantic to focus and active

src/styles/defaultTheme.js
```diff
@@ -22,8 +33,35 @@ export default {
...
+            brand: `0 0 2px ${COLOR_BRAND}`,
+            success: `0 0 2px ${COLOR_SUCCESS}`,
+            error: `0 0 2px ${COLOR_ERROR}`,
+            shadow_1: `0 0 2px 0 ${COLOR_GRAY_3}`,
+            shadow_2: `0 2px 4px 0 ${COLOR_GRAY_2}`,
+            shadow_3: `0 0 1px 0 ${COLOR_GRAY_3}`,
+            shadow_4: `0 1px 2px 0 ${COLOR_GRAY_2}`,
+            shadow_5: `0 0 3px ${COLOR_GRAY_1}`,
+            shadow_6: `0 2px 12px 0 ${COLOR_GRAY_2}`,
         },
```
LeandroTorresSicilia: Good job, but I think we should put this outside palette since shadows are not simply colors

PR #1251, opened 2020-01-02
chore: move react, react-dom, styled-components to peer dependencies

therealparmesh:
fix: #

- Change base dependencies like React to be peer dependencies. This makes it so that consumers never mistakenly bundle multiple versions of React, or resort to use alias-based hacks.

@nexxtway/react-rainbow


LeandroTorresSicilia:
@therealparmesh 
Thanks for your contribution. You are welcome to the rainbow community.

package.json
```diff
@@ -35,12 +35,14 @@
...
         "prop-types": "^15.6.2",
-        "react": "^16.8.6",
         "react-async-script-loader": "^0.3.0",
-        "react-dom": "^16.8.6",
-        "styled-components": "^4.3.2",
         "xlsx": "^0.14.4"
     },
+    "peerDependencies": {
+        "react": "^16.8.6",
+        "react-dom": "^16.8.6",
```
LeandroTorresSicilia: We should not restrict a specific version of react, instead, we should let the user use whatever version greater than 16.8.0 since this is the version where react introduce the hooks.
```
"react": ">=16.8.0",
"react-dom": ">=16.8.0"
```

package.json
```diff
@@ -35,12 +35,14 @@
...
+        "styled-components": "^4.3.2"
```
LeandroTorresSicilia: I think we should leave styled-components as before since we don't want to obligate the user to install styled-components if they don't use it.

Issue #1253, opened 2020-01-05
test: add test to AvatarGroup component


TahimiLeonBravo:
#### helpers: abbreviateNumber
- [ ] should return a formatted number using SI

#### AvatarGroup
- [ ] should pass the props to children as expected

#### Avatars
- [ ] should render default amount (3) of Avatar components when none maxAvatars is passed
- [ ] should render right amount of Avatar components when maxAvatars is passed
- [ ] should Avatar components have the right props passed
- [ ] should render an img when src or icon prop is passed
- [ ] should render an abbr when initials prop is passed

#### Counter
- [ ] should have the right size passed
- [ ] should render the total amount number of avatars passed

LeandroTorresSicilia:
#### helpers: abbreviateNumber
- [ ] should return the same number passed when it is less than 1000 (add here an array with some cases like [8, 75, 999])
- [ ] should attach `k` suffix when number is greater than 1000 and less than ... (same ^^ add some cases with differents digits each one)
- [ ] should attach `M` suffix ...
- [ ] should attach `G` suffix ...
also here add a test case for wrong values 

#### AvatarGroup
- [ ] should pass the prop showCounter as isTrue to RenderIf component

#### Avatars
- [ ] should render default amount (3) of Avatar components when none maxAvatars is passed
- [ ] should render right amount of Avatar components when maxAvatars is passed
- [ ] should pass the right zIndex to each Avatar

#### Counter
- [ ] should have the right size passed
- [ ] should render the total amount number of avatars passed

#### Counter
- [ ] should call abbreviateNumber with the total of avatars
- [ ] should render the total amount number of avatars passed (note that here you need to check  that the text rendered is what return the mock of abbreviateNumber)

PR #1255, opened 2020-01-07
test: add unit tests to avatargroup component

rafaelmg83:
#1253 

- Add unit tests to AvatarGroup component
- Some fixes to color functions

@nexxtway/react-rainbow

### DIFF
--- a/src/components/AvatarGroup/__test__/avatarGroup.spec.js
+++ b/src/components/AvatarGroup/__test__/avatarGroup.spec.js
@@ -1,0 +1,39 @@
+import React from 'react';
+import { mount } from 'enzyme';
+import AvatarGroup from '../';
+import RenderIf from '../../RenderIf';
+
+jest.mock('../../RenderIf', () => () =>
+    // eslint-disable-next-line no-shadow
+    RenderIf => props => <RenderIf {...props} />,
+);
+
+describe('<AvatarGroup />', () => {
+    const avatars = [
+        {
+            src: 'images/user/user1.jpg',
+            assistiveText: 'Jose Leandro',
+            title: 'Jose Leandro',
+        },
+        {
+            src: 'images/user/user2.jpg',
+            assistiveText: 'Tahimi Leon',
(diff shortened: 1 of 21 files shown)


src/components/AvatarGroup/__test__/avatarGroup.spec.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
+import { mount } from 'enzyme';
+import AvatarGroup from '../';
+import RenderIf from '../../RenderIf';
+
+jest.mock('../../RenderIf', () => () =>
+    // eslint-disable-next-line no-shadow
+    RenderIf => props => <RenderIf {...props} />,
+);
```
LeandroTorresSicilia: Mocks are usually used with helper functions, in the case of other components is not necessary to mock them.

src/components/AvatarGroup/__test__/counter.spec.js
```diff
@@ -0,0 +1,35 @@
...
+            title: 'Carlos Miguel',
+        },
+        {
+            src: 'images/user/user4.jpg',
+            assistiveText: 'Jane Doe',
+            title: 'Jane Doe',
+        },
+    ];
+
+    it('should call abbreviateNumber with the total of avatars and should render that number in the text', () => {
```
LeandroTorresSicilia: since we only test the final result as the text rendered, let's change this name to something like:
`should render the right total number of avatars`

src/components/AvatarGroup/helpers/abbreviateNumber.js
```diff
@@ -1,4 +1,7 @@
 export default function abbreviateNumber(number) {
+    if (!number || number < 0) {
+        throw Error('Invalid number');
```
LeandroTorresSicilia: We usually think that it is better to not fail at all and instead show a console error and return some value like an empty string, undefined, or null. Since this function always return a string we should return an empty string here.

Issue #1254, opened 2020-01-05
feat: add prop `color` to Avatar component


TahimiLeonBravo:
### Requirements
- [ ] The color should be validated, if the color passed is invalid make fallback to the currentColor (brand or inverse)

### Test
- [ ] The new functionality should be tested

PR #1256, opened 2020-01-08
feat: add prop color to avatar component

rafaelmg83:
#1254 

- Add prop color to avatar component

@nexxtway/react-rainbow

### DIFF
--- a/src/components/Avatar/index.js
+++ b/src/components/Avatar/index.js
@@ -3,7 +3,6 @@ import PropTypes from 'prop-types';
 import AvatarContent from './avatarContent';
 import AssistiveText from './../AssistiveText';
 import StyledContainer from './styled/container';
-import { validateColor } from '../../styles/helpers/color';
 
 /**
  * An avatar component represents an object or entity
@@ -11,10 +10,8 @@ import { validateColor } from '../../styles/helpers/color';
 export default function Avatar(props) {
     const { className, style, size, assistiveText, color, ...rest } = props;
 
-    const validatedColor = validateColor(color);
-
     return (
-        <StyledContainer className={className} style={style} size={size} color={validatedColor}>
+        <StyledContainer className={className} style={style} size={size} color={color}>
             <AvatarContent {...rest} assistiveText={assistiveText} />
             <AssistiveText text={assistiveText} />
         </StyledContainer>
--- a/src/components/Avatar/index.js
+++ b/src/components/Avatar/index.js
@@ -8,10 +8,15 @@ import StyledContainer from './styled/container';
  * An avatar component represents an object or entity
  */
 export default function Avatar(props) {
-    const { className, style, size, assistiveText, color, ...rest } = props;
+    const { className, style, size, assistiveText, backgroundColor, ...rest } = props;
 
     return (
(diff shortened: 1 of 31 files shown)


src/styles/helpers/color/validateColor.js
```diff
@@ -0,0 +1,10 @@
+export default function validateColor(color) {
+    const EMPTY_STRING = '';
+    if (!color) return EMPTY_STRING;
+    const s = new Option().style;
+    s.color = color;
+    const removeSpaces = colorStr => colorStr.replace(/\s+/g, EMPTY_STRING);
+    const isValidColor = removeSpaces(s.color) === removeSpaces(color);
+    const isValidHex = /^#([0-9A-F]{3}){1,2}$/i.test(color);
+    return isValidColor || isValidHex ? color : EMPTY_STRING;
+}
```
reiniergs: uumm, too complicated 

```js
export default function validateColor(color) {
    const element = document.createElement('a');
    element.style.color = color;
    return element.style.color !=== '';
}
```

I'm can of confuse with the name of the function, it says validateColor which means return true|false .. but you are return empty string sometimes, it's a normalization then ??

reiniergs: I see you are using it above like a normalizer ... uumm ... there are two levels of abstraction in your implementation, you are first normalizing, then in the styled component if empty you fallback to brand color ... think about it .. too many levels of abstraction make harder to understand to somebody else because there are color logic in two places

let's simplify by implementing the logic only in the styled component with a function `isColorValid`, if not then the fallback to brand color ... the avatar component just pass-through the color down ... it makes sense??

src/styles/helpers/color/validateColor.js
```diff
@@ -0,0 +1,5 @@
+export default function validateColor(color) {
```
maxxgreene: it looks good to me. I would just name the function `isValidColor` then in styled component I would do 
```
const brandMainColor = isValidColor(props.color) ? props.color : brand.main;
```

what do you think? :)

src/components/Avatar/index.js
```diff
@@ -42,6 +42,8 @@ Avatar.propTypes = {
     className: PropTypes.string,
     /** An object with custom style applied to the outer element. */
     style: PropTypes.object,
+    /** A valid color. */
```
maxxgreene: @TahimiLeonBravo this description will need more explanation here it's purpose. it will be fine if we merge the PR with the current text or please provide desired text for @rafaelmg83 here

maxxgreene: also @TahimiLeonBravo we should consider name this prop backgroundColor since it could be easily related to CSS color property when it will affect/change the background color of the avatar when initials are being used

TahimiLeonBravo: @rafaelmg83 let's change the name of the prop to `backgrondColor`

src/styles/helpers/color/validateColor.js
rafaelmg83: done!

src/components/Avatar/index.js
rafaelmg83: done!

PR #1261, opened 2020-01-11
fix: the textColor and cursor to the Avatar when hover

cmcartaya:
fix: #1259 

- fix the text color and cursor on Avatar component when hover.

@nexxtway/react-rainbow

### DIFF
--- a/src/components/Avatar/styled/container.js
+++ b/src/components/Avatar/styled/container.js
@@ -6,14 +6,15 @@ const StyledContainer = styled.span.attrs(props => {
     const theme = getTheme(props);
     const { getContrastText, brand } = theme.palette;
     const brandMainColor = isValidColor(props.backgroundColor) ? props.backgroundColor : brand.main;
+    const brandContrastTextColor = getContrastText(brandMainColor);
 
     return {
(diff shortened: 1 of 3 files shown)


src/components/Avatar/styled/container.js
```diff
@@ -25,7 +25,7 @@ const StyledContainer = styled.span.attrs(props => {
 
     &:hover,
     &:focus {
-        color: currentColor;
+        color: ${props => props.getContrastText(props.brandMainColor)};
```
LeandroTorresSicilia: since above we are using already:
```props.getContrastText(props.brandMainColor)}```
we should put this inside attrs function in order to compute getContrastText for brand color only once and inject a brandContrastTextColor like this:
```
const StyledContainer = styled.span.attrs(props => {
    const theme = getTheme(props);
    const { getContrastText, brand } = theme.palette;
    const brandMainColor = isValidColor(props.backgroundColor) ? props.backgroundColor : brand.main;
    const brandContrastTextColor = getContrastText(brandMainColor);
    return {
        brandMainColor,
        brandContrastTextColor,
    };
})
```

Issue #1220, opened 2019-12-18
fix: the theme selected when the user `expand` or `shrink` the InteractiveExample


TahimiLeonBravo:
#### Description
The theme selected is missing when the user expands an interactive example

PR #1266, opened 2020-01-16
fix: the theme selected when the user expand or shrink the interactive example

wildergd:
fix: #1220 

- fix bug with the theme selected when the user `expand` or `shrink` the InteractiveExample

@nexxtway/react-rainbow

### DIFF
--- a/library/styleguideComponents/Wrapper/index.js
+++ b/library/styleguideComponents/Wrapper/index.js
@@ -1,5 +1,5 @@
 /* eslint-disable import/no-extraneous-dependencies */
-import React, { useState, useRef, useEffect } from 'react';
+import React from 'react';
 import PropTypes from 'prop-types';
 import { createStore, combineReducers } from 'redux';
 import { reducer as formReducer } from 'redux-form';
@@ -21,6 +21,7 @@ import {
     StyledColorCircle,
     StyledCheckIcon,
 } from './styled';
+import useCachedState from '../../hooks/useCachedState';
 
 const rootReducer = combineReducers({
     form: formReducer,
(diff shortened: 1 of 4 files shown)


library/styleguideComponents/Wrapper/index.js
```diff
@@ -70,10 +70,63 @@ const cyanDarkTheme = {
...
+    pageName: window.location.hash.split('/')[1],
+    examplesCount: 0,
+    examplesThemes: [],
+};
+
+function registerExample() {
+    const index = pageStatus.examplesCount * 2 + 1;
+    if (!pageStatus.examplesThemes[index]) {
+        pageStatus.examplesThemes[index] = undefined;
+    }
```
LeandroTorresSicilia: I think this is not needed since if pageStatus.examplesThemes[index] does not exists is because it is undefined already

library/styleguideComponents/Wrapper/index.js
```diff
@@ -70,10 +70,63 @@ const cyanDarkTheme = {
...
+}
+
+function reset() {
+    pageStatus.examplesThemes = [];
+    pageStatus.examplesCount = 0;
+}
+
+function getThemeFromCache(id) {
+    const index = (Number(id) - 1) / 2;
+    return pageStatus.examplesThemes[index] || undefined;
```
LeandroTorresSicilia: Not need of make || undefined since pageStatus.examplesThemes[index] will return undefined if it does not exists

library/styleguideComponents/Wrapper/index.js
```diff
@@ -70,10 +70,63 @@ const cyanDarkTheme = {
...
+}
+
+function updateThemeInCache(id, theme) {
+    const index = (Number(id) - 1) / 2;
```
LeandroTorresSicilia: since this is used twice make a function with this

PR #1290, opened 2020-01-23
feat: add theme support to ProgressStep, RadioButtonGroup, RadioGroup and Rating components

wildergd:
fix: #1279 

- add theme support to ProgressStep component
- add theme support to RadioButtonGroup component
- add theme support to RadioGroup component
- add theme support to Rating component

@nexxtway/react-rainbow


wildergd:
Here we need to check the style of RadioButtonGroup component in the first example... set Dark Cyan theme

src/components/RadioButtonGroup/styled/errorText.js
```diff
@@ -0,0 +1,9 @@
+import styled from 'styled-components';
+import attachThemeAttrs from '../../../styles/helpers/attachThemeAttr';
+import ErrorText from '../../Input/styled/errorText';
+
+const StyledTextError = attachThemeAttrs(styled(ErrorText))`
+    text-align: center;
+`;
```
LeandroTorresSicilia: Since there is not use of theme values here remove the attachThemeAttrs. Maybe you can get rid of this styled since ErrorText from Input has an alignSelf prop

src/components/RadioButtonGroup/styled/marker.js
```diff
@@ -32,15 +22,15 @@ const StyledMarker = styled.span.attrs(props => {
     ${props =>
         props.variant === 'inverse' &&
         `
-            box-shadow: 0 0 4px 0 ${props.brandMainColor};
-            background-color: ${props.brandMainColor};
-            border: solid 1px ${props.brandMainColor};
+            box-shadow: 0 0 4px 0 ${props.palette.brand.main};
```
LeandroTorresSicilia: @TahimiLeonBravo  check this shadow

wildergd: this shadow was originally box-shadow: 0 0 4px 0 ${props.brandMainColor}; 
it is not on the shadows list..... and in order to avoid adding more... i wasn't sure what to do... so i only changed the color

wildergd: i also changed the shadow for variant="brand".. it had a rgba(...) color so i change it in order to use theme colors... perhaps instead of having entire shadows definitions ... it would be better to have a shadows color section in palette ... like borders and backgrounds sections...

Issue #1263, opened 2020-01-13
feat: implement PresenceMap component


TahimiLeonBravo:
- [x] Define API in Google Drive
- [ ] Implement component
- [ ] Define tests spec
- [ ] Make tests

#### Requirements
- [ ] A map should be as single as the map itself with all locations, but it should be more complex than that, a map can have some options as "show traffic" and change the map type.
- [ ] The map should be shown and update the locations of the users connected on real-time
- [ ] it should have the option enable and disable the traffic information
- [ ] it should have the possibility of change the map "type"
- [ ] it should have the possibility of control the nocturne mode, but by default should change automatically depending on the locale on the browser.
- [ ] it should have the possibility of put a custom sign to show on the location element, by default we should put own.
- [ ] it should have an action to move quickly to the current location

#### Design
https://zpl.io/aXJKkOP

#### Unit Test
/ add the specs here /

PR #1292, opened 2020-01-24
feat: new component presencemap

rafaelmg83:
fix: #1263 

- New component PresenceMap

@nexxtway/react-rainbow


src/components/PresenceMap/component.js
```diff
@@ -0,0 +1,141 @@
+import React, { useRef, useState, useEffect } from 'react';
```
reiniergs: the name of the file should be index.js, that way the could import from `'src/component/PresenceMap';`

src/components/PresenceMap/component.js
```diff
@@ -0,0 +1,141 @@
...
+import PropTypes from 'prop-types';
+import getMapContainerStyles from './getMapContainerStyles';
+import StyledContainer from './styled/container';
+import StyledMapContainer from './styled/mapContainer';
+
+export default function MapComponent(props) {
+    const container = useRef();
+    const mapContainer = useRef();
+    const [markers, setMarker] = useState([]);
+    const MAX_ZOOM = 14;
```
reiniergs: this constant could be outside of the function

src/components/PresenceMap/component.js
```diff
@@ -0,0 +1,141 @@
...
+    const prevIsScriptLoaded = useRef(props.isScriptLoaded);
+    useEffect(() => {
+        if (!prevIsScriptLoaded.current && props.isScriptLoaded && props.isScriptLoadSucceed) {
+            // eslint-disable-next-line
+            initMap();
+        }
+        // eslint-disable-next-line
+    }, [props.isScriptLoaded, props.isScriptLoadSucceed]);
+
+    const getContainerClassNames = () => {
```
reiniergs: computed classes should be implemented using `styled-components` approach

src/components/PresenceMap/component.js
rafaelmg83: There is an **index.js** already that imports this **component.js**. PresenceMap is imported in `src/components/index.js` and you can test it in http://localhost:6060/#/PresenceMap. Is that what you mean?

src/components/PresenceMap/component.js
```diff
@@ -0,0 +1,141 @@
...
+                map,
+                zIndex: Math.round(marker.position.lat * 100000),
+            });
+            markersCopy.push(markerMap);
+        });
+
+        if (center === 'auto') {
+            const bounds = new window.google.maps.LatLngBounds();
+
+            markersCopy.map(marker => bounds.extend(marker.getPosition()));
```
reiniergs: if you are not return anything why you are using a map, it seems to be a forEach

src/components/PresenceMap/component.js
```diff
@@ -0,0 +1,141 @@
...
+
+            map.setCenter(bounds.getCenter());
+            map.fitBounds(bounds);
+
+            let removeListener = null;
+            const listener = window.google.maps.event.addListener(map, 'bounds_changed', () => {
```
reiniergs: this listener is inside of a function `initMap()` then every time that function gets invoke you are going to attach the listener

I think this logic should happen within a useEffect for props.center with a condition of === 'auto', recall all the properties of the component can change an as a consequence the component needs to react

src/components/PresenceMap/component.js
```diff
@@ -0,0 +1,141 @@
...
+
+        if (showTraffic) {
+            const trafficLayer = new window.google.maps.TrafficLayer();
+            trafficLayer.setMap(map);
+        }
+
+        if (showTransit) {
+            const transitLayer = new window.google.maps.TransitLayer();
+            transitLayer.setMap(map);
+        }
```
reiniergs: same here, what happen if `showTrafic` change ?? initMap will be call based on this ```!prevIsScriptLoaded.current && props.isScriptLoaded && props.isScriptLoadSucceed``` one time right??

src/components/PresenceMap/component.js
```diff
@@ -0,0 +1,141 @@
+import React, { useRef, useState, useEffect } from 'react';
+import PropTypes from 'prop-types';
+import getMapContainerStyles from './getMapContainerStyles';
+import StyledContainer from './styled/container';
+import StyledMapContainer from './styled/mapContainer';
+
+export default function MapComponent(props) {
+    const container = useRef();
+    const mapContainer = useRef();
+    const [markers, setMarker] = useState([]);
```
reiniergs: `setMarkers` ?? although I dont' think we need state for the markers since they can be compute when needed based on the objects

src/components/PresenceMap/component.js
```diff
@@ -0,0 +1,194 @@
...
+
+export default function MapComponent(props) {
+    const { showTraffic, showTransit, center, zoom, objects, style } = props;
+    const container = useRef();
+    const mapContainer = useRef();
+    const [map, setMap] = useState(null);
+
+    const prevIsScriptLoaded = useRef(props.isScriptLoaded);
+    useEffect(() => {
+        if (!prevIsScriptLoaded.current && props.isScriptLoaded && props.isScriptLoadSucceed) {
```
LeandroTorresSicilia: this is a common pattern (checking for prev prop values), in order to reuse we can create a hook:
```
function usePrevious(value) {
  const ref = useRef();
  useEffect(() => {
    ref.current = value;
  });
  return ref.current;
}
```
and put it in:
```src/libs/hooks/index.js```
then it will be used here as:
```const prevIsScriptLoaded = usePrevious(props.isScriptLoaded);```

File: src/components/PresenceMap/component.js [hunk 9]
```diff
@@ -0,0 +1,194 @@
...
+            // eslint-disable-next-line
+            initMap();
```
LeandroTorresSicilia: this eslint exception is not needed if you define the initMap function before call it

src/components/PresenceMap/component.js
```diff
@@ -0,0 +1,194 @@
...
+        }
+        // eslint-disable-next-line
+    }, [props.isScriptLoaded, props.isScriptLoadSucceed]);
```
LeandroTorresSicilia: this eslint exception is not needed if you add initMap as dependency, but like above it need to be declared before using it

src/components/PresenceMap/styled/mapContainer.js
```diff
@@ -0,0 +1,15 @@
+import styled from 'styled-components';
+import { COLOR_GRAY_1 } from '../../../styles/colors';
+import { BORDER_RADIUS_1 } from '../../../styles/borderRadius';
+import { SHADOW_4 } from '../../../styles/shadows';
+
+const StyledMapContainer = styled.div`
+    align-self: center;
+    border: solid 1px ${COLOR_GRAY_1};
+    border-radius: ${BORDER_RADIUS_1};
+    box-shadow: ${SHADOW_4};
```
LeandroTorresSicilia: let's use theme values here

src/components/PresenceMap/readme.md
```diff
@@ -0,0 +1,53 @@
...
+<div>
+    <GlobalHeader src="images/user/user3.jpg" />
+    <div className="rainbow-p-vertical_x-large rainbow-p-horizontal_small" style={styles}>
+        <PresenceMap
+            apiKey={LIBRARY_GOOGLE_MAPS_APIKEY}
+            objects={objects}
+            // zoom={8}
+            // center={{lat: 26.1267513, lng: -80.298911}}
+            // showTraffic={true}
+            // showTransit={true}
```
LeandroTorresSicilia: not comment code

src/components/PresenceMap/index.d.ts
```diff
@@ -0,0 +1,26 @@
...
+    position: IPosition;
+    heading?: number;
+    image?: string | object;
+    onClick?: (event: MouseEvent<HTMLElement>) => void;
+}
+
+export interface PresenceMapProps extends BaseProps {
+    apiKey: string;
+    objects?: MapObject[];
+    style?: object;
```
LeandroTorresSicilia: since PresenceMapProps extends form BaseProps style is not needed to redefine here

src/components/PresenceMap/styled/mapContainer.js
```diff
@@ -0,0 +1,15 @@
+import styled from 'styled-components';
+import { COLOR_GRAY_1 } from '../../../styles/colors';
+import { BORDER_RADIUS_1 } from '../../../styles/borderRadius';
+import { SHADOW_4 } from '../../../styles/shadows';
+
+const StyledMapContainer = styled.div`
+    align-self: center;
+    border: solid 1px ${COLOR_GRAY_1};
+    border-radius: ${BORDER_RADIUS_1};
```
TahimiLeonBravo: The map should not have visual-styles, this should fit the container where it is inserted

src/components/PresenceMap/index.d.ts
```diff
@@ -0,0 +1,26 @@
...
+    zoom?: number;
+    center?: IPosition;
+    showTraffic?: boolean;
+    children?: ReactNode[] | ReactNode;
```
LeandroTorresSicilia: you can simply use:
```children?: ReactNode;```

File: src/components/PresenceMap/readme.md [hunk 16]
```diff
@@ -0,0 +1,53 @@
...
+        >
+        </PresenceMap>
+    </div>
+</div>
```
TahimiLeonBravo: please add an example like the design :)

src/components/PresenceMap/component.js
```diff
@@ -0,0 +1,194 @@
...
+    };
+
+    const initMap = () => {
+        const mapOptions = {
+            zoom: zoom === 'auto' ? MAX_ZOOM : zoom,
+            center: center !== 'auto' ? center : null,
+            disableDefaultUI: true,
+        };
+        // eslint-disable-next-line
+        const map = new window.google.maps.Map(mapContainer.current, mapOptions);
```
LeandroTorresSicilia: redefine map here with other name

src/components/PresenceMap/component.js
```diff
@@ -0,0 +1,194 @@
...
+                })
+                .catch(error => {
+                    // eslint-disable-next-line no-console
+                    console.log(`Error. Details: ${error}`);
+                });
+        });
+    };
+
+    return (
+        <StyledContainer style={style} ref={container}>
```
TahimiLeonBravo: add the className prop

src/components/PresenceMap/component.js
```diff
@@ -0,0 +1,194 @@
...
+    isScriptLoaded: PropTypes.bool.isRequired,
+    isScriptLoadSucceed: PropTypes.bool.isRequired,
+    objects: PropTypes.arrayOf(
+        PropTypes.shape({
+            position: PropTypes.shape({
+                lat: PropTypes.number,
+                lng: PropTypes.number,
+            }),
+            heading: PropTypes.number,
+            image: PropTypes.oneOfType([PropTypes.string, PropTypes.object]),
```
LeandroTorresSicilia: here change to string only

src/components/PresenceMap/component.js
```diff
@@ -0,0 +1,194 @@
...
+    ),
+    style: PropTypes.object,
+    zoom: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
+    center: PropTypes.oneOfType([
+        PropTypes.shape({
+            lat: PropTypes.number,
+            lng: PropTypes.number,
+        }),
+        PropTypes.string,
+    ]),
```
LeandroTorresSicilia: if center will receive some string values like 'auto' we need to define them here

src/components/PresenceMap/component.js
```diff
@@ -0,0 +1,194 @@
...
+    showTraffic: PropTypes.bool,
+    showTransit: PropTypes.bool,
+    // eslint-disable-next-line
+    children: PropTypes.oneOfType([PropTypes.arrayOf(PropTypes.node), PropTypes.node]),
```
LeandroTorresSicilia: children are not being used in this component, we need to pass this prop in order to render things inside the map

src/components/PresenceMap/component.js
```diff
@@ -0,0 +1,194 @@
...
+                .catch(error => {
+                    // eslint-disable-next-line no-console
+                    console.log(`Error. Details: ${error}`);
+                });
+        });
+    };
+
+    return (
+        <StyledContainer style={style} ref={container}>
+            <StyledMapContainer
```
TahimiLeonBravo: I think we are adding an unnecessary wrapper to the component. 
why do we need to have two wrappers with practically the same styles?

src/components/PresenceMap/readme.md
rafaelmg83: I'll be adding features in the upcoming commits to achieve this goal ;-)

src/components/PresenceMap/readme.md
```diff
@@ -0,0 +1,183 @@
...
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { faCrosshairs } from '@fortawesome/free-solid-svg-icons';
+import { COLOR_WHITE } from '../../styles/colors';
+import { SHADOW_1 } from '../../styles/shadows';
+import { BORDER_RADIUS_1 } from '../../styles/borderRadius';
+
+const RegularControl = styled(Input)`
+    background-color: ${COLOR_WHITE};
+    border-radius: ${BORDER_RADIUS_1};
+    box-shadow: ${SHADOW_1};
```
LeandroTorresSicilia: make this example get values from the theme in this case shadow and background

src/components/PresenceMap/readme.md
```diff
@@ -0,0 +1,183 @@
...
+    text-align: center;
+    padding: 5px;
+    width: fit-content;
+    float: left;
+`;
+
+const CircularControl = styled.div`
+    background-color: ${COLOR_WHITE};
+    border-radius: 20px;
+    box-shadow: ${SHADOW_1};
```
LeandroTorresSicilia: same ^^

src/components/PresenceMap/readme.md
```diff
@@ -0,0 +1,183 @@
...
+            objects={objects}
+            showTraffic={showTrafficState}
+            showTransit={showTransitState}
+            zoom={zoomState}
+            center={centerState}
+            type={mapTypeState.name}
+        >
+            <MenuControl
+                id="picklist-mapstyle-1"
+                style={{ float: "right" }}
```
LeandroTorresSicilia: no style inline, use the styled inline, also I think we should avoid using the float property

src/components/PresenceMap/readme.md
```diff
@@ -0,0 +1,183 @@
...
+                onChange={value => handleMapStyle(value)}
+                value={mapTypeState}
+                label=""
```
LeandroTorresSicilia: pass a meaningful value to the label

src/components/PresenceMap/readme.md
```diff
@@ -0,0 +1,183 @@
...
+                onClick={handleShowTraffic}
+                checked={showTrafficState} />
+            <RegularControl
+                className="rainbow-m-around_medium"
+                style={{ float: "right" }}
+                type="checkbox"
+                label="Show Transit"
+                onClick={handleShowTransit}
+                checked={showTransitState} />
+            <CircularControl
```
LeandroTorresSicilia: Here we should use the ButtonIcon component which uses an interactive element like button, div is not meant to be clickable

src/components/PresenceMap/readme.md
```diff
@@ -0,0 +1,183 @@
...
+                className="rainbow-m-around_medium"
+                style={{ float: "right" }}
```
LeandroTorresSicilia: same ^^ for all styles inline

LeandroTorresSicilia:
- When no passing a the position to marker object the component break, and it only should not render the marker but not break
- markers onclick does not work, it should
- type proptype should have explicit what types it accepts like:
```
type: PropTypes.oneOf(['roadmap', 'satellite', ...]),
```
- talk to @TahimiLeonBravo in order to make the example like the design (markers with cars images, border radius of traffic control, icon in to picklist options in type control, etc)

TahimiLeonBravo:
@rafaelmg83 the example should be looks like the design

PR #1323, opened 2020-02-10
docs: new chart implementation example

blakmetall:
docs: new chart implementation example

@nexxtway/react-rainbow

### DIFF
--- a/src/components/Chart/readme.md
+++ b/src/components/Chart/readme.md
@@ -747,6 +747,12 @@ const data = {
     facebook: [30, 60, 68, 79, 100, 112, 100, 121, 130, 105, 145, 185]
 };
 
+const chartTypeMap = {
+    line: 'Lines',
+    horizontalBar: 'Horizontal Bar',
+    bar: 'Bar'
+};
+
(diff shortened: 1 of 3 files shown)


src/components/Chart/readme.md
```diff
@@ -705,3 +705,110 @@ class DoughnutChartExample extends React.Component {
...
+                            onChange={ (e) => setChartType(e.target.value) }
+                        />
+                    </ChartHeading>
+
+                    <CurrentChartType className="rainbow-flex rainbow-align-content_center rainbow-m-bottom_medium">
+                        { chartTypes.map( type => {
+                            if(type.value === chartType) {
+                                return type.label;
+                            }           
+                        })}
```
LeandroTorresSicilia: instead of making an interation here, you can use another state for the label or better you can make a map outside the component like:
```
const chartTypeMap = {
    line: 'Lines',
    horizontalBar: 'Horizontal Bar',
    bar: 'Bars',
}
```
and then use like:
```
<CurrentChartType className="rainbow-flex rainbow-align-content_center rainbow-m-bottom_medium">
    {chartTypeMap[chartType]}
```

PR #1324, opened 2020-02-10
docs: new checkbox group implementation

blakmetall:
docs: new checkbox group implementation

@nexxtway/react-rainbow

### DIFF
--- a/src/components/CheckboxGroup/readme.md
+++ b/src/components/CheckboxGroup/readme.md
@@ -218,8 +218,6 @@ const ProfileLeftCol = styled.div`
     }
 `;
 
-const ProfileRightCol = styled.div``;
-
 const options = [
     { value: 'push-notifications', label: 'Push Notifications', disabled: false },
     { value: 'email', label: 'Email', disabled: false },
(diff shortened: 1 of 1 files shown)


src/components/CheckboxGroup/readme.md
```diff
@@ -164,3 +164,117 @@ class CheckboxGroupTry extends React.Component {
...
+
+const ProfileLeftCol = styled.div`
+    min-width: 250px;
+
+    @media (max-width: 768px) {
+        margin-bottom: 20px;
+    }
+`;
+
+const ProfileRightCol = styled.div``;
```
LeandroTorresSicilia: if this does not have any style the remove it and use the div directly

src/components/CheckboxGroup/readme.md
```diff
@@ -164,3 +164,117 @@ class CheckboxGroupTry extends React.Component {
...
+];
+const defaultOptions = ['email'];
+
+
+const CheckBoxGroupProfile = () => {
+    const [values, setSelected] = useState([]);
+    
+    useEffect( () => {
+         setSelected(defaultOptions);   
+    }, []);
```
LeandroTorresSicilia: you can use:
```const [values, setSelected] = useState(defaultOptions)```
and then not need to use the useEffect

File: src/components/CheckboxGroup/readme.md [hunk 3]
```diff
@@ -164,3 +164,117 @@ class CheckboxGroupTry extends React.Component {
...
+
+                </ProfileLeftCol>
+
+                <ProfileRightCol>
+                    <CheckboxGroup
+                        id="checkbox-group-2"
+                        label="Checkbox Group Label"
+                        options={options}
+                        value={values}
+                        onChange={ (values) => setSelected(values) }
```
LeandroTorresSicilia: if arguments passed to setSelected are the same that onChange receive then you can use:
```onChange={setSelected}```

Issue #1265, opened 2020-01-15
test: add test to the theme functions


TahimiLeonBravo:
#### Add test
- [x] getBrightness
- [x] isDark
- [ ] attachThemeAttrs

### Define Specs
- [x] getBrightness - should return the right value
- [x] isDark - should return false with a light color
- [x] isDark - should return true with a dark color
- [ ] attachThemeAttrs - should return a merge object with props and  default themes values when props.theme.rainbow is an empty
- [ ] attachThemeAttrs - should return a merge object with props and props.theme.rainbow when props.theme.rainbow was passed

yvmunayev:
@LeandroTorresSicilia check the specs

LeandroTorresSicilia:
### Specs
## getBrightness:
- [ ] should return the right value
- [ ] should return the right value when color passed is hsl

## isDark
- [ ] should return false with a light color
- [ ] should return true with a dark color

## attachThemeAttrs
- [ ] should return a merged object with props and props.theme.rainbow
(Here we need only one test since getTheme needs to be mocked and only you need to test that the merge is made between props and what getTheme returns)

PR #1325, opened 2020-02-10
test: add test to the theme functions

yvmunayev:
fix: #1265 

- add test to getBrightness
-  add test to isDark
- fix attachThemeAttrs imports

@nexxtway/react-rainbow


src/components/VisualPickerOption/index.js
```diff
@@ -108,6 +108,7 @@ VisualPickerOption.propTypes = {
      * @ignore
      */
     children: PropTypes.oneOfType([PropTypes.arrayOf(PropTypes.node), PropTypes.object]),
+    ariaDescribedby: PropTypes.string,
```
LeandroTorresSicilia: this is already fixed, this is not a prop passed directly to the VisualPickerOption but a prop that comes in context from parent component (VisualPicker), since it is only internal we should not add it to the proptypes in order to not get it exposed as documentation for consumers

src/styles/helpers/__test__/getBrightness.spec.js
```diff
@@ -0,0 +1,59 @@
...
+            };
+            return result;
+        })
+        .mockImplementationOnce(() => {
+            const result = {
+                type: 'rgb',
+                values: [255, 255, 255],
+            };
+            return result;
+        }),
```
LeandroTorresSicilia: good job, for testing based on different decomposeColor outputs

Issue #1327, opened 2020-02-11
feat: implement color normalization


TahimiLeonBravo:
### Requirements
- [ ] Implement normalization func
- [ ] Add test to normalization

PR #1336, opened 2020-02-15
feat: implement color normalization

wildergd:
fix: #1327

- add colorToRgba helper
- add normalizeThemeColors helper

@nexxtway/react-rainbow


src/styles/helpers/color/colorToRgba.js
```diff
@@ -0,0 +1,22 @@
...
+    if (typeof color !== 'string') return '';
+
+    try {
+        const { type, values } = decomposeColor(color);
+
+        if (!['rgb', 'rgba', 'hsl', 'hsla'].includes(type)) return '';
+
+        if (type === 'rgb') return rgbToRgba(color);
+        if (type === 'hsl') return rgbToRgba(hslToRgb(color));
+        if (type === 'hsla') return hslToRgb(color);
```
LeandroTorresSicilia: in this case we will return rgb instead of rgba?

src/styles/helpers/color/colorToRgba.js
```diff
@@ -0,0 +1,22 @@
...
+
+        return recomposeColor({ type, values });
```
LeandroTorresSicilia: I assume that this is the rgba case, is not better to return the color directly without need to pass through recomposeColor?

src/styles/helpers/normalizeThemeColors.js
```diff
@@ -0,0 +1,25 @@
...
+            if (typeof value === 'string') {
+                normalizedPalette[key] = colorToRgba(value);
+            } else if (typeof value === 'object') {
+                normalizedPalette[key] = normalizePaletteColors(palette[key]);
+            } else {
+                normalizedPalette[key] = palette[key];
+            }
+        });
+        return normalizedPalette;
+    };
```
LeandroTorresSicilia: let's put this function outside normalizeThemeColors function but in this same file

src/styles/helpers/color/colorToRgba.js
wildergd: no the hslToRgb helper returns an rgba if hsla colors is passed as param

PR #1369, opened 2020-02-25
fix: pagination round item for single page for pagination component

blakmetall:
fix: pagination round item for single page for pagination component

- new helper to retrieve proper Border Radius according to pages length

@nexxtway/react-rainbow


src/components/Pagination/styled/pageButton.js
```diff
@@ -16,7 +17,7 @@ const StyledPageButton = styled.li`
 
     :nth-last-child(2) > button {
         margin-right: ${MARGIN_X_SMALL};
-        border-radius: 0 100px 100px 0;
+        border-radius: ${props => getBorderRadius(props.pages)};
```
LeandroTorresSicilia: There should be an approach that use only CSS, if not possible we can simply use:
```
function getBorderRadius() {
    if (pages === 1) {
        return '100px';
    }
    return '0 100px 100px 0';
}
 ```

blakmetall:
CSS selector - **:only-of-type**

Can't be used due to current markup; the pagination elements currently are inside a loop being called in another component,

resulting:

li > left arrow

li > pagination buttons component here

li > right arrow

-----

In order to be able to use :only-of-type selector we should apply to a parent component, which might result in css decoupling

PR #1370, opened 2020-02-25
style: add height and line height to rating component

blakmetall:
styles: add height and line height to rating component

--
* added line height to fieldset container; their childs will inherit those values

@nexxtway/react-rainbow


src/components/Rating/styled/fieldset.js
```diff
@@ -1,6 +1,7 @@
 import styled from 'styled-components';
 
 const StyledFieldset = styled.fieldset`
+    line-height: 1.5em;
```
LeandroTorresSicilia: I think we should use rem because it will have a more deterministic value since em will take the font size of the nearest parent it is more likely to change

Issue #1321, opened 2020-02-09
fix: the info header on experience example


TahimiLeonBravo:
### Requirements
- [ ] Change the badges by bold texts
https://zpl.io/adGORO7
- [ ] make the header texts custom with the theme
https://zpl.io/VQ7ry4k

TahimiLeonBravo:
<img width="1254" alt="Screen Shot 2020-02-08 at 9 03 20 PM" src="https://user-images.githubusercontent.com/32083556/74095453-bc303280-4ab6-11ea-821f-c825bb84f78d.png">

PR #1374, opened 2020-02-26
fix: info header on experience example

rolandin:
fix: #1321

* styles: changed styles on styled components

* fix: added needed key-value pairs to theme objects

* fix: StyledTitle and StyledBadge now take theme as prop

@nexxtway/react-rainbow

### DIFF
--- a/library/styleguideComponents/Wrapper/index.js
+++ b/library/styleguideComponents/Wrapper/index.js
@@ -35,6 +35,8 @@ const orangeTheme = {
             brand: '#f8752d',
             success: '#00a042',
             error: '#e53935',
+            header: 'rgb(248,117,45)',
+            badge: '#061c3f',
         },
     },
 };
@@ -44,6 +46,8 @@ const deepPurpleTheme = {
         palette: {
             brand: '#6860db',
             mainBackground: '#f2f2f2',
+            header: 'rgb(104,96,219)',
+            badge: '#061c3f',
         },
     },
 };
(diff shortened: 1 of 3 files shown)


library/styleguideComponents/Wrapper/index.js
```diff
@@ -65,6 +71,8 @@ const cyanDarkTheme = {
             success: '#b9f6ca',
             error: '#ff5252',
             mainBackground: '#303030',
+            header: 'rgba(255, 255, 255, 0.64)',
+            badge: 'white',
```
LeandroTorresSicilia: we should not add values that do not belong to the theme API

library/styleguideComponents/Wrapper/styled.js
```diff
@@ -60,7 +60,15 @@ export const StyledContentText = styled.h2`
 `;
 
 export const StyledBadge = styled(Badge)`
```
LeandroTorresSicilia: what we want is remove the Badge here and instead use something like:
```
export const StyledBadge = styled.span`
    // styles here
`
```
and then you can clean up all these styles and then use only the bold font
Also if you look at this file `src/styles/defaultTheme.js`
you can see the internal structure of the theme and access from the styled component to values like
```props.palette.text.main``` that already set the right text colors based on the main background

library/styleguideComponents/Wrapper/index.js
rolandin: Got it, updating that now

Issue #1341, opened 2020-02-20
feat: add prop `readOnly` to `Rating` component


TahimiLeonBravo:
### Requirements
- [x] It should have the default cursor when hover
- [x] It should not fill the star when hover
- [x] It should render semi-star filled when functionaries values are passed
- [x] It should be tested

##### rating
- [ ] should not set the value of the star hovered to the state when readOnly is true
##### ratingItems
- [ ] should render half star at the right position when value is a float
- [ ] should not render any half star when value is an integer
##### star
- [ ] should not fire onChange event when readOnly is true
- [ ] should render half star when filled and half props are true

PR #1376, opened 2020-02-26
feat: add prop `readOnly` to `Rating` component

HellWolf93:
fix: #1341

- Add prop 'readOnly' to Rating component
- Add component StarHalf, which renders a semi filled star
- Change Star component to render a StarHalf when value is a float
- Add example in docs
- Add tests to the new features

##### rating
- [x] should not set the value of the star hovered to the state when readOnly is true
##### ratingItems
- [x] should render half star at the right position when value is a float
- [x] should not render any half star when value is an integer
##### star
- [x] should not fire onChange event when readOnly is true
- [x] should render half star when filled and half props are true
___


src/components/Rating/readme.md
```diff
@@ -84,6 +84,93 @@ class SimpleRating extends React.Component {
...
+
+    handleOnChange(event) {
+        console.log(' changed! ');
+        return this.setState({ value: event.target.value });
+    }
+
+    render() {
+        return <Rating value={this.state.value} onChange={this.handleOnChange} readOnly />;
+    }
+}
```
TahimiLeonBravo: Since this example is `readOnly` we don't need to use a state, use directly the `Rating` component inside the Card

src/components/Rating/index.d.ts
```diff
@@ -6,6 +6,7 @@ export interface RatingProps extends BaseProps {
     onChange?: (event: ChangeEvent<HTMLElement>) => void;
     name?: string;
     label?: ReactNode;
+    readonly?: boolean;
```
LeandroTorresSicilia: readOnly

src/components/Rating/index.js
```diff
@@ -69,6 +75,8 @@ Rating.propTypes = {
     className: PropTypes.string,
     /** An object with custom style applied for the outer element. */
     style: PropTypes.object,
+    /** Whether it is read only or not */
```
LeandroTorresSicilia: ` Specifies that the Rating is read-only. This value defaults to false.`

src/components/Rating/__test__/ratingItems.spec.js
```diff
@@ -10,10 +10,30 @@ describe('<RatingItems />', () => {
...
+            half: false,
+            readOnly: false,
         });
     });
     it('should render the right amount of Star components', () => {
         const component = mount(<RatingItems />);
         expect(component.children().length).toBe(5);
     });
+    it('should render half star at the right position when value is a float', () => {
+        const component = mount(<RatingItems value={'3.5'} />);
```
LeandroTorresSicilia: here you can use a string directly:
```value="3.5"```

File: src/components/Rating/ratingItems.js [hunk 5]
```diff
@@ -3,15 +3,24 @@ import PropTypes from 'prop-types';
...
-    const { onChange, value, name } = props;
+    const { onChange, value, name, readOnly } = props;
 
     return Array(5)
         .fill(0)
         .map((item, index) => {
             const key = `star-${index}`;
-            const filled = index < value;
+            const filled = index < Math.ceil(value);
+            const half = value < index + 1 && !Number.isInteger(Number(value));
```
LeandroTorresSicilia: let's use isHalf and isFilled to express the intention of a boolean (normally we prefix all boolean variables names with is, are, or has)

src/components/Rating/__test__/ratingItems.spec.js
```diff
@@ -10,10 +10,30 @@ describe('<RatingItems />', () => {
...
+        const star = component.childAt(3);
+        expect(star.find('StarHalf').exists()).toBe(true);
+    });
```
LeandroTorresSicilia: I think this test belongs to ```src/components/Rating/star.js``` component since it is who decides what icon render
the test should check that the StarHalf icon is rendered when the half prop is passed

src/components/Rating/__test__/ratingItems.spec.js
```diff
@@ -10,10 +10,30 @@ describe('<RatingItems />', () => {
...
+            .fill(0)
+            .forEach((item, index) => {
+                expect(
+                    component
+                        .childAt(index)
+                        .find('StarHalf')
+                        .exists(),
+                ).toBe(false);
+            });
+    });
```
LeandroTorresSicilia: same ^^

src/components/Rating/__test__/ratingItems.spec.js
```diff
@@ -10,10 +10,30 @@ describe('<RatingItems />', () => {
...
 });
```
LeandroTorresSicilia: here we need to add test cases for the filled and half values passed like:
- should pass the right filled when the value passed is float
- should set half to true when the value passed is not integer

src/components/Rating/readme.md
```diff
@@ -84,6 +84,65 @@ class SimpleRating extends React.Component {
...
+        <Card
+            style={cardStyles}
+            footer={
+                <div>
+                    <div className="rainbow-flex rainbow-flex_column rainbow-align_center rainbow-m-bottom_x-small">
+                        <StyledHeader className="rainbow-font-size-heading_medium">
+                            Rainbow Image
+                        </StyledHeader>
+                    </div>
+                    <Rating value={'2.45'} readOnly label="2.45 of 5" />
```
LeandroTorresSicilia: same ^^ ```pass value="2.45"```

src/components/Rating/star.js
```diff
@@ -39,7 +40,8 @@ export default class Star extends Component {
                     id={this.starId}
                     value={value}
                     name={name}
-                    onChange={onChange}
+                    onChange={!readOnly ? onChange : null}
```
LeandroTorresSicilia: let's move this logic to a function and then:
```onChange={this.handleChange}```

File: src/components/Rating/__test__/star.spec.js [hunk 11]
```diff
@@ -26,18 +26,29 @@ describe('<Star />', () => {
         component.find('input').simulate('change');
         expect(onChangeFn).toHaveBeenCalledTimes(1);
     });
-    it('should render the StarFill when the prop filled is true', () => {
-        const component = mount(<Star filled />);
+    it('should render the StarFill when the prop isFilled is true', () => {
+        const component = mount(<Star isFilled />);
         component.find('input').simulate('change');
```
LeandroTorresSicilia: not need to simulate the change event here

src/components/Rating/__test__/star.spec.js
```diff
@@ -26,18 +26,29 @@ describe('<Star />', () => {
...
         expect(component.find('StarFill').exists()).toBe(true);
     });
-    it('should render the StarBordered when the prop filled is false', () => {
-        const component = mount(<Star filled={false} />);
+    it('should render the StarBordered when the prop isFilled is false', () => {
+        const component = mount(<Star isFilled={false} />);
         component.find('input').simulate('change');
```
LeandroTorresSicilia: same ^^

PR #1383, opened 2020-03-01
fix: add the option selected to the dropdown on the Picklist option

HellWolf93:
fix: #1362

## Changes proposed in this PR:
- Add selected option to Picklist dropdown
- Add checkmark to selected PicklistOption
- Remove tests that checked PicklistOption register and unregister
- Add test for the checkmark

#### Test specs
- [ ] should render a checkmark icon when selected
____
- [X] I have followed (at least) the [PR section of the contributing guide](https://github.com/nexxtway/react-rainbow/blob/master/CONTRIBUTING.md#submitting-a-pull-request).

PR #1387, opened 2020-03-03
fix: review getPageBorderRadius testing from Pagination component

blakmetall:
* filename renamed from filename.js. to filename.spec.js
* test updated

@nexxtway/react-rainbow


blakmetall:
Integration tests are failing over a non related component for the current updates

PickListOption has recent updates, they might be causing the failed integration tests

Related pull-request:
https://github.com/nexxtway/react-rainbow/pull/1383

PR #1390, opened 2020-03-04
fix: remove integration test 'should hide the option selected'

HellWolf93:
- Removed integration test that checked the option selected were hidden


integration/specs/Picklist/picklist-1.spec.js
```diff
@@ -99,7 +99,7 @@ describe('Picklist base example', () => {
         option.hover();
         expect(option.isActive()).toBe(true);
     });
-    it('should set active the first option when other is active and close the menu and open it again', () => {
+    it.skip('should set active the first option when other is active and close the menu and open it again', () => {
```
LeandroTorresSicilia:
```suggestion
    it('should set active the first option when other is active and close the menu and open it again', () => {
```

Issue #1329, opened 2020-02-12
feat: add test for `attachThemeAttrs` function


TahimiLeonBravo:
#### Specs
 - [ ] attachThemeAttrs - should return a merge object with props and default themes values when props.theme.rainbow is an empty
  - [ ] attachThemeAttrs - should return a merge object with props and props.theme.rainbow when props.theme.rainbow was passed

PR #1391, opened 2020-03-05
feat: add test for `attachThemeAttrs` function

HellWolf93:
fix: #1329

#### Test specs
attachThemeAttrs
- should return a merge object with props and default themes values when props.theme.rainbow is empty
- should return a merge object with props and props.theme.rainbow when props.theme.rainbow was passed

### DIFF
--- a/src/styles/helpers/__test__/attachThemeAttrs.spec.js
+++ b/src/styles/helpers/__test__/attachThemeAttrs.spec.js
@@ -1,0 +1,42 @@
+import React from 'react';
+import styled from 'styled-components';
+import { mount } from 'enzyme';
+import attachThemeAttrs from '../attachThemeAttrs';
+import defaultTheme from '../../defaultTheme';
+
+describe('attachThemeAttrs', () => {
+    let interpolatedProps = null;
+    const interpolatedFn = props => {
+        interpolatedProps = props;
+    };
+    const StyledComponent = attachThemeAttrs(styled.div)`
+        position: relative;
+        ${props => interpolatedFn(props)}
+    `;
+
+    beforeEach(() => {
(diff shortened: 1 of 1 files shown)


src/styles/helpers/__test__/attachThemeAttrs.spec.js
```diff
@@ -0,0 +1,42 @@
...
+    `;
+
+    beforeEach(() => {
+        interpolatedProps = null;
+    });
+
+    it('should return a merge object with props and default themes values when props.theme.rainbow is an empty', () => {
+        mount(<StyledComponent test />);
+        expect(interpolatedProps.test).toBe(true);
+        expect(interpolatedProps.palette).toBe(defaultTheme.palette);
```
LeandroTorresSicilia: toBe is used to compare primitve values like boolean, string, number, etc and toEqual is used to compare objects and arrays.
Since in javascript object are compared by reference in this case toBe make a shallow equal comparison and how they have the same reference (defaultTheme) then it will pass. In order to make a real check of all values we need to use toEqual which make a deep comparison

src/styles/helpers/__test__/attachThemeAttrs.spec.js
```diff
@@ -0,0 +1,42 @@
...
+                        main: {
+                            color: 'red',
+                        },
+                    },
+                },
+            },
+        };
+        mount(<StyledComponent test theme={theme} />);
+        expect(interpolatedProps.test).toBe(true);
+        expect(interpolatedProps.palette).toBe(theme.rainbow.palette);
```
LeandroTorresSicilia: same ^^

Issue #921, opened 2019-09-13
feat: add Payment component with Stripe


TahimiLeonBravo:
#### Steps
- [x] Define Component API 
https://docs.google.com/document/d/1G5h9gDSRDeuCSnK4C74GGa3NJ0HTA-gS5IRO3IxpdkI/edit?usp=sharing
- [x] Implement component
- [x] Add tests

### Design
https://zpl.io/VD9yd6l

### Test (add test-cases)
- [x]  should fire onChange with specific event

PR #1396, opened 2020-03-07
feat: add Payment component with Stripe

yvmunayev:
fix: #921

- feat: add Payment component with Stripe

@nexxtway/react-rainbow


src/components/PaymentStripe/index.d.ts
```diff
@@ -0,0 +1,16 @@
+import { ReactNode, MouseEvent, FormEvent } from 'react';
+import { PaymentIntent, StripeError } from '@stripe/stripe-js';
+import { BaseProps } from '../types';
+
+export interface PaymentStripe extends BaseProps {
+    apiKeyGMap?: string;
```
LeandroTorresSicilia: let's use a simple Input instead of GoogleAddressLookup and then remove this prop

src/components/PaymentStripe/checkoutForm.js
```diff
@@ -0,0 +1,370 @@
+import React, { useState } from 'react';
```
maxxgreene: @TahimiLeonBravo @LeandroTorresSicilia should we name the component `StripeCheckout` instead?

src/libs/hooks/useLocale.js
```diff
@@ -0,0 +1,8 @@
+import { useContext } from 'react';
+import { AppContext } from '../../components/Application/context';
+import getBrowserLocale from '../utils/getBrowserLocale';
+
+export default function useLocale(localProp) {
+    const context = useContext(AppContext);
+    return localProp || (context && context.locale) || getBrowserLocale();
+}
```
maxxgreene: beautiful!

src/libs/hooks/useTheme.js
```diff
@@ -0,0 +1,17 @@
+import { useContext } from 'react';
+import { ThemeContext } from 'styled-components';
+import normalizeThemeColors from '../../styles/helpers/normalizeThemeColors';
+import defaultTheme from '../../styles/defaultTheme';
+
+export default function useTheme() {
+    const themeContext = useContext(ThemeContext);
+    if (!themeContext) {
```
maxxgreene: avoid negation in an `if` ... it reads better the other way around 
```js
if (themeContext) {
    ...
}
return ...
```

src/components/PaymentStripe/index.js
```diff
@@ -0,0 +1,85 @@
...
+import CheckoutForm from './checkoutForm';
+
+/**
+ * Payment Stripe component are used for freeform data entry.
+ * @category Form
+ */
+const PaymentStripe = props => {
+    const { apiKeyStripe, className, style, ...rest } = props;
+    const locale = useLocale();
+    const elementOptions = { locale };
```
maxxgreene: we can memo this value here as well, if not it will pass a different object `options` every time the function runs

src/components/PaymentStripe/checkoutForm.js
```diff
@@ -0,0 +1,370 @@
...
+    } = props;
+    const stripe = useStripe();
+    const elements = useElements();
+    const theme = useTheme().rainbow;
+    const [loading, setLoading] = useState(true);
+    const [billingDetails, setBillingDetails] = useState(initialBillingDetail);
+    const [cardDetails, setCardDetails] = useState({});
+    const [rememberMe, setRememberMe] = useState(false);
+    const [error, setError] = useState();
+    const cardElementOptions = {
```
maxxgreene: every time we create an object in a component that we know we can memoize I think we should because it will be created differently every time and it will be pass to a child component, then the framework it won't be able to do proper optimization because the component is receiving a different object

src/components/PaymentStripe/checkoutForm.js
```diff
@@ -0,0 +1,370 @@
...
+        }
+    };
+
+    return (
+        <Container inline={inline}>
+            <RenderIf isTrue={loading}>
+                <Spinner size="large" />
+            </RenderIf>
+            <form onSubmit={handleSubmit}>
+                <div style={loading ? loadingStyle : null}>
```
maxxgreene: I'm not sure if we should even show spinner while the stripe SDK is loading or while stripe is resolving the token ... I believe we can just no render anything until stripe is ready ... and disabled the form elements while the stripe is resolving the token

src/components/PaymentStripe/checkoutForm.js
```diff
@@ -0,0 +1,370 @@
...
+    };
+    const handleSubmit = async event => {
+        event.preventDefault();
+        if (!stripe || !elements || !getPaymentIntent || !onProcessed) {
+            return;
+        }
+        try {
+            setLoading(true);
+            // TODO: validate fields and normalize address
+            const response = await getPaymentIntent();
```
maxxgreene: paymentIntent are made from server-side, what is the use case for this in the component?

src/components/PaymentStripe/checkoutForm.js
```diff
@@ -0,0 +1,370 @@
...
+            const cardElement = inline
+                ? elements.getElement(CardElement)
+                : elements.getElement(CardNumberElement);
+            const result = await stripe.confirmCardPayment(response.client_secret, {
```
maxxgreene: uumm, let's talk about it

src/components/PaymentStripe/cardSection.js
```diff
@@ -0,0 +1,136 @@
...
+            hidePostalCode,
+            hideIcon,
+            iconStyle,
+            disabled,
+            value: { postalCode },
+        }),
+        [theme, hidePostalCode, hideIcon, iconStyle, disabled, postalCode],
+    );
+    useEffect(() => {
+        if (!elements) {
```
maxxgreene: The other way around, avoid negation in if conditions 
```
if (elements) {
...
}
```

src/components/PaymentStripe/cardSection.js
```diff
@@ -0,0 +1,136 @@
...
+        const cardElement = elements.getElement(CardElement);
+        if (cardElement) {
+            cardElement.update({
+                value: { postalCode },
+            });
+        }
+    }, [elements, postalCode]);
+
+    const handleChange = event => {
+        if (!stripe || !elements) {
```
maxxgreene: i"m not sure if we are being over defensive here, I guess that onChange is not going to be fired if the elements and stripe are not ready, in fact, the component shouldn't render if those values are not ready. But if I'm wrong ... make the if condition different

```js
if (stripe && elements) {
...
}
```

src/components/PaymentStripe/cardSection.js
```diff
@@ -0,0 +1,136 @@
...
+            return;
+        }
+        const payment = {
+            ...event,
```
maxxgreene: just `event` to avoid name conflicts if event data change at stripe implement end

src/components/PaymentStripe/cardSection.js
```diff
@@ -0,0 +1,136 @@
...
+        onChange(payment);
+    };
+    return (
+        <StyledContainer className={className} style={style}>
+            <Label label={label} hideLabel={hideLabel} inputId={cardElementId} required />
+            <CardElement
+                id={cardElementId}
+                options={cardElementOptions}
+                onChange={handleChange}
+                onReady={() => onReady()}
```
maxxgreene: let remove the onReady callback since we are sending the event in onChange, it will be more deterministic providing a single way to know the same

src/components/PaymentStripe/cardSection.js
```diff
@@ -0,0 +1,136 @@
...
+            value: { postalCode },
+        }),
+        [theme, hidePostalCode, hideIcon, iconStyle, disabled, postalCode],
+    );
+    useEffect(() => {
+        if (!elements) {
+            return;
+        }
+        const cardElement = elements.getElement(CardElement);
+        if (cardElement) {
```
maxxgreene: what is the use case for pre-filling the postalCode?? let's remove this logic as well, if needed we can add it later

src/components/PaymentStripe/cardSection.js
```diff
@@ -0,0 +1,136 @@
...
+                    color: theme.palette.text.main,
+                    fontSize: '1.1rem',
+                    '::placeholder': {
+                        color: theme.palette.text.header,
+                        fontVariant: 300,
+                    },
+                },
+            },
+            hidePostalCode,
+            hideIcon,
```
maxxgreene: let's remove the logic for removing the postalCode and hide the card icon if we have the use case in future we add it

src/components/PaymentStripe/index.js
```diff
@@ -0,0 +1,68 @@
...
+    /** Text label for the component. */
+    label: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
+    /** A boolean to hide the input label. */
+    hideLabel: PropTypes.bool,
+    /** Shows the help message below the input. */
+    bottomHelpText: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
+    /** Specifies that an input field must be filled out before submitting the form. */
+    error: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
+    /** Appearance of the icon in the Card Element. */
+    iconStyle: PropTypes.oneOf(['default', 'solid']),
```
maxxgreene: let's remove this as well

src/components/PaymentStripe/cardSection.js
```diff
@@ -0,0 +1,117 @@
...
+                empty: event.empty,
+                complete: event.complete,
+                brand: event.brand,
+                error: event.error
+                    ? {
+                          code: event.error.code,
+                          type: event.error.type,
+                          message: event.error.message,
+                      }
+                    : undefined,
```
maxxgreene: move the massage you are doing to the event  to a helper functions

src/components/PaymentStripe/index.d.ts
```diff
@@ -0,0 +1,33 @@
...
+import { BaseProps } from '../types';
+
+interface Payment {
+    stripe?: Stripe;
+    element?: StripeElement;
+    empty?: boolean;
+    complete?: boolean;
+    brand?: 'visa' | 'mastercard' | 'amex' | 'discover' | 'diners' | 'jcb' | 'unionpay' | 'unknown';
+    error?:
+        | undefined
```
maxxgreene: if it's optional(?) then we don't need to specify undefined as type, right?

src/components/PaymentStripe/cardSection.js
```diff
@@ -0,0 +1,117 @@
...
+    );
+
+    const handleChange = event => {
+        if (stripe && elements) {
+            const payment = {
+                stripe,
+                element: elements.getElement(CardElement),
+                empty: event.empty,
+                complete: event.complete,
+                brand: event.brand,
```
maxxgreene: maybe a good opportunity to improve the names of stripe props
- empty -> isEmpty 
- complete -> isComplete
- brand -> cardBrand

src/components/StripeCardInput/__test__/stripeCardInput.spec.js
```diff
@@ -0,0 +1,52 @@
...
+describe('<StripeCardInput>', () => {
+    it('should fire onChange with specific event', () => {
+        const onChangeMockFn = jest.fn();
+        const component = mount(
+            <StripeCardInput apiKey="STRIPE_API_KEY" onChange={onChangeMockFn} />,
+        );
+        component.find('input').simulate('change', event);
+        expect(onChangeMockFn).toHaveBeenCalledWith({
+            brand: 'unknown',
+            iEmpty: false,
```
maxxgreene: typo?

src/components/StripeCardInput/cardInput.js
```diff
@@ -0,0 +1,98 @@
...
+    const cardElementOptions = useMemo(() => getCardElementOptions(theme, disabled), [
+        theme,
+        disabled,
+    ]);
+    const handleChange = event => {
+        if (stripe && elements) {
+            const card = {
+                stripe,
+                element: elements.getElement(CardElement),
+                iEmpty: event.empty,
```
maxxgreene: yes, typo

maxxgreene: How a consumer of this component it will use the object payload is being sent in the onChange callback

```js
const handleOnChange =  async (event) => {
     const { stripe, element } = event;
     ...
     const result = await stripe.confirmCardPayment('{CLIENT_SECRET}', {
          payment_method: {
              card: element,
              billing_details: {
                  name: 'Jenny Rosen',
              },
        }
    });
}

<StripeCardInput ... onChange={handleOnChange} />
```

my point the other day was to make the props in the event exactly like the props stripe use in their api then 

```js
const handleOnChange =  async (event) => {
     const { stripe, card } = event;
     ...
     const result = await stripe.confirmCardPayment('{CLIENT_SECRET}', {
          payment_method: {
              `card`,
              billing_details: {
                  name: 'Jenny Rosen',
              },
        }
    });
}
```

also `brand` prop could be changed to `cardBrand`

### suggested changes 

```js
const handleChange = event => {
        if (stripe && elements) {
            onChange({
                stripe,
                card: elements.getElement(CardElement),
                isEmpty: event.empty,
                isComplete: event.complete,
                cardBrand: event.brand,
                error: getError(event.error),
            };);
        }
    };
```

src/components/StripeCardInput/cardInput.js
```diff
@@ -0,0 +1,98 @@
...
+    const stripe = useStripe();
+    const elements = useElements();
+    const cardElementId = useUniqueIdentifier('card-element');
+    const theme = useTheme().rainbow;
+    const cardElementOptions = useMemo(() => getCardElementOptions(theme, disabled), [
+        theme,
+        disabled,
+    ]);
+    const handleChange = event => {
+        if (stripe && elements) {
```
maxxgreene: is there onchange event happening before stripe and elements got ready?? if so, should we render nothing (null) until those object are ready?? I think we are being defensive here where we dont need it

src/components/StripeCardInput/helpers/getErrror.js
```diff
@@ -0,0 +1,9 @@
+export default function getError(error) {
+    return error
+        ? {
+              code: error.code,
+              type: error.type,
+              message: error.message,
+          }
+        : undefined;
```
maxxgreene:
```suggestion
    if (error) {
        return {
              code: error.code,
              type: error.type,
              message: error.message,
          };
    }
    return undefined;
      
```

src/components/StripeCardInput/cardInput.js
```diff
@@ -0,0 +1,96 @@
...
+    const theme = useTheme().rainbow;
+    const cardElementOptions = useMemo(() => getCardElementOptions(theme, disabled), [
+        theme,
+        disabled,
+    ]);
+    const handleChange = event => {
+        const stripeCardEvent = {
+            stripe,
+            card: elements.getElement(CardElement),
+            iEmpty: event.empty,
```
maxxgreene:
```suggestion
            isEmpty: event.empty,
```

src/components/StripeCardInput/__test__/stripeCardInput.spec.js
```diff
@@ -0,0 +1,52 @@
...
+describe('<StripeCardInput>', () => {
+    it('should fire onChange with specific event', () => {
+        const onChangeMockFn = jest.fn();
+        const component = mount(
+            <StripeCardInput apiKey="STRIPE_API_KEY" onChange={onChangeMockFn} />,
+        );
+        component.find('input').simulate('change', event);
+        expect(onChangeMockFn).toHaveBeenCalledWith({
+            cardBrand: 'unknown',
+            iEmpty: false,
```
maxxgreene:
```suggestion
            isEmpty: false,
```

src/components/StripeCardInput/index.js
```diff
@@ -0,0 +1,65 @@
...
+    /** The action triggered when some value of the component changes. */
+    onChange: PropTypes.func,
+    /** The action triggered when the element receives focus. */
+    onFocus: PropTypes.func,
+    /** The action triggered when the element releases focus. */
+    onBlur: PropTypes.func,
+    /** A CSS class for the outer element, in addition to the component's base classes. */
+    className: PropTypes.string,
+    /** An object with custom style applied to the outer element. */
+    style: PropTypes.object,
```
TahimiLeonBravo: should we add the `required` prop?

TahimiLeonBravo:
@yvmunayev 
The placeholder styles should be the same to the styles on the placeholder on the Input components
Screen Shot 2020-03-25 at 11 44 26 PM

src/components/StripeCardInput/index.js
```diff
@@ -0,0 +1,65 @@
...
+import { useLocale } from '../../libs/hooks';
+import CardInput from './cardInput';
+
+/**
+ * Stripe Card Input component are used for freeform data entry.
+ * @category Form
+ */
+const StripeCardInput = props => {
+    const { apiKey, ...rest } = props;
+    const locale = useLocale();
```
TahimiLeonBravo: should we add a `locale` prop?

src/components/StripeCardInput/index.js
```diff
@@ -0,0 +1,65 @@
...
+    /** Specifies that an input field must be filled out before submitting the form. */
+    error: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
+    /** Specifies that an input element should be disabled. This value defaults to false. */
+    disabled: PropTypes.bool,
+    /** The action triggered when some value of the component changes. */
+    onChange: PropTypes.func,
+    /** The action triggered when the element receives focus. */
+    onFocus: PropTypes.func,
+    /** The action triggered when the element releases focus. */
+    onBlur: PropTypes.func,
```
TahimiLeonBravo: should we add the `tabIndex` prop?

src/components/StripeCardInput/index.js
```diff
@@ -0,0 +1,92 @@
...
+const stripeApiV3Url = 'https://js.stripe.com/v3';
+
+/**
+ * Stripe Card Input component are used for freeform data entry.
+ * @category Form
+ */
+const StripeCardInput = props => {
+    const { ...rest } = props;
+
+    const Component = useCallback(scriptLoader(stripeApiV3Url)(CardInput), []);
```
maxxgreene: if it doesn't depend on any props why is inside of the component, it could be another `const`, right?

maxxgreene:
@yvmunayev disabled is not working as expected, the placeholders should remain

src/components/StripeCardInput/index.d.ts
```diff
@@ -0,0 +1,63 @@
+import { ReactNode } from 'react';
+import { Stripe, StripeCardElement } from '@stripe/stripe-js';
+import { BaseProps } from '../types';
+
+type Error = {
```
maxxgreene: we should export this I guess, also it's to generic name, `StripeCardError`

src/components/StripeCardInput/index.d.ts
```diff
@@ -0,0 +1,63 @@
...
+    type: 'validation_error';
+    code: string;
+    message: string;
+};
+interface StripeCardEvent {
```
maxxgreene: same export this

src/components/StripeCardInput/index.js
```diff
@@ -0,0 +1,197 @@
...
+                };
+                onChange(stripeCardEvent);
+            });
+
+            card.on('focus', onFocus);
+            card.on('blur', onBlur);
+            const handlerBlur = () => card.blur();
+            const handlerFocus = () => card.focus();
+            cardNode.addEventListener('blur', handlerBlur);
+            cardNode.addEventListener('focus', handlerFocus);
```
maxxgreene: why do you need a `ref` to listen for focus and `blur` ?? it should work on the JSX like onFocus onBlur ... although is this really necessary?? you are can listen for those events in the card element as well, right?

maxxgreene: I see what you are doing passing through the focus the stripe input inside ... first, the blur logic is wrong .. think about it ... secund that focus dance is not that simple, what happens with (shift + tab) ??

Issue #1372, opened 2020-02-25
feat: add the possibility of pass a label to the ButtonMenu


TahimiLeonBravo:
As a customer, I want to have the possibility to pass a label with an icon to the ButtonMenu component

#### Specs
- [ ]  Add the ```label``` prop to the ButtonMenu. It should be of type ```node``` or ```string```
- [ ] Add the ```iconPosition``` prop to the ButtonMenu. The values to pass to this props are ```right``` and ```left``` default to ```left```
- [ ] make tests for the new functionality

#### Requirements
- [ ] as a customer if I pass only an icon then to the component the ButtonMenu should render a ButtonIcon with the icon passed.
- [ ] as a customer if I pass only a label then to the component the ButtonMenu should render a Button component with the text passed on the label.
- [ ] as a customer if I pass a label and an icon it should render a button with an icon and a text but default the icon is rendered on the left.
- [ ] as a customer if I want to pass an icon on the right side to the button I need to use ```iconPosition``` prop with ```right``` as a value.

##### buttonMenu
- [ ] should render a ButtonIcon when no label is passed
- [ ] should render a ButtonTrigger when label is passed
- [ ] should pass the right icon, label and iconPosition to ButtonTrigger
##### buttonTrigger
- [ ] should render a Button with label only when no icon is passed
- [ ] should render a Button with label and left icon when iconPosition value is different from ```right```
- [ ] should render a Button with label and right icon when iconPosition value is ```right```

PR #1397, opened 2020-03-08
feat: add the possibility of pass a label to the ButtonMenu

HellWolf93:
fix: #1372

- Add the label prop to the ButtonMenu. It should be of type node or string
- Add the iconPosition prop to the ButtonMenu. The values to pass to this props are right and left default to left 

##### buttonMenu
- [ ] should render a ButtonIcon when no label is passed
- [ ] should render a ButtonTrigger when label is passed
- [ ] should pass the right icon, label and iconPosition to ButtonTrigger
##### buttonTrigger
- [ ] should render a Button with label only when no icon is passed
- [ ] should render a Button with label and left icon when iconPosition value is different from ```right```
- [ ] should render a Button with label and right icon when iconPosition value is ```right```
____


TahimiLeonBravo:
@HellWolf93 let's change the current ```buttonVariant``` description by 
```The variant changes the look of the button. Accepted variants include base, brand, success, destructive, neutral, outline-brand, border, border-filled, inverse and border-inverse.

One of: base, brand, success, destructive, neutral, outline-brand, border, border-filled, border-inverse, inverse
```

src/components/ButtonMenu/index.js
```diff
@@ -37,31 +44,38 @@ export default function ButtonMenu(props) {
...
+     * This value defaults to left. */
+    iconPosition: PropTypes.oneOf(['left', 'right']),
     /** The content of the ButtonMenu. Used to render the menuItem elements
      * when the ButtonMenu is open. */
     children: PropTypes.oneOfType([PropTypes.arrayOf(PropTypes.node), PropTypes.object]),
-    /** The variant changes the look of the button. Accepted variants include base,
-     * brand, success, border, border-filled, bare-inverse, and inverse.
-     * This value defaults to border-filled. */
+    /** The variant changes the look of the button.
+     * Accepted variants include base, brand, success, destructive, neutral, outline-brand, border, border-filled, inverse and border-inverse. */
```
TahimiLeonBravo: Let's add ``This value defaults to border-filled.`` to the end of the description.

src/components/ButtonMenu/buttonTrigger.js
```diff
@@ -0,0 +1,39 @@
...
+import RenderIf from '../RenderIf';
+
+export default function ButtonTrigger(props) {
+    const { label, icon, iconPosition, ...rest } = props;
+    const hasLeftIcon = iconPosition !== 'right';
+
+    const marginClass = hasLeftIcon ? 'rainbow-m-right_xx-small ' : 'rainbow-m-left_xx-small ';
+    const className =
+        icon && icon.props.className ? marginClass.concat(icon.props.className) : marginClass;
+    const iconElement = icon !== null ? React.cloneElement(icon, { className }) : null;
```
LeandroTorresSicilia: This implies that the user should be responsible to pass an icon that have the className prop passed as className attr to its container. Since we should not depend on this let's add a wrapper element to the icon with the margin styles.
Also, note that we use styled-components for all our styles (https://styled-components.com/).

PR #1414, opened 2020-03-11
fix: improve header styles when table has scroll in table component

blakmetall:
fix: improve header styles when table has scroll in table component

**update comments:**
* new styled component to render shadow added at the bottom of each direct child from th inside the table; receives hasScroll prop.
* was not possible to apply box-shadow to direct th childs due to box-shadow nature

@nexxtway/react-rainbow


src/components/Table/index.js
```diff
@@ -221,6 +222,16 @@ export default class Table extends Component {
...
     }
 
+    detectActiveScroll() {
+        if (this.scrollableY.current) {
+            const { clientHeight, scrollHeight } = this.scrollableY.current;
+            const hasScroll = clientHeight < scrollHeight;
+            this.setState({ hasScroll });
+        } else {
+            this.setState({ hasScroll: false });
+        }
```
LeandroTorresSicilia: let's avoid using state for this, instead make a method that simply returns a boolean

PR #1417, opened 2020-03-12
fix: scroll on background of ImportRecordFlow

blakmetall:
fix: scroll on background of ImportRecordFlow

**information**

-- when a modal is present inside a page and then is opened and closed, **componentDidUpdate** resets scroll. (fixed to reset only where there is no modal opened).

-- when a modal is opened and then we open a new modal above and then close that modal, **componentDidUnmount** resets the scroll when it should not due to the previous opened modal.
(before: condition to reset all body scrolls was triggered even if there are modals in **CounterManager**).

-- when a previous modal was opened after a closed one above, the status **isOpen** and **prevIsOpen** from first modal opened are **false**, this was not reseting the scroll inside **componentDidUpdate** due to the previous modal was losing their **prevIsOpen** status.

-- **componentWillUnmount** triggers until leaving the page, and if multiple modals were opened the last one was not reseting inside **componentWillUnmount** neither inside **componentDidUpdate**.

**fix applied:**

-- inside **componentWillUnmount** **clearAllBodyScrollLocks()**  was moved inside CounterManager condition

-- added a new catch status inside **componendDidUpdate** when modal is updated and there is no previous interaction on it (**isOpen and prevIsOpen to false**), to solve the lost interactions values after open a new modal over an opened modal.

@nexxtway/react-rainbow


LeandroTorresSicilia:
I found a problem:
- go to http://localhost:6060/#!/Modal/11
- press Open Modal
- press Select a Time field to open the time picker modal
- click in backdrop two times to close the two modals
- try to scroll the page

current behavior:
- scroll is blocked

expected behavior:
- scroll works fine

since we don't have a test for this case we should add one

src/components/Modal/__test__/modal.spec.js
```diff
@@ -172,6 +171,47 @@ describe('<Modal/>', () => {
...
+        const component = mount(
+            <Modal>
+                <p />
+            </Modal>,
+        );
+        component.setProps({
+            isOpen: false,
+        });
+        expect(enableBodyScroll).not.toHaveBeenCalled();
+    });
```
LeandroTorresSicilia, line 214: Sorry, what I meant with the test that we should add was integration test, then remove these units since above they are already tested.
Also, the issue described continues.

integration/specs/Modal/modal-11.spec.js
```diff
@@ -1,14 +1,28 @@
...
 
 const BUTTON = '#button-11';
 const MODAL = '#modal-11';
+const MODAL_DATEPICKER = '#modal-datepicker-11';
 const TITLE_INPUT = '#modal-11 input[placeholder="Enter company name"]';
 const DATE_PICKER_INPUT = '#modal-11 input[placeholder="Select a date"]';
 const TIME_PICKER_INPUT = '#modal-11 input[name="time"]';
 const MODAL_LOOKUP = '#modal-lookup-11';
 
+const getScrollTop = () => {
```
LeandroTorresSicilia, line 14: If the intent of this is return the scroll top position then let's call it `getScrollTopPosition`

PR #1419, opened 2020-03-13
fix: improve ImportRecordsFlow component

blakmetall:
fix: improve ImportRecordsFlow component

- bold styles added to assign field title
- apply different UI when dragging a file

@nexxtway/react-rainbow


src/components/ImportRecordsFlow/stepTwo/styled/container.js
```diff
@@ -1,17 +1,11 @@
...
-    const { light: brandLightColor } = brand;
-
-    return {
-        brandLightColor,
-    };
-})`
+const StyledContainer = attachThemeAttrs(styled.div)`
     border-radius: 4px;
-    border: dashed 1px #eaeef5;
+    border: dashed 3px;
```
TahimiLeonBravo: let's keep the `border-width` of 1px

TahimiLeonBravo:
@blakmetall follow styles like the design, the message content should be centered on the container

Screen Shot 2020-03-17 at 9 34 08 AM

src/components/ImportRecordsFlow/stepThree/assignFieldModalTitle.js
```diff
@@ -0,0 +1,25 @@
...
+import StyledAssignFieldTitle from './styled/assignFieldTitle';
+
+const AssignFieldModalTitle = props => {
+    const { field } = props;
+    return (
+        <StyledAssignFieldHeader>
+            <StyledAssignFieldTitle>
+                Assign to the field: <b>{field}</b>
+            </StyledAssignFieldTitle>
+        </StyledAssignFieldHeader>
```
LeandroTorresSicilia: can we use only one wrapper like:
```
<StyledAssignFieldTitle>
    Assign to the field: <b>{field}</b>
</StyledAssignFieldTitle>
```

src/components/ImportRecordsFlow/stepTwo/index.js
```diff
@@ -68,8 +72,23 @@ export default function StepTwo(props) {
             onDrop={handleDrop}
             draggable
         >
-            <UploadFileButton onChange={handleChange} />
-            <StyledText>Find the file in your desktop.</StyledText>
+            <StyledUploadFileWrapper isDragOver={isDragOver}>
+                <UploadFileButton onChange={handleChange} />
+            </StyledUploadFileWrapper>
```
LeandroTorresSicilia: if you want to hide UploadFileButton when isDragOver you can pass styles directly inside UploadFileButton or simply use a RenderIf, there is no need to create an extra wrapper

src/components/ImportRecordsFlow/stepTwo/styled/text.js
```diff
@@ -10,6 +11,8 @@ const StyledText = styled.h1`
     margin: 0;
     padding: 0;
     box-sizing: border-box;
+
+    ${props => props.isDragOver && `display: none;`};
```
LeandroTorresSicilia: let's use a RenderIf to render this markup when isDragOver

src/components/ImportRecordsFlow/stepTwo/styled/text.js
```diff
@@ -1,6 +1,7 @@
 import styled from 'styled-components';
+import attachThemeAttrs from '../../../../../src/styles/helpers/attachThemeAttrs';
 
-const StyledText = styled.h1`
+const StyledText = attachThemeAttrs(styled.h1)`
```
LeandroTorresSicilia: let's use a p instead of h1 since it is not a heading

src/components/ImportRecordsFlow/stepThree/assignFieldModalTitle.js
blakmetall: we can

src/components/ImportRecordsFlow/stepTwo/index.js
blakmetall: that will be better

src/components/ImportRecordsFlow/stepTwo/styled/text.js
```diff
@@ -1,7 +1,7 @@
 import styled from 'styled-components';
-import attachThemeAttrs from '../../../../styles/helpers/attachThemeAttrs';
+import attachThemeAttrs from '../../../../../src/styles/helpers/attachThemeAttrs';
```
LeandroTorresSicilia: here you can
```suggestion
import attachThemeAttrs from '../../../../styles/helpers/attachThemeAttrs';
```

Issue #1411, opened 2020-03-11
test: add unit and integration test to Drawer component


TahimiLeonBravo:
#### Unit tests specs
- [x] should render the children passed
- [x] should render the footer passed
- [x] should set tabIndex as -1 in section element
- [x] should set the same generated id to section as aria-labelledby and to Header as id
- [x] should set the same generated id to section as aria-describedby and to children container as id
- [x] should set aria-modal to true in section element
- [x] should set aria-hidden to false in section element when drawer is open
- [x] should fire an event when the close button is clicked and the drawer is open
- [x] should fire an event when the drawer backdrop is clicked and the drawer is open
- [x] should fire an event when the ESC key is pressed and the drawer is open
- [x] should call disableBodyScroll when open the drawer
- [x] should call enableBodyScroll when drawer is closed
- [x] should call CounterManager.decrement when the component unmount and it is open
- [x] should not call CounterManager.decrement when the component unmount and it is not open
- [x] should render the close button by default
- [x] should not render the close button when the hideCloseButton prop is equal to true
- [x] should set the right animation states when opened
- [x] should set the right animation states when closed
- [x] should call onOpened when drawer state switch to OPENED
- [x] shoudl call manageTab when tab key is pressed

#### Integration tests proposal
[example #1]
- [x] should open the drawer
- [x] should close the drawer when is opened and press ESC
- [x] should return focus to trigger element when close the drawer with ESC key
- [x] should focus the close button when the drawer is opened and press TAB
- [x] should close the drawer when is opened and click the closeButton
- [x] should close the drawer when click outside
- [x] should return focus to trigger element when close the drawer with close button

[example #3]
- [x] should return focus to the close button when tabbing all elements in drawer
- [x] should not close the drawer when is opened and press ESC if the lookup has value typed
- [x] should not close the drawer when close the date picker input modal 
- [x] should close the drawer when is opened and press ESC if the lookup value typed was clear
- [x] should close the drawer when select an option and then press ESC
- [x] should close the drawer when select an option with keyboard and then press ESC

PR #1420, opened 2020-03-14
test: drawer component

wildergd1980:
fix: #1411 

- add unit tests for Drawer component
- add integration tests for Drawer component

@nexxtway/react-rainbow

### DIFF
--- a/src/components/Drawer/__test__/drawer.a11y.spec.js
+++ b/src/components/Drawer/__test__/drawer.a11y.spec.js
@@ -5,18 +5,16 @@ import Drawer from '..';
 import Button from '../../Button';
 
 describe('<Drawer/>', () => {
-    it('should be accessible', () => {
+    it('should be accessible', async () => {
+        expect.assertions(1);
         const wrapper = mount(
-            <Drawer isOpen title="Drawer a11y" footer={<Button label="button in drawer footer" />}>
+            <Drawer isOpen header="Drawer a11y" footer={<Button label="button in drawer footer" />}>
                 <p>A rainbow is a meteorological phenomenon ...</p>
             </Drawer>,
         );
 
-        setTimeout(async () => {
-            expect.assertions(1);
-            const html = wrapper.html();
-            const results = await axe(html);
-            expect(results).toHaveNoViolations();
-        }, 500);
+        const html = wrapper.html();
+        const results = await axe(html);
+        expect(results).toHaveNoViolations();
     });
 });
--- a/src/components/Drawer/__test__/drawer.a11y.spec.js
+++ b/src/components/Drawer/__test__/drawer.a11y.spec.js
@@ -11,6 +11,7 @@ describe('<Drawer/>', () => {
                 <p>A rainbow is a meteorological phenomenon ...</p>
             </Drawer>,
         );
+
         setTimeout(async () => {
             expect.assertions(1);
             const html = wrapper.html();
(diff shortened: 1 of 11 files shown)


src/components/Drawer/__test__/drawer.a11y.spec.js
```diff
@@ -0,0 +1,21 @@
...
+            <Drawer isOpen title="Drawer a11y" footer={<Button label="button in drawer footer" />}>
+                <p>A rainbow is a meteorological phenomenon ...</p>
+            </Drawer>,
+        );
+        setTimeout(async () => {
+            expect.assertions(1);
+            const html = wrapper.html();
+            const results = await axe(html);
+            expect(results).toHaveNoViolations();
+        }, 500);
```
maxxgreene:
```suggestion
    Promise.resolve().then(() => {
        expect.assertions(1);
        const html = wrapper.html();
        const results = await axe(html);
        expect(results).toHaveNoViolations();                
    });
```

maxxgreene: same for all the setTimeout, it should work with the promise resolving in the next tick

src/components/Drawer/__test__/drawer.spec.js
```diff
@@ -0,0 +1,262 @@
...
+    });
+
+    it('should render the children passed', () => {
+        const component = mount(
+            <Drawer isOpen>
+                <p data-id="drawer-children">drawer content</p>
+            </Drawer>,
+        );
+        setTimeout(() => {
+            expect(component.find('p[data-id="drawer-children"]').text()).toBe('drawer content');
```
maxxgreene: The use of `expect.assertions(1);` is important because ensure the assertions always runs

src/components/Drawer/__test__/drawer.a11y.spec.js
maxxgreene: if setTimeout is totally necessary because of the animation of the drawer then you should use the `done` function inside of the setTimeout

```js
it('should', (done) => {
    ...
    setTimeout(() => {
        ...
        done();
    }, 500);
});
```

src/components/Drawer/__test__/drawer.spec.js
wildergd: so it should be include in all tests?

src/components/Drawer/__test__/drawer.a11y.spec.js
wildergd: yes i think i will use the done approach as long as the setTimeout is mandatory due to the animation...

wildergd: by using the promise approach on accessibility test.. it failed... and also fail with the done approach... and it fails because "wrapper.html()" returns null... not sure why...  the only way it runs is the way i implement it... not sure what i am doing wrong..

src/components/Drawer/__test__/drawer.spec.js
LeandroTorresSicilia: it is only needed for async testing:
https://jestjs.io/docs/en/asynchronous

src/components/Drawer/index.js
```diff
@@ -138,6 +138,7 @@ export default function Drawer(props) {
                     <Header id={headerId} content={header} />
                     <RenderIf isTrue={!hideCloseButton}>
                         <StyledCloseButton
+                            data-id="drawer-close-button"
```
LeandroTorresSicilia: since this is a ButtonIcon data-id will not work because it does not have a data-id prop, instead you should use id prop

Issue #1427, opened 2020-03-19
fix: type inference in ColumnProps


Jyozi:
-   [x] I have searched the [issues](https://github.com/nexxtway/react-rainbow/issues) of this repository and believe that this is not a duplicate.

## Describe the bug

Type inference is not working for `component` in ColumnProps.

## To Reproduce

1. Use the Typescript
2. Use the Column component and `component` prop. (ex. [Simple Column](https://react-rainbow.firebaseapp.com/#/Colum))
3. See error

## Expected Behavior

Quote from https://react-rainbow.firebaseapp.com/#/Column

```
const data = [
  {
    name: 'Leandro Torres',
    company: 'Nexxtway',
    email: 'leandro@gmail.com',
    status: 'verified',
  },
]

const badgeStyles = { color: '#1de9b6' }

type StatusBadgeProp = {
  value: string
}
const StatusBadge: React.FC<StatusBadgeProp> = (
  props: StatusBadgeProp
): JSX.Element => (
  <Badge label={props.value} variant="lightest" style={badgeStyles} />
)

export const SimpleColumn: React.FC = () => (
  <div className="rainbow-p-bottom_xx-large">
    <Table data={data} keyField="name">
      <Column header="Name" field="name" />
      <Column header="Status" field="status" component={StatusBadge} />
      <Column header="Company" field="company" />
      <Column header="Email" field="email" />
    </Table>
  </div>
)

```

### error

```
no error
```

## Current Behavior
```
export const SimpleColumn: React.FC = () => (
  <div className="rainbow-p-bottom_xx-large">
    <Table data={data} keyField="name">
      <Column header="Name" field="name" />  
      <Column header="Status" field="status" component={StatusBadge} />
                                         {/* ↑ type error */}
      <Column header="Company" field="company" />
      <Column header="Email" field="email" />
    </Table>
  </div>
)
```

### error

```
Type 'FC<StatusBadgeProp>' is not assignable to type 'ComponentClass<{}, any> | FunctionComponent<{}> | undefined'.
Type 'FunctionComponent<StatusBadgeProp>' is not assignable to type 'FunctionComponent<{}>'.
Types of parameters 'props' and 'props' are incompatible.
Type '{ children?: ReactNode; }' is not assignable to type 'PropsWithChildren<StatusBadgeProp>'.
Property 'value' is missing in type '{ children?: ReactNode; }' but required in type 'StatusBadgeProp'.
```

## Environment

-   OS: linux, unix
[...]

PR #1428, opened 2020-03-19
fix: type inference in column props

Jyozi:
fix: #1427 

- add generics to ColumnProps

@nexxtway/react-rainbow

### DIFF
--- a/src/components/Column/index.d.ts
+++ b/src/components/Column/index.d.ts
@@ -10,4 +10,4 @@ export interface ColumnProps<P> {
     type?: 'text' | 'action';
 }
 
-export default function<P>(props: ColumnProps<P>  & P): JSX.Element | null;
+export default function<P>(props: ColumnProps<P> & P): JSX.Element | null;
--- a/src/components/Column/index.d.ts
+++ b/src/components/Column/index.d.ts
@@ -10,4 +10,4 @@ export interface ColumnProps<P> {
     type?: 'text' | 'action';
 }
 
-export default function<P>(props: ColumnProps<P>): JSX.Element | null;
+export default function<P>(props: ColumnProps<P>  & P): JSX.Element | null;
--- a/src/components/Column/index.d.ts
+++ b/src/components/Column/index.d.ts
@@ -1,8 +1,8 @@
 import { ReactNode, ComponentType } from 'react';
 
-export interface ColumnProps {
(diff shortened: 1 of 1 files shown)


src/components/Column/index.d.ts
```diff
@@ -1,13 +1,13 @@
...
+    component?: ComponentType<P>;
     field?: string;
     sortable?: boolean;
     width?: number | string;
     defaultWidth?: number | string;
     type?: 'text' | 'action';
 }
 
-export default function(props: ColumnProps): JSX.Element | null;
+export default function<P>(props: ColumnProps<P>): JSX.Element | null;
```
LeandroTorresSicilia: we need to let the user pass extra props to Column
```suggestion
export default function<P>(props: ColumnProps<P>  & P): JSX.Element | null;
```

LeandroTorresSicilia:
Hi @Jyozi thanks for the PR
I made a simple change:
```export default function<P>(props: ColumnProps<P>  & P): JSX.Element | null;```
we need to add `& P` to the props interface to let the user pass extra props to Column component

Keep in mind that you need to use it as:
```
const data = [
  {
    name: 'Leandro Torres',
    company: 'Nexxtway',
    email: 'leandro@gmail.com',
    status: 'verified',
  },
]

const badgeStyles = { color: '#1de9b6' }

type StatusBadgeProp = {
  value?: string
}
const StatusBadge: React.FC<StatusBadgeProp> = (
  props: StatusBadgeProp
): JSX.Element => (
  <Badge label={props.value} variant="lightest" style={badgeStyles} />
)

export const SimpleColumn: React.FC = () => (
  <div className="rainbow-p-bottom_xx-large">
    <Table data={data} keyField="name">
      <Column header="Name" field="name" />
      <Column<StatusBadgeProps> header="Status" field="status" component={StatusBadge} />
      <Column header="Company" field="company" />
      <Column header="Email" field="email" />
    </Table>
  </div>
)

```
two differences from your example
1. <Column<StatusBadgeProps> header="Status" field="status" component={StatusBadge} />
you need to pass the StatusBadgeProps interface declaration to Column
2. value?: string
you need to set value as optional since the user can pass the data without the status value:
```
const data = [
  {
    name: 'Leandro Torres',
    company: 'Nexxtway',
    email: 'leandro@gmail.com',
  },
]
```

Issue #1421, opened 2020-03-16
feat: implement ButtonGroupPicker


TahimiLeonBravo:
#### API
- https://docs.google.com/document/d/1_6CpzbFYH2nNEwf3-uH0JwkfbOZcohU_g_po2AzZyEs/edit
- https://docs.google.com/document/d/1WtLmGEooaFjZtuiq5NU6lCOdGyA777EkoOlhJi6t2mc/edit

#### Design
https://zpl.io/bAyv4oO

#### Requirements
- [ ] It should allow selecting only one option.
- [ ] It should allow multiple options at the same time.
- [ ] It should highlight the selected options.
- [ ]  It should allow disabled how many options do you want.
- [ ] It should show a message if an error occurred.

#### Accessibility
We should follow the w3c accessibility requirements
https://www.w3.org/

Also, we should review the slds 
https://www.lightningdesignsystem.com/components/checkbox-button-group/
https://www.lightningdesignsystem.com/components/radio-button-group/

PR #1434, opened 2020-03-24
feat: implement ButtonGroupPicker

HellWolf93:
fix: #1421

- Add component ButtonGroupPicker
- Add component ButtonOption


src/components/ButtonGroupPicker/index.js
```diff
@@ -0,0 +1,125 @@
...
+                <RenderIf isTrue={!!error}>
+                    <ErrorText id={this.getErrorMessageId()}>{error}</ErrorText>
+                </RenderIf>
+            </StyledContainer>
+        );
+    }
+}
+
+ButtonGroupPicker.propTypes = {
+    className: PropTypes.string,
```
TahimiLeonBravo: We need to add the descriptions of the props :)
Let's use the descriptions defined on the API doc.

src/components/ButtonGroupPicker/index.js
```diff
@@ -0,0 +1,125 @@
...
+import PropTypes from 'prop-types';
+import StyledContainer from './styled/container';
+import StyledButtonGroup from './styled/buttonGroup';
+import StyledLegend from './styled/label';
+import { uniqueId } from '../../libs/utils';
+import { Provider } from './context';
+import RenderIf from '../RenderIf';
+import HelpText from './styled/helpText';
+import ErrorText from './styled/errorText';
+
```
TahimiLeonBravo: Add the description of the component
```
/**
 * ButtonGroupPicker can be used to group related options. The ButtonGroupPicker will control the selected state of its child ButtonOption.
 */
```

src/components/ButtonGroupPicker/readme.md
```diff
@@ -0,0 +1,151 @@
...
+
+    handleOnChange(values) {
+        this.setState({ values });
+    }
+
+    render() {
+        const { values } = this.state;
+        return (
+            <div className="rainbow-p-vertical_large rainbow-align-content_center rainbow-flex_wrap">
+                <ButtonGroupPicker onChange={this.handleOnChange} value={values} name="filter" size="medium" label="Select view type" bottomHelpText="Select one option">
```
TahimiLeonBravo:
```suggestion
                <ButtonGroupPicker
                    onChange={this.handleOnChange}
                    value={values}
                    name="filter"
                    size="medium"
                    label="Select view type"
                    bottomHelpText="Select one option">
```

TahimiLeonBravo: Let's change to this way

src/components/ButtonGroupPicker/readme.md
```diff
@@ -0,0 +1,151 @@
...
+
+    handleOnChange(values) {
+        this.setState({ values });
+    }
+
+    render() {
+        const { values } = this.state;
+        return (
+            <div className="rainbow-p-vertical_large rainbow-align-content_center rainbow-flex_wrap">
+                <ButtonGroupPicker className="rainbow-m-around_medium" multiple onChange={this.handleOnChange} value={values} name="text-options-multiple">
```
TahimiLeonBravo: same ^^

src/components/ButtonGroupPicker/readme.md
```diff
@@ -0,0 +1,151 @@
...
+
+    handleOnChange(values) {
+        this.setState({ values });
+    }
+
+    render() {
+        const { values } = this.state;
+        return (
+            <div className="rainbow-p-vertical_large rainbow-align-content_center rainbow-flex_wrap">
+                <ButtonGroupPicker className="rainbow-m-around_medium" onChange={this.handleOnChange} value={values} name="text-options-multiple">
```
TahimiLeonBravo: same ^^

src/components/ButtonGroupPicker/readme.md
```diff
@@ -0,0 +1,151 @@
...
+
+    handleOnChange(values) {
+        this.setState({ values });
+    }
+
+    render() {
+        const { values } = this.state;
+        return (
+            <div className="rainbow-p-vertical_large rainbow-align-content_center rainbow-flex_wrap">
+                <ButtonGroupPicker onChange={this.handleOnChange} value={values} name="text-options-multiple" label="Select view type" error="This field is required">
```
TahimiLeonBravo: same ^^

src/components/ButtonOption/index.js
```diff
@@ -0,0 +1,53 @@
...
+            />
+            <StyledText onClick={onClick} disabled={disabled} checked={checked}>
+                {label}
+            </StyledText>
+        </StyledLabel>
+    );
+}
+
+ButtonOption.propTypes = {
+    className: PropTypes.string,
```
TahimiLeonBravo: We need to add the descriptions of the props :)
Let's use the descriptions defined on the API doc.

src/components/ButtonGroupPicker/context.js
```diff
@@ -0,0 +1,4 @@
+import React from 'react';
+
+export const ButtonGroupPickerContext = React.createContext();
+export const { Provider, Consumer } = ButtonGroupPickerContext;
```
LeandroTorresSicilia: you can make this in one line:
```export const { Provider, Consumer } = React.createContext();```

File: src/components/ButtonGroupPicker/index.d.ts [hunk 9]
```diff
@@ -0,0 +1,17 @@
+import { BaseProps } from '../types';
+import { ChangeEvent, ReactNode, ComponentType } from 'react';
```
LeandroTorresSicilia: for convention let's put absolute imports before relative ones

src/components/ButtonGroupPicker/index.d.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export interface ButtonGroupPickerProps extends BaseProps {
+    label: string | ReactNode;
```
LeandroTorresSicilia: here use only ReactNode since it already includes the string type

src/components/ButtonGroupPicker/index.d.ts
```diff
@@ -0,0 +1,17 @@
...
+    value: string | string[];
+    onChange: (event: ChangeEvent<HTMLElement>) => void;
```
LeandroTorresSicilia: here instead of an event we should send the value which is:
```type Value = string[] | string;```
then you can:
```
value?: Value;
onChange?: (value: Value) => void;
```

File: src/components/ButtonGroupPicker/index.d.ts [hunk 12]
```diff
@@ -0,0 +1,17 @@
...
+    multiple: boolean;
+    children: ReactNode;
+    name: string;
+    size: string;
```
LeandroTorresSicilia: here we need to be more specific:
```size?:  'x-small' | 'small' | 'medium' |  'large'```

File: src/components/ButtonGroupPicker/index.d.ts [hunk 13]
```diff
@@ -0,0 +1,17 @@
...
+    error: string | ReactNode;
```
LeandroTorresSicilia: same ^^ only ReactNode

src/components/ButtonGroupPicker/index.d.ts
```diff
@@ -0,0 +1,17 @@
...
+    bottomHelpText: string | ReactNode;
```
LeandroTorresSicilia: same ^^

src/components/ButtonGroupPicker/index.d.ts
```diff
@@ -0,0 +1,17 @@
...
+}
```
LeandroTorresSicilia: all props should be optional:
```
multiple?: boolean;
children?: ReactNode;
```

src/components/ButtonGroupPicker/index.js
```diff
@@ -0,0 +1,145 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import StyledContainer from './styled/container';
+import StyledButtonGroup from './styled/buttonGroup';
+import StyledLegend from './styled/label';
+import { uniqueId } from '../../libs/utils';
+import { Provider } from './context';
+import RenderIf from '../RenderIf';
+import HelpText from './styled/helpText';
+import ErrorText from './styled/errorText';
```
LeandroTorresSicilia: since HelpText and ErrorText are the same in all inputables we can reuse the ones defined inside Input component in
`src/components/Input/styled/`

src/components/ButtonGroupPicker/index.js
```diff
@@ -0,0 +1,145 @@
...
+        this.handleOnChange = this.handleOnChange.bind(this);
+    }
+
+    getValue() {
+        const { value } = this.props;
+        if (typeof value === 'string') {
+            return [];
+        }
+        return value;
+    }
```
LeandroTorresSicilia: we don't need to do this since value can be a string, when multiple is passed value will be an array of string and when not it will be a simple string with the name of the selected option

src/components/ButtonGroupPicker/index.js
```diff
@@ -0,0 +1,145 @@
...
+        const { error } = this.props;
+        if (error) {
+            return this.errorMessageId;
+        }
+        return undefined;
+    }
+
+    handleOnChange(event) {
+        const { value, checked } = event.target;
+        const { value: values, multiple, onChange } = this.props;
```
LeandroTorresSicilia: for better readability we can use value normally and value from event as eventValue:
```
const { value: eventValue, checked } = event.target;
const { value, multiple, onChange } = this.props;
```

src/components/ButtonGroupPicker/index.js
```diff
@@ -0,0 +1,145 @@
...
+
+        if (!multiple) {
+            return onChange([value]);
```
LeandroTorresSicilia: when no multiple we should send the event value direclty:
```return onChange(eventValue);```

File: src/components/ButtonGroupPicker/index.js [hunk 20]
```diff
@@ -0,0 +1,145 @@
...
+            size,
+            error,
+            bottomHelpText,
+        } = this.props;
+
+        const context = {
+            onChange: this.handleOnChange,
+            values: this.getValue(),
+            type: multiple ? 'checkbox' : 'radio',
+            name: name || uniqueId('option'),
```
LeandroTorresSicilia: here we need to pass groupNameId defined in constructor:
```this.groupNameId```

File: src/components/ButtonGroupPicker/index.js [hunk 21]
```diff
@@ -0,0 +1,145 @@
...
+            ariaDescribedBy: this.getErrorMessageId(),
```
LeandroTorresSicilia: in order to avoid calling getErrorMessageId two times in each rerender you can put it in a constant to use here and in the ErrorText

src/components/ButtonGroupPicker/index.js
```diff
@@ -0,0 +1,145 @@
...
+        };
+
+        return (
+            <StyledContainer className={className} style={style}>
+                <RenderIf isTrue={!!label}>
+                    <StyledLegend>{label}</StyledLegend>
+                </RenderIf>
+                <StyledButtonGroup size={size} role="group">
+                    <Provider value={{ ...context }}>{children}</Provider>
```
LeandroTorresSicilia: you can use:
```value={context}```

File: src/components/ButtonGroupPicker/index.js [hunk 23]
```diff
@@ -0,0 +1,145 @@
...
+            name: name || uniqueId('option'),
+            ariaDescribedBy: this.getErrorMessageId(),
+        };
+
+        return (
+            <StyledContainer className={className} style={style}>
+                <RenderIf isTrue={!!label}>
+                    <StyledLegend>{label}</StyledLegend>
+                </RenderIf>
+                <StyledButtonGroup size={size} role="group">
```
LeandroTorresSicilia: remove role="group" since here is not needed

src/components/ButtonGroupPicker/index.js
```diff
@@ -0,0 +1,145 @@
...
+    error: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
+    /** Shows the help message below the ButtonGroupPicker. */
+    bottomHelpText: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
+};
+
+ButtonGroupPicker.defaultProps = {
+    className: undefined,
+    style: undefined,
+    label: undefined,
+    value: [],
```
LeandroTorresSicilia: let's use undefined as default value

src/components/ButtonGroupPicker/readme.md
```diff
@@ -0,0 +1,171 @@
+##### ButtonGroupPicker base
+
+```js
+import React from 'react';
+import { ButtonGroupPicker, ButtonOption } from 'react-rainbow-components';
+
+class ButtonGroupPickerTry extends React.Component {
+    constructor(props) {
+        super(props);
+        this.state = { values: [] };
```
LeandroTorresSicilia: it is confusing using an empty array in the value state when is not multiple

src/components/ButtonGroupPicker/readme.md
```diff
@@ -0,0 +1,171 @@
...
+        const { values } = this.state;
+        return (
+            <div className="rainbow-p-vertical_large rainbow-align-content_center rainbow-flex_wrap">
+                <ButtonGroupPicker 
+                    className="rainbow-m-around_medium" 
+                    value={values} 
+                    onChange={this.handleOnChange} 
+                    name="text-options-multiple"
+                >
+                    <ButtonOption label={<FontAwesomeIcon icon={faAlignLeft} />} name="left" disabled />
```
LeandroTorresSicilia: disabled styles are not working as expected, for example, it changes icon color on hover, and if it is pressed it makes the animation, I am pretty sure this should change but check with @TahimiLeonBravo what should be the behavior

src/components/ButtonGroupPicker/styled/label.js
```diff
@@ -0,0 +1,22 @@
...
+    color: ${props => props.palette.text.label};
+    font-size: ${FONT_SIZE_TEXT_MEDIUM};
+    line-height: 1.5;
+    margin: 0 auto 0.1rem auto;
+    text-align: center;
+    box-sizing: border-box;
+    ${props =>
+        props.variant === 'inverse' &&
+        `
+            color: ${props.palette.isDark ? '#576574' : 'rgb(178,178,178)'}
```
LeandroTorresSicilia: we can't use colors inline, all colors must come from theme, here you can use some of the text variants in the theme like ```props.palette.text.title``` or other that match with the design color

src/components/ButtonOption/index.d.ts
```diff
@@ -0,0 +1,12 @@
+import { BaseProps } from '../types';
+import { ReactNode, MouseEvent, ComponentType } from 'react';
+
+export interface ButtonOptionProps extends BaseProps {
+    name: string;
+    label: string | ReactNode;
+    onClick: (event: MouseEvent<HTMLElement>) => void;
+    disabled: boolean;
+}
```
LeandroTorresSicilia: same ^^ all options should be optionals

src/components/ButtonOption/index.d.ts
```diff
@@ -0,0 +1,12 @@
+import { BaseProps } from '../types';
+import { ReactNode, MouseEvent, ComponentType } from 'react';
+
+export interface ButtonOptionProps extends BaseProps {
+    name: string;
+    label: string | ReactNode;
```
LeandroTorresSicilia: use only ReactNode

src/components/ButtonOption/index.d.ts
```diff
@@ -0,0 +1,12 @@
...
+    onClick: (event: MouseEvent<HTMLElement>) => void;
```
LeandroTorresSicilia: here what we send is not the native event but the an object like: ```{ selected:  true }```

src/components/ButtonOption/index.d.ts
```diff
@@ -0,0 +1,12 @@
+import { BaseProps } from '../types';
+import { ReactNode, MouseEvent, ComponentType } from 'react';
```
LeandroTorresSicilia: same ^^ change the order

src/components/ButtonOption/index.js
```diff
@@ -0,0 +1,59 @@
...
+import PropTypes from 'prop-types';
+import { ButtonGroupPickerContext } from '../ButtonGroupPicker/context';
+import HiddenElement from '../Structural/hiddenElement';
+import StyledLabel from './styled/label';
+import StyledText from './styled/text';
+import isOptionSelected from './helpers/isOptionSelected';
+import { uniqueId } from '../../libs/utils';
+
+export default function ButtonOption(props) {
+    const inputId = uniqueId('button-option');
```
LeandroTorresSicilia: this will cause the inputId change in each rerender, that is why we normally use this in the constructor
for functional components using hooks we created a custom hook that get called only onces
`import { useUniqueIdentifier } from '../../libs/hooks';`
use this instead

src/components/ButtonOption/index.js
```diff
@@ -0,0 +1,59 @@
...
+    const { className, style, name, label, disabled, onClick } = props;
+
+    const context = useContext(ButtonGroupPickerContext);
+    const checked = isOptionSelected(context.values, name);
+
+    return (
+        <StyledLabel className={className} style={style} checked={checked} for={inputId}>
```
LeandroTorresSicilia: I am not sure but I think react uses htmlFor instead of the native for

src/components/ButtonOption/index.js
```diff
@@ -0,0 +1,59 @@
...
+import StyledLabel from './styled/label';
+import StyledText from './styled/text';
+import isOptionSelected from './helpers/isOptionSelected';
+import { uniqueId } from '../../libs/utils';
+
+export default function ButtonOption(props) {
+    const inputId = uniqueId('button-option');
+    const { className, style, name, label, disabled, onClick } = props;
+
+    const context = useContext(ButtonGroupPickerContext);
```
LeandroTorresSicilia: here you can:
```const { value, type, name ... } = useContext(ButtonGroupPickerContext);```
in order to avoid repeating context. each time a prop from the context is used

File: src/components/ButtonGroupPicker/index.js [hunk 35]
```diff
@@ -0,0 +1,145 @@
...
+     * The size of the ButtonOption.
+     * Options includes x-small, small, medium and large.
+     * This value defaults to medium.
+     */
+    size: PropTypes.oneOf(['x-small', 'small', 'medium', 'large']),
+    /** pecifies that a ButtonGroupPicker must be filled out before submitting the form. */
+    error: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
+    /** Shows the help message below the ButtonGroupPicker. */
+    bottomHelpText: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
+};
```
LeandroTorresSicilia: here we should add the required prop and then render inside the Legend the RequiredAsterisk component if required is true

src/components/ButtonGroupPicker/index.js
```diff
@@ -0,0 +1,145 @@
...
+    onChange: () => {},
+    multiple: false,
+    name: undefined,
+    size: 'medium',
+    error: undefined,
+    bottomHelpText: undefined,
+    children: undefined,
+};
+
+export default ButtonGroupPicker;
```
LeandroTorresSicilia: since this is a form component we need to export it through withReduxForm as:
```export default withReduxForm(ButtonGroupPicker);```
this is used to make seamless integration with redux-form library

File: src/components/ButtonOption/styled/text.js [hunk 37]
```diff
@@ -0,0 +1,110 @@
...
+    }
+    return {
+        inverse: {
+            text: COLOR_WHITE,
+            active: COLOR_GRAY_3,
+            border: COLOR_WHITE,
+            disabled: COLOR_GRAY_3,
+        },
+    };
+})`
```
LeandroTorresSicilia: I think we should not do this here, it was needed in another component for some transparency issues but here I think we should pass fixed colors from theme, talk to @TahimiLeonBravo about this if you have doubt

src/components/index.js
```diff
@@ -10,8 +10,10 @@ export { default as Breadcrumb } from './Breadcrumb';
 export { default as Breadcrumbs } from './Breadcrumbs';
 export { default as Button } from './Button';
 export { default as ButtonGroup } from './ButtonGroup';
+export { default as ButtonGroupPicker } from './ButtonGroupPicker';
 export { default as ButtonIcon } from './ButtonIcon';
 export { default as ButtonMenu } from './ButtonMenu';
+export { default as ButtonOption } from './ButtonOption';
```
LeandroTorresSicilia: we need to make this same exports in the global index.d.ts

src/components/ButtonGroupPicker/styled/buttonGroup.js
```diff
@@ -0,0 +1,84 @@
+/* stylelint-disable no-descending-specificity */
+import styled from 'styled-components';
+
+const StyledButtonGroup = styled.div`
+    display: inline-flex;
+
+    span {
```
TahimiLeonBravo: We should not do introspection on the component's styles.  Instead, to use nested selectors we should make the `span` a styled

src/components/ButtonGroupPicker/styled/buttonGroup.js
```diff
@@ -0,0 +1,84 @@
...
+        border-radius: 0;
+        border-width: 1px;
+        margin-left: -1px;
+        height: 2.5rem;
+        line-height: 2.5;
+
+        svg,
```
TahimiLeonBravo: same^^

TahimiLeonBravo:
@HellWolf93 we need add a file .md with a simple example on the ButtonOption component

src/components/ButtonGroupPicker/index.js
```diff
@@ -0,0 +1,150 @@
...
+
+        if (!multiple) {
+            return onChange(eventValue);
+        }
+
+        if (checked && Array.isArray(value)) {
+            return onChange(value.concat([eventValue]));
+        }
+        if (checked && !Array.isArray(value)) {
+            return onChange(eventValue);
```
LeandroTorresSicilia, line 47: What happens when the user does not pass an initial value and then it is undefined or if it passes a wrong value initially?
I think here we should do the following:
```suggestion
            return onChange([eventValue]);
```
In this way, we can recover from the initial undefined or wrong value and the component will work as expected.

src/components/ButtonGroupPicker/index.js
```diff
@@ -0,0 +1,150 @@
...
+
+ButtonGroupPicker.propTypes = {
+    /** A CSS class for the outer element, in addition to the component's base classes. */
+    className: PropTypes.string,
+    /** An object with a custom style applied to the outer element. */
+    style: PropTypes.object,
+    /** The title at the top of the component. */
+    label: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
+    /** The name of the ButtonOption selected or if multiple an array of names. */
+    value: PropTypes.oneOfType([PropTypes.string, PropTypes.array]).isRequired,
```
LeandroTorresSicilia, line 106: Normally all props are not required, only carefully discussed props are required.
```suggestion
    value: PropTypes.oneOfType([PropTypes.string, PropTypes.array]),
```

src/components/ButtonGroupPicker/index.js
```diff
@@ -0,0 +1,150 @@
...
+     * Defaults to false.
+     */
+    multiple: PropTypes.bool,
+    /**
+     * This prop that should not be visible in the documentation.
+     * @ignore
+     */
+    children: PropTypes.oneOfType([PropTypes.arrayOf(PropTypes.node), PropTypes.object]),
+    /** The name of the options inside the form. */
+    name: PropTypes.string.isRequired,
```
LeandroTorresSicilia, line 122: same ^^
```suggestion
    name: PropTypes.string,
```

src/components/ButtonOption/index.js
```diff
@@ -0,0 +1,62 @@
...
+    );
+}
+
+ButtonOption.propTypes = {
+    /** A CSS class for the outer element, in addition to the component's base classes. */
+    className: PropTypes.string,
+    /** An object with a custom style applied to the outer element. */
+    style: PropTypes.object,
+    /** It is a unique value that identifies the picker option. */
+    name: PropTypes.string.isRequired,
```
LeandroTorresSicilia, line 46: same ^^
```suggestion
    name: PropTypes.string,
```

Issue #1422, opened 2020-03-17
feat: render a spinner when the Tree is loading


TahimiLeonBravo:
A Spinner should be rendered on the arrow button when it is loading the children

###Design
https://zpl.io/blE8m9D
<img width="914" alt="Screen Shot 2020-03-17 at 5 41 52 PM" src="https://user-images.githubusercontent.com/32083556/76912076-34d59c00-6878-11ea-80ec-b9d9b17b4d07.png">

PR #1436, opened 2020-03-26
feat: render a spinner when the Tree is loading

rgah2107:
#1422

- render a spinner when the Tree Node is loading


src/components/Tree/expandCollapseButton.js
```diff
@@ -11,16 +12,27 @@ function getIcon(isExpanded) {
...
 }
 
+function getButton(isExpanded, isLoading, onClick) {
+    const spinner = <BrandSpinner variant="brand" size="x-small" />;
+    return !isExpanded && isLoading ? (
+        <Button size="x-small" variant="base" icon={spinner} disabled />
+    ) : (
+        <Button size="x-small" icon={getIcon(isExpanded)} onClick={onClick} />
+    );
+}
```
maxxgreene: please use regular if condition, it's a better reading

maxxgreene: also we should put an spinner when is loading, not a button with spinner inside

src/components/Tree/expandCollapseButton.js
```diff
@@ -11,16 +12,27 @@ function getIcon(isExpanded) {
...
+
 export default function ExpandCollapseButton(props) {
-    const { hasChildren, isExpanded, onClick } = props;
+    const { hasChildren, isExpanded, isLoading, onClick } = props;
+
     if (hasChildren) {
-        return <Button size="x-small" icon={getIcon(isExpanded)} onClick={onClick} />;
+        return getButton(isExpanded, isLoading, onClick);
```
maxxgreene: if this is a function then it could be a component, I don't think we need this extra function ... all the logic could be here

```js
if (isLoading) {
    return <Spinner />;
}
if (hasChildren) {
    <Button size="x-small" icon={getIcon(isExpanded)} onClick={onClick} />;
}
return null;
```

src/components/Tree/expandCollapseButton.js
rgah2107: received, working on it

src/components/Tree/expandCollapseButton.js
rgah2107: received, working on it

rgah2107: solved according to the comments

src/components/Tree/expandCollapseButton.js
rgah2107: solved according to the comments

src/components/Tree/styled/spinner.js
```diff
@@ -0,0 +1,14 @@
+import styled from 'styled-components';
+import attachThemeAttrs from '../../../styles/helpers/attachThemeAttrs';
+import Spinner from '../../Spinner';
+
+const BrandSpinner = attachThemeAttrs(styled(Spinner))`
+    color: ${props => props.palette.brand.main};
```
TahimiLeonBravo, line 6: You don't need to pass a color for the spinner, it receives the brand color from the variant ;)

src/components/Tree/styled/spinner.js
```diff
@@ -0,0 +1,14 @@
...
+    margin-top: 18px;
+    margin-left: 8px;
```
TahimiLeonBravo, line 8: I think we should avoid using these margins here, instead, we can add a container with 
`position: relative;
 display: flex;
 justify-content: center;
 align-items: center;`

PR #1437, opened 2020-03-27
test: importRecordsFlow, stepOne, stepTwo and stepFour

blakmetall:
test: importRecordsFlow, stepOne, stepTwo and stepFour

@nexxtway/react-rainbow

### DIFF
--- a/src/components/ImportRecordsFlow/__test__/footer.spec.js
+++ b/src/components/ImportRecordsFlow/__test__/footer.spec.js
@@ -1,27 +1,29 @@
 import React from 'react';
 import { mount } from 'enzyme';
 import Footer from '../footer';
+import StyledBackButton from '../styled/backButton';
+import StyledNextButton from '../styled/nextButton';
 
 describe('<Footer />', () => {
     it('should have back button disabled when isBackButtonDisabled is set to true', () => {
         const component = mount(<Footer isBackButtonDisabled />);
-        const prevButton = component.find('button').first();
+        const prevButton = component.find(StyledBackButton).at(0);
         expect(prevButton.prop('disabled')).toBe(true);
     });
     it('should have next button disabled when isNextButtonDisabled is set to true', () => {
         const component = mount(<Footer isNextButtonDisabled />);
-        const nextButton = component.find('button').last();
+        const nextButton = component.find(StyledNextButton).at(0);
         expect(nextButton.prop('disabled')).toBe(true);
     });
     it('should have right label for all steps but the last', () => {
         const component = mount(<Footer isNextButtonDisabled currentStep="step-2" />);
-        const nextButton = component.find('button').last();
-        expect(nextButton.text()).toBe('Next');
+        const nextButton = component.find(StyledNextButton).at(0);
+        expect(nextButton.prop('label')).toBe('Next');
     });
     it('should have right label for last step button', () => {
         const lastStep = 'step-4';
         const component = mount(<Footer isNextButtonDisabled currentStep={lastStep} />);
-        const nextButton = component.find('button').last();
-        expect(nextButton.text()).toBe('Start Import');
+        const nextButton = component.find(StyledNextButton).at(0);
+        expect(nextButton.prop('label')).toBe('Start Import');
     });
 });
--- a/src/components/ImportRecordsFlow/__test__/footer.spec.js
+++ b/src/components/ImportRecordsFlow/__test__/footer.spec.js
@@ -1,0 +1,27 @@
+import React from 'react';
+import { mount } from 'enzyme';
+import Footer from '../footer';
+
+describe('<Footer />', () => {
+    it('should have back button disabled when isBackButtonDisabled is set to true', () => {
+        const component = mount(<Footer isBackButtonDisabled />);
+        const prevButton = component.find('button').first();
+        expect(prevButton.prop('disabled')).toBe(true);
+    });
+    it('should have next button disabled when isNextButtonDisabled is set to true', () => {
+        const component = mount(<Footer isNextButtonDisabled />);
+        const nextButton = component.find('button').last();
+        expect(nextButton.prop('disabled')).toBe(true);
+    });
+    it('should have right label for all steps but the last', () => {
+        const component = mount(<Footer isNextButtonDisabled currentStep="step-2" />);
+        const nextButton = component.find('button').last();
+        expect(nextButton.text()).toBe('Next');
+    });
+    it('should have right label for last step button', () => {
+        const lastStep = 'step-4';
+        const component = mount(<Footer isNextButtonDisabled currentStep={lastStep} />);
+        const nextButton = component.find('button').last();
+        expect(nextButton.text()).toBe('Start Import');
+    });
+});
--- a/src/components/ImportRecordsFlow/__test__/footer.spec.js
+++ b/src/components/ImportRecordsFlow/__test__/footer.spec.js
@@ -7,23 +7,23 @@ import StyledNextButton from '../styled/nextButton';
 describe('<Footer />', () => {
     it('should have back button disabled when isBackButtonDisabled is set to true', () => {
         const component = mount(<Footer isBackButtonDisabled />);
-        const prevButton = component.find(StyledBackButton).at(0);
+        const prevButton = component.find(StyledBackButton);
         expect(prevButton.prop('disabled')).toBe(true);
     });
     it('should have next button disabled when isNextButtonDisabled is set to true', () => {
         const component = mount(<Footer isNextButtonDisabled />);
-        const nextButton = component.find(StyledNextButton).at(0);
+        const nextButton = component.find(StyledNextButton);
         expect(nextButton.prop('disabled')).toBe(true);
     });
     it('should have right label for all steps but the last', () => {
         const component = mount(<Footer isNextButtonDisabled currentStep="step-2" />);
-        const nextButton = component.find(StyledNextButton).at(0);
+        const nextButton = component.find(StyledNextButton);
         expect(nextButton.prop('label')).toBe('Next');
     });
     it('should have right label for last step button', () => {
         const lastStep = 'step-4';
         const component = mount(<Footer isNextButtonDisabled currentStep={lastStep} />);
-        const nextButton = component.find(StyledNextButton).at(0);
+        const nextButton = component.find(StyledNextButton);
         expect(nextButton.prop('label')).toBe('Start Import');
     });
 });
--- a/src/components/ImportRecordsFlow/stepFour/__test__/stepFour.spec.js
+++ b/src/components/ImportRecordsFlow/stepFour/__test__/stepFour.spec.js
@@ -1,0 +1,30 @@
+import React from 'react';
+import { mount } from 'enzyme';
+import getPreviewDataToImport from '../../helpers/getPreviewDataToImport';
+import StepFour from '../';
+
+jest.mock('../../helpers/getPreviewDataToImport', () => jest.fn());
+
+const props = {
+    schemaFields: ['name', 'email'],
+    data: [{ Name: 'John', Email: 'john@gmail.com' }, { Name: 'Marie', Email: 'marie@gmail.com' }],
+    fieldsMap: {
+        name: 'Name',
+        email: 'Email',
+    },
+    attributes: {
+        name: { required: true },
+        email: {},
+    },
+};
+
+describe('<Preview />', () => {
+    it('should get preview data to import with the right values', () => {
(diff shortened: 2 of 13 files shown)


src/components/ImportRecordsFlow/__test__/footer.spec.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import { mount } from 'enzyme';
+import Footer from '../footer';
+
+describe('<Footer />', () => {
+    it('should have back button disabled when isBackButtonDisabled is set to true', () => {
+        const component = mount(<Footer isBackButtonDisabled />);
+        const prevButton = component.find('button').first();
```
LeandroTorresSicilia: Inside Footer, we are using the Button component and here what we need to check is that the disabled prop on the Button (StyledBackButton) component is the same value of isBackButtonDisabled because checking for the button tag element is introspection on the Button component. Remember that Footer does not know implementation details about Button.

src/components/ImportRecordsFlow/__test__/footer.spec.js
```diff
@@ -0,0 +1,27 @@
...
+    });
+    it('should have right label for all steps but the last', () => {
+        const component = mount(<Footer isNextButtonDisabled currentStep="step-2" />);
+        const nextButton = component.find('button').last();
+        expect(nextButton.text()).toBe('Next');
+    });
+    it('should have right label for last step button', () => {
+        const lastStep = 'step-4';
+        const component = mount(<Footer isNextButtonDisabled currentStep={lastStep} />);
+        const nextButton = component.find('button').last();
```
LeandroTorresSicilia: same for all test

src/components/ImportRecordsFlow/__test__/footer.spec.js
```diff
@@ -0,0 +1,27 @@
...
+        expect(prevButton.prop('disabled')).toBe(true);
+    });
+    it('should have next button disabled when isNextButtonDisabled is set to true', () => {
+        const component = mount(<Footer isNextButtonDisabled />);
+        const nextButton = component.find('button').last();
+        expect(nextButton.prop('disabled')).toBe(true);
+    });
+    it('should have right label for all steps but the last', () => {
+        const component = mount(<Footer isNextButtonDisabled currentStep="step-2" />);
+        const nextButton = component.find('button').last();
```
LeandroTorresSicilia: same ^^ only check the label prop in the Button component

src/components/ImportRecordsFlow/stepFour/__test__/stepFour.spec.js
```diff
@@ -0,0 +1,30 @@
...
+        email: 'Email',
+    },
+    attributes: {
+        name: { required: true },
+        email: {},
+    },
+};
+
+describe('<Preview />', () => {
+    it('should get preview data to import with the right values', () => {
```
LeandroTorresSicilia: should call ...

src/components/ImportRecordsFlow/stepFour/__test__/stepFour.spec.js
```diff
@@ -0,0 +1,30 @@
...
+        mount(<StepFour {...props} />);
+        expect(getPreviewDataToImport).toHaveBeenCalledWith(
+            props.data.slice(0, 5),
```
LeandroTorresSicilia: I think it is more deterministic if we put the values directly instead of make the same code implementation

src/components/ImportRecordsFlow/stepFour/__test__/stepFour.spec.js
```diff
@@ -0,0 +1,30 @@
...
+            props.fieldsMap,
+            props.attributes,
+        );
+    });
+});
```
LeandroTorresSicilia: We should add another test:
`should render the right amount of columns`

src/components/ImportRecordsFlow/stepTwo/__test__/preview.spec.js
```diff
@@ -0,0 +1,48 @@
...
+    it('should show default labels', () => {
+        const component = mount(<Preview />);
+        const title = component.find(StyledFileCardTitle);
+        const type = component.find(StyledFileCardDescription);
+        expect(title.text()).toBe('Unknow File Name');
+        expect(type.text()).toBe('Unknow File Type');
+    });
+    it('should set the right labels according to params', () => {
+        const params = { fileName: 'preview-title', fileType: 'text/csv' };
+        const component = mount(<Preview {...params} />);
```
LeandroTorresSicilia: I think it is more explicit if we use in this way:
```const component = mount(<Preview fileName="preview-title" fileType="text/csv" />);```

File: src/components/ImportRecordsFlow/stepTwo/__test__/preview.spec.js [hunk 8]
```diff
@@ -0,0 +1,48 @@
...
+            columns: ['First_Name', 'Last_Name'],
+            isLoading: true,
+        };
+        const component = mount(<Preview {...props} />);
+        const expectedPassedData = {
+            data: component.find('PreviewTable').prop('data'),
+            columns: component.find('PreviewTable').prop('columns'),
+            isLoading: component.find('PreviewTable').prop('isLoading'),
+        };
+        expect(expectedPassedData).toEqual(props);
```
LeandroTorresSicilia: It is more readable doing:
```
expect(component.find('PreviewTable').props()).toEqual(props)
```

src/components/ImportRecordsFlow/stepTwo/__test__/preview.spec.js
```diff
@@ -0,0 +1,48 @@
...
+    });
+});
```
LeandroTorresSicilia: I think we should put PreviewTable in other file and test it:
- should render a spinner when isLoading is passed
- should render a table when isLoading is not passed
- should render the right amount of columns

src/components/ImportRecordsFlow/stepTwo/__test__/previewTable.spec.js
```diff
@@ -0,0 +1,22 @@
...
+import PreviewTable from '../previewTable';
+
+const data = [
+    { First_Name: 'John', Last_Name: 'Doe' },
+    { First_Name: 'Marie', Last_Name: 'Sun' },
+    { First_Name: 'Jane', Last_Name: 'Doe' },
+];
+const columns = ['First_Name', 'Last_Name'];
+
+describe('<Preview />', () => {
```
LeandroTorresSicilia: Oh, I see, great. Only need to add the test for when isLoading is false and put the PreviewTable in another file, also change the describe name to `<PreviewTable />`

src/components/ImportRecordsFlow/stepTwo/__test__/stepTwo.spec.js
```diff
@@ -0,0 +1,24 @@
...
+    it('should show default upload presentation on load', () => {
+        const component = mount(<StepTwo />);
+        const renderIfBlock = component.find('RenderIf').at(1);
+        expect(renderIfBlock.prop('isTrue')).toBe(true);
+    });
+    it('should hide the draggable upload presentation on load', () => {
+        const component = mount(<StepTwo />);
+        const renderIfBlock = component.find('RenderIf').at(0);
+        expect(renderIfBlock.prop('isTrue')).toBe(false);
+    });
```
LeandroTorresSicilia: Let's try to simulate dragOver and check to render if conditions are inverted.
Same for when dragLeave and also I think we could simulate the drop event and test both cases when has the correct file and when not

src/components/ImportRecordsFlow/stepTwo/__test__/uploadFileButton.spec.js
```diff
@@ -0,0 +1,22 @@
...
+import { uniqueId } from '../../../../libs/utils';
+
+jest.mock('../../../../libs/utils/uniqueId', () => jest.fn(() => '101'));
+
+describe('<UploadFileButton />', () => {
+    it('should have unique id set properly', () => {
+        const component = mount(<UploadFileButton onChange={() => {}} />);
+        const input = component.find('input');
+        expect(uniqueId).toHaveBeenCalledWith('upload-file-input');
+        expect(input.prop('id')).toBe('101');
```
LeandroTorresSicilia: I think these should be two tests:
- should call uniqueId with the right value
- should associate the input and label with the same id
in order to check accessibility to this last test case, we need to add another expect that check the value of the htmlFor attribute in the label element

src/components/ImportRecordsFlow/stepTwo/__test__/uploadFileButton.spec.js
```diff
@@ -0,0 +1,22 @@
...
+    });
+    it('should run change event when a file is added to the input', () => {
+        const onChangeFn = jest.fn(() => {});
+        const component = mount(<UploadFileButton onChange={onChangeFn} />);
+        const input = component.find('input');
+        input.simulate('change', { target: { files: ['dummy-value'] } });
+        expect(onChangeFn).toHaveBeenCalled();
+    });
+});
```
LeandroTorresSicilia: We should add test cases to check the input has these props:
```
type="file"
accept="text/csv"
multiple={false}
```

src/components/ImportRecordsFlow/stepTwo/previewTable.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Column from '../../Column';
+import Spinner from '../../Spinner';
+import StyledSpinnerContainer from './styled/spinnerContainer';
+import StyledTable from '../styled/table';
+
+export default function PreviewTable(props) {
```
LeandroTorresSicilia: Good job

src/components/ImportRecordsFlow/__test__/footer.spec.js
```diff
@@ -0,0 +1,29 @@
...
+    });
+    it('should have right label for all steps but the last', () => {
+        const component = mount(<Footer isNextButtonDisabled currentStep="step-2" />);
+        const nextButton = component.find(StyledNextButton).at(0);
+        expect(nextButton.prop('label')).toBe('Next');
+    });
+    it('should have right label for last step button', () => {
+        const lastStep = 'step-4';
+        const component = mount(<Footer isNextButtonDisabled currentStep={lastStep} />);
+        const nextButton = component.find(StyledNextButton).at(0);
```
LeandroTorresSicilia, line 26: In all these tests there is no need to use `at` method since StyledBackButton and StyledNextButton are used only once, instead use:
```const nextButton = component.find(StyledNextButton);```

File: src/components/ImportRecordsFlow/stepTwo/__test__/uploadFileButton.spec.js [hunk 16]
```diff
@@ -0,0 +1,38 @@
...
+
+jest.mock('../../../../libs/utils/uniqueId', () => jest.fn(() => 'uniqueId-value'));
+
+describe('<UploadFileButton />', () => {
+    it('should call uniqueId with the right value', () => {
+        mount(<UploadFileButton onChange={() => {}} />);
+        expect(uniqueId).toHaveBeenCalledWith('upload-file-input');
+    });
+    it('should associate the input and label with the same id', () => {
+        expect.assertions(2);
```
LeandroTorresSicilia, line 16: expect.assertions is only needed for async tests

src/components/ImportRecordsFlow/stepTwo/__test__/uploadFileButton.spec.js
```diff
@@ -0,0 +1,38 @@
...
+        const component = mount(<UploadFileButton onChange={() => {}} />);
+        const uploadWrapper = component.find(StyledUploadFileLabel).at(0);
```
LeandroTorresSicilia, line 18: same ^^ no need to use at

src/components/ImportRecordsFlow/stepTwo/__test__/uploadFileButton.spec.js
```diff
@@ -0,0 +1,38 @@
...
+    });
+    it('should run change event when a file is added to the input', () => {
+        const onChangeFn = jest.fn(() => {});
+        const component = mount(<UploadFileButton onChange={onChangeFn} />);
+        const input = component.find('input');
+        input.simulate('change', { target: { files: ['dummy-value'] } });
+        expect(onChangeFn).toHaveBeenCalled();
+    });
+    it('should have the right props for the upload input', () => {
+        expect.assertions(3);
```
LeandroTorresSicilia, line 31: same ^^

Issue #1446, opened 2020-04-01
test: add tests to the ButtonGroupPicker and ButtonOption components


TahimiLeonBravo:
#### Requirements
- [ ] Add unit test
- [ ] Add integrations test

##### buttonGroupPicker
- [x] should render a label when label prop is passed
- [x] should render bottom help text when bottomHelpText prop is passed
- [x] should render required asterisk when required prop is true
- [x] should pass type checkbox in context when multiple is true
- [x] should pass type radio in context when multiple is false
- [x] should render an error text when error prop is passed
- [x] should set the error message id as id for the error text
- [x] should pass the error message id to context
- [x] should fire onchange callback when selection changes
- [x] should fire onchange with the right value when multiple is false and option is checked
- [x] should fire onchange with an array of values including the value checked when multiple is true
- [x] should fire onchange with an array containing the value checked when multiple is true and nothing is selected
- [x] should fire onchange with an array excluding the unchecked value when multiple is true and option is unchecked
##### buttonOption
- [x] should set disabled prop of input to true when disabled prop is true
- [x] should set the label passed as label for the input
- [x] should fire onclick callback when clicked 
- [x] should pass a generated id to the input element and set the same id to the htmFor of the label element
- [x] should set the name passed through context to the name prop of the input element
- [x] should set the prop name passed to the value prop of the input element
- [x] should set the ariaDescribedBy passed through context to the aria-describedby prop of the input element
- [x] should set the type passed through context to the type prop of the input element
- [x] should set the input selected prop equal to the value returned by function isOptionSelected
##### isOptionSelected
- [x] should return true when name is present in the array of values
- [x] should return false when name is not present in the array of values
- [x] should return true when values is a string and it is equal to name
- [x] should return false when values is a string and it is different to name
- [x] should return false when the value passed is other type than array and string

##### buttonGroupPicker-1
- [x] should focus the option clicked
- [x] should check the option clicked
- [x] should focus the next option when press arrow down key
- [x] should check the next option when press arrow down key
- [x] should uncheck the current option when press arrow down key
- [x] should check the previous option when press arrow left key
- [x] should check the next option when press arrow right key
- [x] should uncheck the current option when press arrow left key
- [x] should check the first option when last option is checked and press arrow right key
##### buttonGroupPicker-3
- [x] should not check the next option when multiple is true and press arrow down key
- [x] should uncheck the current option when multiple is true and click a checked option
- [x] should allow several options checked at the same time
[...]

PR #1457, opened 2020-04-03
test: add tests to the ButtonGroupPicker and ButtonOption components

HellWolf93:
fix: #1446

- add 'id' prop to ButtonGroupPicker and some refactor
- move onclick callback to label
- add tests to the ButtonGroupPicker and ButtonOption components
___


integration/specs/ButtonGroupPicker/buttonGroupPicker-3.spec.js
```diff
@@ -0,0 +1,40 @@
...
+        expect(option.isChecked()).toBe(false);
+    });
+
+    it('should allow several options checked at the same time', () => {
+        const groupPicker = new PageButtonGroupPicker(BUTTON_GROUP_PICKER);
+        const option1 = groupPicker.getItem(0);
+        const option2 = groupPicker.getItem(1);
+        const option3 = groupPicker.getItem(2);
+        option3.click();
+        expect(option1.isChecked() && option2.isChecked() && option3.isChecked()).toBe(true);
```
LeandroTorresSicilia, line 38: to be more consistent with what we do instead use:
```
expect(option1.isChecked()).toBe(true);
expect(option2.isChecked()).toBe(true);
expect(option3.isChecked()).toBe(true);
```

src/components/ButtonGroupPicker/__test__/buttonGroupPicker.spec.js
```diff
@@ -0,0 +1,129 @@
...
+    });
+
+    it('should render bottom help text when bottomHelpText prop is passed', () => {
+        const component = mount(<ButtonGroupPicker bottomHelpText="My help text" />);
+        expect(component.find('div[children="My help text"]').length).toBe(1);
+    });
+
+    it('should render required asterisk when required prop is true', () => {
+        const component = mount(<ButtonGroupPicker label="My label" required />);
+        expect(component.find('abbr[title="required"]').length).toBe(1);
```
LeandroTorresSicilia, line 19: since abbr is an implementation detail of the RequiredAsterisk component we should treat it as a black box, the we should do:
```expect(component.find('RequiredAsterisk').exists()).toBe(true);```

File: src/components/ButtonGroupPicker/__test__/buttonGroupPicker.spec.js [hunk 3]
```diff
@@ -0,0 +1,129 @@
...
+    });
+
+    it('should pass type checkbox in context when multiple is true', () => {
+        const component = shallow(<ButtonGroupPicker multiple />);
+        const context = component.instance().getContext();
```
LeandroTorresSicilia, line 24: instead of checking for internal getContext, we should find the provider and check for the value prop has the right type

src/components/ButtonGroupPicker/__test__/buttonGroupPicker.spec.js
```diff
@@ -0,0 +1,129 @@
...
+        expect(context.type).toBe('checkbox');
+    });
+
+    it('should pass type radio in context when multiple is false', () => {
+        const component = shallow(<ButtonGroupPicker />);
+        const context = component.instance().getContext();
```
LeandroTorresSicilia, line 30: same ^^

src/components/ButtonGroupPicker/__test__/buttonGroupPicker.spec.js
```diff
@@ -0,0 +1,129 @@
...
+        expect(context.type).toBe('radio');
+    });
+
+    it('should render an error text when error prop is passed', () => {
+        const component = mount(<ButtonGroupPicker error="error" />);
+        expect(component.find('div[children="error"]').length).toBe(1);
```
LeandroTorresSicilia, line 36: same ^^ you can simply check that the StyledErrorText exists instead of knowing implementation details like the div:
```
import { StyledErrorText } from '../styled';
...
expect(component.find(StyledErrorText).exists()).toBe(true);
```

src/components/ButtonGroupPicker/__test__/buttonGroupPicker.spec.js
```diff
@@ -0,0 +1,129 @@
...
+    });
+
+    it('should set the error message id as id for the error text', () => {
+        const component = mount(<ButtonGroupPicker error="error" />);
+        const errorMessageId = component.instance().errorMessageId;
```
LeandroTorresSicilia, line 41: same ^^ let's try to avoid treatment with instance variables (implementation details that could change for example if we migrate this component to functional) and instead mock `uniqueId` function and check that the prop id is passed to StyledErrorText is what the mock returned

src/components/ButtonGroupPicker/__test__/buttonGroupPicker.spec.js
```diff
@@ -0,0 +1,129 @@
...
+        const errorElement = component.find('div[children="error"]').first();
```
LeandroTorresSicilia, line 42: same ^^ find StyledErrorText

src/components/ButtonGroupPicker/__test__/buttonGroupPicker.spec.js
```diff
@@ -0,0 +1,129 @@
...
+                <ButtonOption label="Option 1" name="option1" />
+                <ButtonOption label="Option 2" name="option2" />
+            </ButtonGroupPicker>,
+        );
+        component
+            .find('input')
+            .at(1)
+            .simulate('change', { target: { checked: true } });
+        expect(onChangeFn).toHaveBeenCalledTimes(1);
+    });
```
LeandroTorresSicilia, line 66: I think we should remove this test since the below test when check with what is being called also is check that actually was called

src/components/ButtonGroupPicker/index.js
```diff
@@ -88,7 +89,7 @@ class ButtonGroupPicker extends Component {
                     <StyledHelpText>{bottomHelpText}</StyledHelpText>
                 </RenderIf>
                 <RenderIf isTrue={!!error}>
-                    <StyledErrorText id={errorMessageId}>{error}</StyledErrorText>
+                    <StyledErrorText id={this.errorMessageId}>{error}</StyledErrorText>
```
LeandroTorresSicilia, line 92: here we need to use ```this.getErrorMessageId()``` since when error prop is not passed then this id should be undefined

src/components/ButtonOption/__test__/buttonOption.spec.js
```diff
@@ -0,0 +1,100 @@
...
+
+    it('should fire onclick callback when clicked', () => {
+        const onClickFn = jest.fn();
+        const component = mount(
+            <ButtonGroupPicker>
+                <ButtonOption label="Option 1" onClick={onClickFn} />
+            </ButtonGroupPicker>,
+        );
+        component.find('label').simulate('click');
+        expect(onClickFn).toHaveBeenCalledTimes(1);
```
LeandroTorresSicilia, line 35: we should change this test to check with what was called

src/components/ButtonOption/__test__/buttonOption.spec.js
```diff
@@ -0,0 +1,100 @@
...
+
+    it('should pass a generated id to the input element and set the same id to the htmFor of the label element', () => {
+        const component = mount(
+            <ButtonGroupPicker>
+                <ButtonOption label="Option 1" />
+            </ButtonGroupPicker>,
+        );
+        const htmlFor = component.find('label').prop('htmlFor');
+        const inputId = component.find('input').prop('id');
+        expect(htmlFor).toBe(inputId);
```
LeandroTorresSicilia, line 46: good job, clever way of check this

src/components/ButtonOption/__test__/buttonOption.spec.js
```diff
@@ -0,0 +1,100 @@
...
+    });
+
+    it('should set the name passed through context to the name prop of the input element', () => {
+        const component = mount(
+            <ButtonGroupPicker>
+                <ButtonOption label="Option 1" />
+            </ButtonGroupPicker>,
+        );
+        const contextName = component.instance().getContext().name;
```
LeandroTorresSicilia, line 55: same ^^ let's avoid the use of component.instance() and instead get this value from value prop in Provider component

src/components/ButtonOption/__test__/buttonOption.spec.js
```diff
@@ -0,0 +1,100 @@
...
+        expect(value).toBe('option1');
+    });
+
+    it('should set the ariaDescribedBy passed through context to the aria-describedby prop of the input element', () => {
+        const component = mount(
+            <ButtonGroupPicker>
+                <ButtonOption label="Option 1" />
+            </ButtonGroupPicker>,
+        );
+        const contextAriaDescribedBy = component.instance().getContext().ariaDescribedBy;
```
LeandroTorresSicilia, line 76: same ^^

src/components/ButtonOption/__test__/buttonOption.spec.js
```diff
@@ -0,0 +1,100 @@
...
+        expect(contextAriaDescribedBy).toBe(inputAriaDescribedBy);
+    });
+
+    it('should set the type passed through context to the type prop of the input element', () => {
+        const component = mount(
+            <ButtonGroupPicker>
+                <ButtonOption label="Option 1" />
+            </ButtonGroupPicker>,
+        );
+        const contextType = component.instance().getContext().type;
```
LeandroTorresSicilia, line 87: same ^^

src/components/ButtonOption/__test__/buttonOption.spec.js
```diff
@@ -0,0 +1,100 @@
...
+        const inputType = component.find('input').prop('type');
+        expect(contextType).toBe(inputType);
+    });
+
+    it('should set the input selected prop equal to the value returned by function isOptionSelected', () => {
```
LeandroTorresSicilia, line 92: let's simply call this:
```should pass the right checked value to the input```
note that we need to not tight couple the name of the test with implementations details like the isOptionSelected function, because if we change this we also need to maintain the test name

File: src/components/ButtonOption/pageObject/index.js [hunk 16]
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    /**
+     * Returns true when the radiobutton has the focus.
+     * @method
+     * @returns {bool}
+     */
+    hasFocus() {
+        return $(this.rootElement)
+            .$('input[type="radio"]')
```
LeandroTorresSicilia, line 29: this should not be restricted to radio since checkbox type also can be focused

src/components/ButtonOption/pageObject/index.js
```diff
@@ -0,0 +1,45 @@
...
+
+    /** Click the radiobutton.
+     * @method
+     */
+    click() {
+        $(this.rootElement).click();
+    }
+
+    /**
+     * Returns true when the radiobutton has the focus.
```
LeandroTorresSicilia, line 23: readiobutton  => ButtonOption

src/components/ButtonOption/pageObject/index.js
```diff
@@ -0,0 +1,45 @@
...
+    /**
+     * Create a new ButtonOption page object.
+     * @constructor
+     * @param {string} rootElement - The selector of the Radio root element.
+     */
+    constructor(rootElement) {
+        this.rootElement = rootElement;
+    }
+
+    /** Click the radiobutton.
```
LeandroTorresSicilia, line 15: readiobutton => ButtonOption

src/components/ButtonOption/pageObject/index.js
```diff
@@ -0,0 +1,45 @@
...
+     * @returns {bool}
+     */
+    hasFocus() {
+        return $(this.rootElement)
+            .$('input[type="radio"]')
+            .isFocused();
+    }
+
+    /**
+     * Returns true when the radio is checked.
```
LeandroTorresSicilia, line 34: radio => ButtonOption

src/components/ButtonGroupPicker/__test__/buttonGroupPicker.spec.js
```diff
@@ -16,31 +20,30 @@ describe('<ButtonGroupPicker />', () => {
...
-        expect(component.find('div[children="error"]').length).toBe(1);
+        expect(component.find(StyledErrorText).exists()).toBe(true);
     });
 
     it('should set the error message id as id for the error text', () => {
         const component = mount(<ButtonGroupPicker error="error" />);
-        const errorMessageId = component.instance().errorMessageId;
-        const errorElement = component.find('div[children="error"]').first();
-        expect(errorElement.prop('id')).toBe(errorMessageId);
+        const errorElement = component.find(StyledErrorText).first();
```
LeandroTorresSicilia, line 45: since StyledErrorText is used only once there is no need to use first

Issue #1429, opened 2020-03-19
feat: implement the minus icon when some children are selected on the Tree component


TahimiLeonBravo:
### Requirements
- Implement a new example with a real use-case

### Specs
- Add the `indeterminate` state
- When the state of the checkmark is indeterminate a minus icon is rendered inside the checkbox

#### Review the example 4 on the design
https://zpl.io/blE8m9D

<img width="808" alt="Screen Shot 2020-04-03 at 11 26 28 AM" src="https://user-images.githubusercontent.com/32083556/78388368-196bd000-759e-11ea-8715-af5006bc3b07.png">

LeandroTorresSicilia:
Steps:
1. Move table checkbox to src/components/PrimitiveCheckbox in order to reuse in both Table and Tree
2. Change prop checked in PrimitiveCheckbox to receive true | false | 'indeterminate'
3. Change Tree checked value in data shape to receive true | false | 'indeterminate'
4. When the checked value is undefined should not render the Checkbox in Tree component

PR #1459, opened 2020-04-04
feat: implement the minus icon when some children are selected on the…

rgah2107:
fix: #1429

- implement the minus icon when some children are selected on the Tree component

@nexxtway/react-rainbow


src/components/PrimitiveCheckbox/index.js
```diff
@@ -1,27 +1,27 @@
...
 
     componentDidUpdate(prevProps) {
-        const { indeterminate: prevIndeterminate } = prevProps;
-        const { indeterminate } = this.props;
-        if (prevIndeterminate !== indeterminate) {
-            this.inputRef.current.indeterminate = indeterminate;
+        const { checked: prevChecked } = prevProps;
+        const { checked } = this.props;
+        if (prevChecked !== checked) {
+            this.inputRef.current.checked = checked;
```
LeandroTorresSicilia, line 24: We change how we use the component but we can't change the native approach to make this. I think we should maintain the indeterminate here.
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#Indeterminate_state_checkboxes

src/components/PrimitiveCheckbox/styled/innerContainer.js
```diff
@@ -60,21 +79,6 @@ const StyledInnerContainer = attachThemeAttrs(styled.div)`
...
-        display: block;
-        position: absolute;
-        top: 50%;
-        left: 50%;
-        width: 0.5rem;
-        height: 2px;
-        background-color: ${props => props.palette.brand.main};
-        border: 0;
-        transform: translate3d(-50%, -50%, 0);
-    }
```
LeandroTorresSicilia, line 76: I think using the native indeterminate you don't need to make any change in this file

src/components/Table/__test__/table.spec.js
```diff
@@ -411,23 +411,23 @@ describe('<Table />', () => {
...
         const component = mount(
             <Table data={tableData} selectedRows={['1234qwerty']} showCheckboxColumn keyField="id">
                 <Column field="name" header="Name" />
             </Table>,
         );
-        expect(component.state().bulkSelection).toBe('some');
+        expect(component.state().bulkSelection).toBe('indeterminate');
     });
-    it('should set the bulkSelection initially to "all" when all rows are selected', () => {
+    it('should set the bulkSelection initially to true when all rows are selected', () => {
```
LeandroTorresSicilia, line 430: I think we should not change this states

src/components/Tree/readme.md
```diff
@@ -94,27 +94,27 @@
...
-        { label: 'Tree Item', icon: <FileIcon /> },
+        { label: 'Tree Item',checked: false, icon: <FileIcon /> },
+        { label: 'Tree Item',checked: true, icon: <FileIcon /> },
+        { label: 'Tree Item',checked: false, icon: <FileIcon /> },
         {
             label: 'Tree Branch',
             icon: <FolderCloseIcon />,
             isExpanded: true,
-            isChecked: false,
+            checked: 'indeterminate',
```
LeandroTorresSicilia, line 104: here keep using isChecked

src/components/Table/head/header.js
```diff
@@ -149,7 +149,7 @@ Header.propTypes = {
     onDeselectAllRows: PropTypes.func,
     tableId: PropTypes.string.isRequired,
     maxRowSelection: PropTypes.number,
-    bulkSelection: PropTypes.oneOf(['none', 'some', 'all']),
+    bulkSelection: PropTypes.oneOf([true, false, 'indeterminate']),
```
LeandroTorresSicilia, line 152: In order to not make too many changes, we can use a map instead of change all this values

src/components/PrimitiveCheckbox/index.js
```diff
@@ -1,27 +1,27 @@
...
+        this.inputRef.current.indeterminate = checked === 'indeterminate';
     }
 
     componentDidUpdate(prevProps) {
         const { indeterminate: prevIndeterminate } = prevProps;
-        const { indeterminate } = this.props;
-        if (prevIndeterminate !== indeterminate) {
-            this.inputRef.current.indeterminate = indeterminate;
+        const { checked } = this.props;
+        if (prevIndeterminate !== (checked === 'indeterminate')) {
```
LeandroTorresSicilia, line 23: Since now there is not indeterminate prop now you need to remove:
```const { indeterminate: prevIndeterminate } = prevProps;```
This code es meant to update the `this.inputRef.current.indeterminate` when checked is changed dynamically:
1. when checked = 'indeterminate' change to true or false
2. when checked = true | false change to 'indeterminate'
This is needed for the component to react to a prop change

File: src/components/Table/head/selectableHeader.js [hunk 7]
```diff
@@ -18,8 +18,8 @@ export default function SelectableHeader(props) {
     const name = `${tableId}-options`;
     const isDisabled = maxRowSelection === 0;
     const isRadio = maxRowSelection === 1;
-    const isIndeterminate = bulkSelection === 'some';
-    const isAllSelected = bulkSelection === 'all';
+    const bulkStateMap = new Map([['all', true], ['some', 'indeterminate'], ['none', false]]);
+    const checked = bulkStateMap.get(bulkSelection);
```
LeandroTorresSicilia, line 22: good job
only put bulkStateMap outside the component since it has fixed values that do not depend on any prop or state because here the map is recreating in each re-render without need it.
Also, you can simply use an object:
const bulkStateMap = {
    all: true,
    some: 'indeterminate',
    none: false,
};

src/components/Table/head/__test__/selectableHeader.spec.js
```diff
@@ -3,54 +3,36 @@ import { mount } from 'enzyme';
...
-        const component = mount(<SelectableHeader bulkSelection="some" />);
-        expect(component.find('InputCheckbox').prop('indeterminate')).toBe(true);
-    });
-    it('should set indeterminate to false in InputCheckbox component when bulkSelection is other than "some"', () => {
-        const values = ['none', 'all', '', 123, undefined];
-        values.forEach(value => {
-            const component = mount(<SelectableHeader bulkSelection={value} />);
-            expect(component.find('InputCheckbox').prop('indeterminate')).toBe(false);
-        });
-    });
```
LeandroTorresSicilia, line 44: This three tests case needs to be converted:
- should set checked to false in InputCheckbox component when bulkSelection is "none"
- should set checked to "indeterminate" in InputCheckbox component when bulkSelection is "some"
- should set checked to true in InputCheckbox component when bulkSelection is "all"

src/components/Tree/child.js
```diff
@@ -23,7 +24,7 @@ export default function Child(props) {
         onSelect,
     } = props;
     const hasChildren = Array.isArray(children);
-    const hasCheckbox = typeof isChecked === 'boolean';
+    const hasCheckbox = isChecked !== undefined;
```
LeandroTorresSicilia, line 27: I think we should do:
```const hasCheckbox = typeof isChecked === 'boolean' || isChecked === 'indeterminate'```
in order to avoid render the checkbox when pass a wrong value that is different to `undefined`

File: src/components/Tree/child.js [hunk 10]
```diff
@@ -62,7 +65,7 @@ export default function Child(props) {
 
 Child.propTypes = {
     label: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
-    isChecked: PropTypes.bool,
+    isChecked: PropTypes.oneOf([true, false, 'indeterminate']),
```
LeandroTorresSicilia, line 68: We need to update this in index.js in order to get the right documentation in the props table and also in index.d.ts for properly typing when using typescript:
```isChecked?: boolean | 'indeterminate';```

File: src/components/Tree/readme.md [hunk 11]
```diff
@@ -94,17 +94,17 @@
...
         {
             label: 'Tree Branch',
             icon: <FolderCloseIcon />,
             isExpanded: true,
-            isChecked: false,
+            isChecked: 'indeterminate',
             children: [
                 { label: 'Tree Item', isChecked: false },
-                { label: 'Tree Item', isChecked: false },
+                { label: 'Tree Item', isChecked: true },
```
LeandroTorresSicilia, line 107: I think we should make this example works, e.g:
- select all when checking the parent checkbox
- deselect all when unchecking the parent checkbox
- set to indeterminate the parent checkbox when only some and not all child checkbox are selected

src/components/Tree/styled/inputContainer.js
```diff
@@ -0,0 +1,12 @@
+import styled from 'styled-components';
+
+const InputContainer = styled.div`
+    display: flex;
+    position: relative;
+    align-items: center;
+    justify-content: center;
+    width: 20px;
+    margin-right: 8px;
```
LeandroTorresSicilia, line 9: I think there is no need to use a wrapper, instead, you can make the PrimitiveCheckbox a styled and pass your own styles to it since it already has a wrapper which receives a style prop.
Also, the only style you need to pass is the `margin-right: 8px;`

src/components/Tree/styled/inputCheckbox.js
```diff
@@ -0,0 +1,9 @@
+import styled from 'styled-components';
+import attachThemeAttrs from '../../../styles/helpers/attachThemeAttrs';
+import PrimitiveCheckbox from '../../PrimitiveCheckbox';
+
+const InputCheckbox = attachThemeAttrs(styled(PrimitiveCheckbox))`
```
LeandroTorresSicilia, line 5: when not use any value from theme there is no need to use attachThemeAttrs

src/components/PrimitiveCheckbox/index.js
```diff
@@ -65,6 +65,7 @@ export default class InputCheckbox extends Component {
             id,
             name,
             checked,
+            indeterminate,
```
LeandroTorresSicilia, line 68: remember that now there is no  indeterminate prop

src/components/PrimitiveCheckbox/index.js
```diff
@@ -83,6 +84,7 @@ export default class InputCheckbox extends Component {
                         onKeyDown={onKeyDown}
                         disabled={disabled}
                         checked={checked}
+                        indeterminate={indeterminate}
```
LeandroTorresSicilia, line 87: remove here

src/components/PrimitiveCheckbox/index.js
```diff
@@ -104,14 +106,14 @@ InputCheckbox.propTypes = {
     onFocus: PropTypes.func,
     onBlur: PropTypes.func,
     onKeyDown: PropTypes.func,
-    checked: PropTypes.bool,
+    checked: PropTypes.oneOf([true, false, 'indeterminate']),
     indeterminate: PropTypes.bool,
```
LeandroTorresSicilia, line 110: same ^^ remove here

PR #1462, opened 2020-04-06
feat: new TimeRangePicker component

rafaelmg83:
fix: #1443 

- New Component TimeRangePicker

@nexxtway/react-rainbow


src/components/TimePicker/index.js
```diff
@@ -136,7 +137,7 @@ class TimePicker extends Component {
         return (
             <StyledContainer id={id} className={className} style={style}>
                 <Input
-                    id="time-picker_time-input"
+                    id={uniqueId('time-picker_time-input')}
```
reiniergs, line 140: unique Ids are being use wrong here because every time the component renders it will have diff id, they should unique by instance ... I think it should be used as state or using memo

reiniergs: this approach is being used everywhere in the implementation

src/components/TimeRangePicker/index.js
```diff
@@ -0,0 +1,65 @@
...
+import RenderIf from '../RenderIf';
+import Label from '../Input/label';
+import { uniqueId } from '../../libs/utils';
+
+/**
+ * TimeRangePicker is a graphical user interface widget that
+ * allows the user to select a time range.
+ * @category Form
+ */
+export default function TimeRangePicker(props) {
```
reiniergs, line 14: the component is Inputable, right? I think it needs more work

Issue #1444, opened 2020-03-31
feat: implement WeeklyDayPicker component


TahimiLeonBravo:
#### Requirements
- [x] Define API
- [x] Implement component
- [x] Testing

#####  API
https://docs.google.com/document/d/1Jqih4ixwaS-dIOHWJKT4EYtBNKs3wRm-wUfzgJhhK6M/edit?usp=sharing

##### Design
https://zpl.io/aXvK588

PR #1476, opened 2020-04-11
feat: implement WeekDayPicker component

blakmetall:
#1444 

- Implement WeekDayPicker component

@nexxtway/react-rainbow


src/components/WeekDayPicker/index.d.ts
```diff
@@ -0,0 +1,19 @@
+import { ReactNode, ChangeEvent } from 'react';
+import { BaseProps } from '../types';
+
+export interface WeekDayPickerProps extends BaseProps {
+    id?: string;
+    name?: string;
+    value?: string | Array<string>;
+    label?: string | ReactNode;
```
LeandroTorresSicilia, line 8: here use only the ReactNode since it already contains the string type

src/components/WeekDayPicker/index.d.ts
```diff
@@ -0,0 +1,19 @@
...
+    hideLabel?: boolean;
+    disabled?: boolean;
+    required?: boolean;
+    readOnly?: boolean;
+    multiple?: boolean;
+    error?: string | ReactNode;
```
LeandroTorresSicilia, line 14: same ^^

src/components/WeekDayPicker/index.d.ts
```diff
@@ -0,0 +1,19 @@
...
+    onChange?: (event: ChangeEvent<HTMLElement>) => void;
```
LeandroTorresSicilia, line 15: here what we send in not a native event, instead:
```
type Value = string | string[];

export interface ButtonGroupPickerProps extends BaseProps {
    ...
    onChange?: (value: Value) => void;
    ...
}
```

src/components/WeekDayPicker/index.js
```diff
@@ -0,0 +1,86 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { useUniqueIdentifier } from '../../libs/hooks';
+import RequiredAsterisk from '../RequiredAsterisk';
+import RenderIf from '../RenderIf';
+import WeekDayItems from './weekDayItems';
+import { StyledFieldset, StyledLabel, StyledTextError } from './styled';
+
+const WeekDayPicker = props => {
```
LeandroTorresSicilia, line 9: For convention let's use function declaration, only exceptional case like using React.forwardRef
Also, fix this in  helper functions

src/components/WeekDayPicker/index.js
```diff
@@ -0,0 +1,86 @@
...
+    label: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
+    hideLabel: PropTypes.bool,
+    disabled: PropTypes.bool,
+    required: PropTypes.bool,
+    readOnly: PropTypes.bool,
+    multiple: PropTypes.bool,
+    error: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
+    onChange: PropTypes.func,
+    className: PropTypes.string,
+    style: PropTypes.object,
```
LeandroTorresSicilia, line 67: we need to add descriptions here in index.js, since the table props documentation is generated from descriptions declared here

src/components/WeekDayPicker/index.d.ts
```diff
@@ -0,0 +1,19 @@
+import { ReactNode, ChangeEvent } from 'react';
+import { BaseProps } from '../types';
+
+export interface WeekDayPickerProps extends BaseProps {
+    id?: string;
+    name?: string;
+    value?: string | Array<string>;
```
LeandroTorresSicilia, line 7: here we need to be more specific about what values are used here:
```
type WeekDays = 'monday' | 'tuesday' | ....
type Value = WeekDays | WeekDays[];
...
value?: Value;
...
```

src/components/WeekDayPicker/index.js
```diff
@@ -0,0 +1,86 @@
...
+                <StyledTextError>{error}</StyledTextError>
+            </RenderIf>
+        </StyledFieldset>
+    );
+};
+
+WeekDayPicker.propTypes = {
+    id: PropTypes.string,
+    name: PropTypes.string,
+    value: PropTypes.oneOfType([PropTypes.string, PropTypes.array]),
```
LeandroTorresSicilia, line 57: same ^^ we need to be more specific:
```
PropTypes.oneOf(['monday', ... , Proptype.arrayOf(...)])
```

src/components/WeekDayPicker/index.js
```diff
@@ -0,0 +1,86 @@
...
+    error: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
+    onChange: PropTypes.func,
+    className: PropTypes.string,
+    style: PropTypes.object,
+};
+
+WeekDayPicker.defaultProps = {
+    id: undefined,
+    name: undefined,
+    value: '',
```
LeandroTorresSicilia, line 73: here let's use undefined too

src/components/WeekDayPicker/readme.md
```diff
@@ -0,0 +1,175 @@
...
+
+```js
+import React, { useState } from 'react';
+import { WeekDayPicker } from 'react-rainbow-components';
+
+const SimpleWeekDayPicker = (props) => {
+    const [day, setDay] = useState('');
+
+    const handleOnChange = e => {
+        const weekDay = e.currentTarget.value;
```
LeandroTorresSicilia, line 11: As declared in the interface this component is meant to send here a string with 'monday', 'tuesday', .... or an array of theses string values.

src/components/WeekDayPicker/readme.md
```diff
@@ -0,0 +1,175 @@
...
+        setDay(weekDay);
+        console.log(weekDay);
```
LeandroTorresSicilia, line 13: remove console log

src/components/WeekDayPicker/readme.md
```diff
@@ -0,0 +1,175 @@
...
+    };
+
+    return (
+        <WeekDayPicker 
+            label="Select a day" 
+            value={day}
+            onChange={(e) => handleOnChange(e)}
```
LeandroTorresSicilia, line 20: here you can use:
```
onChange={handleOnChange}
```

src/components/WeekDayPicker/readme.md
```diff
@@ -0,0 +1,175 @@
+##### WeekDayPicker base:
+
+```js
+import React, { useState } from 'react';
+import { WeekDayPicker } from 'react-rainbow-components';
+
+const SimpleWeekDayPicker = (props) => {
```
LeandroTorresSicilia, line 7: same ^^ use function

src/components/WeekDayPicker/readme.md
```diff
@@ -0,0 +1,175 @@
...
+```js
+import React, { useState, useEffect } from 'react';
+import { WeekDayPicker } from 'react-rainbow-components';
+
+const MultipleWeekDayPicker = (props) => {
+    const [days, setDays] = useState([]);
+
+    useEffect( () => {
+        console.log(days);
+    }, [days])
```
LeandroTorresSicilia, line 41: there is no need to use useEffect here

src/components/WeekDayPicker/readme.md
```diff
@@ -0,0 +1,175 @@
...
+
+    const handleOnChange = e => {
+        const weekDay = e.currentTarget.value;
+        const isChecked = e.currentTarget.checked;
+
+        if (isChecked && !days.includes(weekDay)) {
+            setDays([...days, weekDay]);
+        } else {
+            setDays(days.filter(day => day !== weekDay))
+        }
```
LeandroTorresSicilia, line 51: same ^^ this logic should be implemented in the component index.js since we don't want the users to have the need to reimplement each time they used the component

src/components/WeekDayPicker/readme.md
```diff
@@ -0,0 +1,175 @@
...
+    };
+
+    return (
+        <WeekDayPicker 
+            label="Select active days" 
+            value={days}
+            multiple={true}
+            onChange={(e) => handleOnChange(e)}
```
LeandroTorresSicilia, line 59: same ^^

src/components/WeekDayPicker/readme.md
```diff
@@ -0,0 +1,175 @@
...
+import { WeekDayPicker } from 'react-rainbow-components';
+
+const ReadOnlyWeekDayPicker = (props) => {
+    const [day, setDay] = useState('tuesday');
+
+    return (
+        <WeekDayPicker 
+            label="Select a day (readOnly)" 
+            value={day}
+            readOnly={true}
```
LeandroTorresSicilia, line 82: only `readOnly`

src/components/WeekDayPicker/readme.md
```diff
@@ -0,0 +1,175 @@
...
+import { WeekDayPicker } from 'react-rainbow-components';
+
+const DisabledWeekDayPicker = (props) => {
+
+    return (
+        <WeekDayPicker 
+            label="Select a day" 
+            value={['monday', 'saturday']}
+            multiple={true}
+            disabled={true}
```
LeandroTorresSicilia, line 167: same ^^

src/components/WeekDayPicker/styled/index.js
```diff
@@ -0,0 +1,101 @@
...
+    :not(:last-child) {
+        margin-right: 0.25rem;
+    }
+`;
+
+const StyledWeekDayLabel = attachThemeAttrs(styled.label).attrs(props => {
+    const { palette } = props;
+
+    const defaultColor = palette.text.label;
+    const defaultBackground = '#ffffff';
```
LeandroTorresSicilia, line 50: we can't use colors inline, instead all values should come from theme

src/components/index.js
```diff
@@ -72,4 +72,5 @@ export { default as VerticalSectionOverflow } from './VerticalSectionOverflow';
 export { default as VisualPicker } from './VisualPicker';
 export { default as VisualPickerOption } from './VisualPickerOption';
 export { default as VisualPickerOptionFooter } from './VisualPickerOptionFooter';
+export { default as WeekDayPicker } from './WeekDayPicker';
```
LeandroTorresSicilia, line 75: we also need to export it in index.d.ts

src/components/WeekDayPicker/styled/index.js
```diff
@@ -0,0 +1,101 @@
...
+    font-size: 10px;
+    color: ${props => props.defaultColor};
+    background: ${props => props.defaultBackground};
+    border: 1px solid ${props => props.defaultColor};
+
+    ${props =>
+        !props.readOnly &&
+        !props.disabled &&
+        `
+        &:hover {
```
TahimiLeonBravo, line 75: We need to put the cursor pointer when the hover over the element.
Also, we need to add some shadow brands like other focusable components when the element is in focus.

src/components/WeekDayPicker/styled/index.js
```diff
@@ -0,0 +1,101 @@
...
+            border: 1px solid ${props.activeBackground};
+        }
+    `}
+
+    ${props =>
+        props.isChecked &&
+        `
+        color: ${props.activeColor};
+        background: ${props.activeBackground};
+        border: 1px solid ${props.activeBackground};
```
TahimiLeonBravo, line 87: We need to add some styles when the checked element is hover and focus, it should look like the calendar when someday is selected (you can follow this appearance)

src/components/WeekDayPicker/index.js
```diff
@@ -0,0 +1,86 @@
...
+        style,
+    } = props;
+
+    const defaultFieldsetName = useUniqueIdentifier('week-day-items');
+    const fieldsetName = name || defaultFieldsetName;
+
+    return (
+        <StyledFieldset className={className} style={style} id={id}>
+            <RenderIf isTrue={!hideLabel}>
+                <StyledLabel>
```
TahimiLeonBravo, line 32: You don't need to create a new Label element,  you should reuse the Label we have on the Input component instead of creating another new with the same styles

TahimiLeonBravo:
@blakmetall  let's use the color `text.disabled` for the text when it is disabled
Screen Shot 2020-04-16 at 10 18 58 AM

src/components/WeekDayPicker/index.d.ts
```diff
@@ -0,0 +1,22 @@
...
+
+type WeekDays = 'monday' | 'tuesday' | 'wednesday' | 'thursday' | 'friday' | 'saturday' | 'sunday';
+type Value = WeekDays | WeekDays[];
+
+export interface WeekDayPickerProps extends BaseProps {
+    id?: string;
+    name?: string;
+    value?: Value;
+    label?: ReactNode;
+    hideLabel?: boolean;
```
LeandroTorresSicilia, line 12: remove hideLabel since it is no part of the API

src/components/WeekDayPicker/index.d.ts
```diff
@@ -0,0 +1,22 @@
...
+    disabled?: boolean;
+    required?: boolean;
+    readOnly?: boolean;
+    multiple?: boolean;
+    error?: ReactNode;
+    onChange?: (value: Value) => void;
```
LeandroTorresSicilia, line 18: we should update this file with the new API
remove hideLabel
add availableDates and bottomHelpText

src/components/WeekDayPicker/index.js
```diff
@@ -0,0 +1,144 @@
...
+        error,
+        onChange,
+        className,
+        style,
+    } = props;
+
+    const defaultFieldsetName = useUniqueIdentifier('week-day-items');
+    const fieldsetName = name || defaultFieldsetName;
+
+    const prepareReturnValue = (weekDayValue, isChecked) => {
```
LeandroTorresSicilia, line 37: let's call this method `getNormalizedValue`

src/components/WeekDayPicker/index.js
```diff
@@ -0,0 +1,144 @@
...
+    const handleOnChange = e => {
+        const weekDayValue = e.currentTarget.value;
+        const isChecked = e.currentTarget.checked;
+
+        return onChange(prepareReturnValue(weekDayValue, isChecked));
+    };
+
+    return (
+        <StyledFieldset className={className} style={style} id={id}>
+            <Label label={label} hideLabel={!label} required={required} readOnly={readOnly} />
```
LeandroTorresSicilia, line 58: sorry for this, let's move this back as before since this need to be a legend element and Label components uses a label tag

src/components/WeekDayPicker/index.js
```diff
@@ -0,0 +1,144 @@
...
+        }),
+        PropTypes.oneOf([
+            'monday',
+            'tuesday',
+            'wednesday',
+            'thursday',
+            'friday',
+            'saturday',
+            'sunday',
+            '',
```
LeandroTorresSicilia, line 98: let's not included an empty string here

src/components/WeekDayPicker/index.js
```diff
@@ -0,0 +1,144 @@
...
+        ]),
+    ]),
+    /** The WeekDayPicker label. */
+    label: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
+    /** Shows the help message below the WeekDayPicker */
+    bottomHelpText: PropTypes.string,
```
LeandroTorresSicilia, line 104: this need to be as label and error: ```PropTypes.oneOfType([PropTypes.string, PropTypes.node]),```

src/components/WeekDayPicker/index.js
```diff
@@ -0,0 +1,144 @@
...
+    required: false,
+    readOnly: false,
+    multiple: false,
+    error: null,
+    onChange: () => {},
+    className: undefined,
+    style: undefined,
+};
+
+export default WeekDayPicker;
```
LeandroTorresSicilia, line 144: we need to export in this way:
```export default withReduxForm(WeekDayPicker);```

File: src/components/WeekDayPicker/weekDayItems.js [hunk 30]
```diff
@@ -0,0 +1,93 @@
...
+        }),
+        PropTypes.oneOf([
+            'monday',
+            'tuesday',
+            'wednesday',
+            'thursday',
+            'friday',
+            'saturday',
+            'sunday',
+            '',
```
LeandroTorresSicilia, line 69: same ^^ remove empty string

src/components/WeekDayPicker/index.js
```diff
@@ -0,0 +1,144 @@
...
+        ]),
+    ]),
+    /** The WeekDayPicker label. */
+    label: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
+    /** Shows the help message below the WeekDayPicker */
+    bottomHelpText: PropTypes.string,
+    /** Specifies the available days from the week for selection. */
+    availableDates: PropTypes.arrayOf((propValue, key, componentName, location, propFullName) => {
+        return valuePropValidation(propValue, key, componentName, location, propFullName);
+    }),
```
LeandroTorresSicilia, line 108: here I think you can use:
```availableDates: PropTypes.arrayOf(['monday', ...])```

File: src/components/WeekDayPicker/weekDayItems.js [hunk 32]
```diff
@@ -0,0 +1,93 @@
...
+        ]),
+    ]),
+    availableDates: PropTypes.arrayOf((propValue, key, componentName, location, propFullName) => {
+        return valuePropValidation(propValue, key, componentName, location, propFullName);
+    }),
```
LeandroTorresSicilia, line 74: same ^^

src/components/WeekDayPicker/styled/index.js
```diff
@@ -0,0 +1,94 @@
...
+    text-align: center;
+    font-size: 10px;
+    color: ${props => props.defaultColor};
+    background: ${props => props.defaultBackground};
+    border: 1px solid ${props => props.borderColor};
+    transition: all 0.2s ease;
+
+    ${props =>
+        !props.readOnly &&
+        !props.disabled &&
```
LeandroTorresSicilia, line 57: If this is neither `readOnly` nor `disabled` then it is default right,
why we need to do this check?

src/components/WeekDayPicker/styled/index.js
```diff
@@ -0,0 +1,94 @@
...
+        !props.isChecked &&
+        `
+        border-color: ${props.palette.border.disabled};
+        background-color: ${props.palette.background.disabled};
+    `};
+
+    
+`;
+
+export { StyledFieldset, StyledHelpText, StyledWeekDayContainer, StyledWeekDayLabel };
```
LeandroTorresSicilia, line 94: export this on the const declaration instead the end of the file, you can make this:

`export const StyledName = styled.span`
    ....
`

TahimiLeonBravo:
@blakmetall we need to review the keyboard navigation when the component has the multiple props,
This should go with each element focus when the user presses the `tab` key

TahimiLeonBravo:
@blakmetall  When the component is readOnly it means that it is only for show information, It should not be possible to take actions over them.
let's fix the styles for the selected days. this should not change the styles when hovering and the cursor should not be `pointer`

src/components/WeekDayPicker/readme.md
```diff
@@ -0,0 +1,192 @@
...
+    const [day, setDay] = useState('monday');
+
+    const handleOnChange = weekDay => {
+        setDay(weekDay);
+    };
+
+    return (
+        <WeekDayPicker 
+            label="Week Days" 
+            bottomHelpText="Pick the a day"
```
TahimiLeonBravo, line 17: let's change this text by `Select your working days`

src/components/WeekDayPicker/index.js
blakmetall: If I don't include the empty string, it throws error when using WeekDayPicker for single day selection.

So, as I understand on this part, we will be allowing specific strings, but for the default value, If I don't add that '' empty default value, PropTypes throws the error ;(

If there is another possible solution I can apply, or maybe I need to some advice on this PropTypes definition. Maybe it is a little more difficult because we are limiting user to specific values (which is good practice on this case as I see).

src/components/WeekDayPicker/index.js
blakmetall: Another issue here, I tried and search a bit about it, I reviewed another arrayOf examples used inside the project and I didn't found any clues on how to implement.

Also I reviewed Google and arrayOf could be only for PropTypes types (string, number, etc) I might be wrong.

The only way I saw was implementing the same method above used.

If you see the "value PropTypes configuration", you will see this same code to be able to have an arraOf specific items.

What I am seeing here is that maybe we could extend this method to one of ours to be able to pass an extra parameter to valuePropValidation and maybe we can create a PropTypes helper function for custom arrayOf specific items.

Maybe I am bit crazy I don't know jahahahaha se me ocurrió, I saw an oportunity to extend this prop type method configuration for arrayOf

src/components/WeekDayPicker/styled/index.js
blakmetall: Totally makes sense, even I was shocked for not seeing hahahaha (first thought).

It is needed for the case when user sends specific availableDates array.

So we need the hover action only on enabled days... this is because programmatically is being disabled the days sent by the user.

Another solution might be having the default hover block without restrictions and then I will need to have an extra control to disable the days passed by user. I thought that this solution is shorter and cleaner.

src/components/WeekDayPicker/index.d.ts
```diff
@@ -0,0 +1,23 @@
...
+type WeekDays = 'monday' | 'tuesday' | 'wednesday' | 'thursday' | 'friday' | 'saturday' | 'sunday';
+type Value = WeekDays | WeekDays[];
+
+export interface WeekDayPickerProps extends BaseProps {
+    id?: string;
+    name?: string;
+    value?: Value;
+    label?: ReactNode;
+    bottomHelpText?: string;
+    availableDates: WeekDays[];
```
LeandroTorresSicilia, line 13: can't be required

src/components/WeekDayPicker/weekDay.js
```diff
@@ -21,6 +21,8 @@ export default function WeekDay(props) {
     const weekDayId = useUniqueIdentifier('week-day');
     const inputType = multiple ? 'checkbox' : 'radio';
 
+    const [isFocused, setIsFocused] = useState();
```
LeandroTorresSicilia, line 24: If you want to pass styles when focused you don't need to use state, simply make a selector that takes the label that is a sibling of the input focused

src/components/WeekDayPicker/weekDay.js
```diff
@@ -35,11 +37,15 @@ export default function WeekDay(props) {
                 readOnly={readOnly}
                 error={error}
                 onChange={onChange}
+                onFocus={() => setIsFocused(true)}
+                onBlur={() => setIsFocused(false)}
```
LeandroTorresSicilia, line 41: with the approach above you don't need this

src/components/WeekDayPicker/weekDay.js
```diff
@@ -35,11 +37,15 @@ export default function WeekDay(props) {
...
+                ref={ref}
```
LeandroTorresSicilia, line 42: ref should be only passed to the first input not to all ones

src/components/WeekDayPicker/weekDay.js
```diff
@@ -35,11 +37,15 @@ export default function WeekDay(props) {
...
             />
 
             <StyledWeekDayLabel
                 htmlFor={weekDayId}
                 isChecked={isChecked}
+                isFocused={isFocused}
```
LeandroTorresSicilia, line 48: same ^^ not needed

TahimiLeonBravo:
@blakmetall the days of the week initials shown on the component need be translated
we need to add a new locale prop like another date components

#### Expected Behavior
- When a language is passed on the `locale` it should be translated.
- If the prop `locale` is not passed it should be fallback to the context of the Application locale

src/components/WeekDayPicker/__test__/weekDayPicker.spec.js
```diff
@@ -0,0 +1,157 @@
...
+jest.mock('../../../libs/hooks/useUniqueIdentifier', () => jest.fn(() => 'week-day'));
+
+describe('<WeekDayPicker />', () => {
+    it('should render a label when label prop is passed', () => {
+        const component = mount(<WeekDayPicker label="My label" />);
+        expect(component.find('legend').length).toBe(1);
+    });
+    it('should render bottom help text when bottomHelpText prop is passed', () => {
+        const component = mount(<WeekDayPicker bottomHelpText="My help text" />);
+        expect(component.find('div[children="My help text"]').length).toBe(1);
```
LeandroTorresSicilia, line 16: use StyledHelpText since div is an implementation detail of the HelpText component

src/components/WeekDayPicker/__test__/weekDayPicker.spec.js
```diff
@@ -0,0 +1,157 @@
...
+        expect(component.find(StyledTextError).exists()).toBe(true);
+    });
+    it('should have the opposite days disabled when a list of availableDates is sent', () => {
+        const component = mount(<WeekDayPicker availableDates={['sunday', 'monday']} />);
+        expect(
+            component
+                .find('input')
+                .at(5)
+                .prop('disabled'),
+        ).toBe(true);
```
LeandroTorresSicilia, line 51: try to find a way to iterate through the array of day and check who of them are disabled and who not

src/components/WeekDayPicker/__test__/weekDayPicker.spec.js
```diff
@@ -0,0 +1,157 @@
...
+    });
+    it('should have readonly prop as true when readOnly param is passed', () => {
+        const component = mount(<WeekDayPicker readOnly />);
+        expect(
+            component
+                .find('input')
+                .at(0)
+                .prop('readOnly'),
+        ).toBe(true);
+    });
```
LeandroTorresSicilia, line 76: this test is part of the weekDayItems file

src/components/WeekDayPicker/__test__/weekDayPicker.spec.js
```diff
@@ -0,0 +1,157 @@
...
+        const component = mount(<WeekDayPicker onChange={onChangeFn} disabled />);
+        component
+            .find('input')
+            .at(0)
+            .simulate('change', { target: { checked: true, value: 'sunday' } });
+        expect(onChangeFn).not.toHaveBeenCalled();
+    });
+    it('should run onChange with the right value when multiple is not set', () => {
+        const onChangeFn = jest.fn();
+        const component = mount(<WeekDayPicker value={'monday'} onChange={onChangeFn} />);
```
LeandroTorresSicilia, line 140: use ```value="monday"```
eslint should warn you about this

src/components/WeekDayPicker/index.d.ts
```diff
@@ -0,0 +1,23 @@
...
+
+type WeekDays = 'monday' | 'tuesday' | 'wednesday' | 'thursday' | 'friday' | 'saturday' | 'sunday';
+type Value = WeekDays | WeekDays[];
+
+export interface WeekDayPickerProps extends BaseProps {
+    id?: string;
+    name?: string;
+    value?: Value;
+    label?: ReactNode;
+    bottomHelpText?: string;
```
LeandroTorresSicilia, line 12: bottomHelpText is a ReactNode, we pass all type of text as nodes since there are react internationalization libraries that pass a react node like react intl which pass a component to make the translation:
```<FormattedMessage defaultMessage="my text in english" />```

File: src/components/WeekDayPicker/weekDay.js [hunk 46]
```diff
@@ -0,0 +1,80 @@
...
+        disabled,
+        required,
+        readOnly,
+        multiple,
+        error,
+        onChange,
+    } = props;
+
+    const weekDayId = useUniqueIdentifier('week-day');
+    const inputType = multiple ? 'checkbox' : 'radio';
```
yvmunayev, line 22: We have to optimize as much as possible. Use the hook useMemo () here.

src/components/WeekDayPicker/weekDayItems.js
```diff
@@ -0,0 +1,101 @@
...
+        }
+        return true;
+    };
+
+    const isDisabled = day => {
+        return disabled || !isDayAvailable(day);
+    };
+
+    return getWeekDays().map((weekDay, index) => {
+        const key = `week-day-${index}`;
```
yvmunayev, line 36: You can use the value of the day of the week as a key.

src/components/WeekDayPicker/index.js
```diff
@@ -0,0 +1,163 @@
...
+        style,
+    } = props;
+
+    const context = useContext(AppContext);
+    const locale = getLocale(context, localeProp);
+
+    const defaultFieldsetName = useUniqueIdentifier('week-day-items');
+    const fieldsetName = name || defaultFieldsetName;
+
+    const getNormalizedValue = (weekDayValue, isChecked) => {
```
yvmunayev, line 44: I think this function should happen for a helper

src/components/WeekDayPicker/readme.md
```diff
@@ -0,0 +1,210 @@
...
+        <WeekDayPicker
+            label="Week Days"
+            bottomHelpText="Select your working days"
+            value={day}
+            onChange={handleOnChange}
+        />
+    );
+};
+
+<div className="rainbow-align-content_center rainbow-m-around_xx-large">
```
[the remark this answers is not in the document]
LeandroTorresSicilia: In the examples, I think we should keep using classNames whenever possible in order to not pollute the example with too much code.

src/components/WeekDayPicker/weekDay.js
LeandroTorresSicilia: Here I think there is no need to use useMemo, keep in mind that useMemo is a function that does more things inside it and it is called every renders too, we should use this optimization when:
1. Referential equality is needed to keep thing at the same reference like objects, arrays or functions.
2. Computationally expensive calculations.
Here is a good guide:
https://kentcdodds.com/blog/usememo-and-usecallback

src/components/WeekDayPicker/__test__/weekDayPicker.spec.js
```diff
@@ -112,11 +116,24 @@ describe('<WeekDayPicker />', () => {
...
+            component
+                .find(WeekDayItems)
+                .at(0)
+                .prop('name'),
+        ).toBe('week-day-items');
+    });
+    it('should have locale configured properly if locale is sent', () => {
+        const component = mount(<WeekDayPicker locale="en-US" />);
+        expect(component.prop('locale')).toBe('en-US');
+    });
```
LeandroTorresSicilia, line 136: what is being tested here?
if you pass ```locale="en-US"``` to WeekDayPicker then you don't need to check that itself receive the locale prop

src/components/WeekDayPicker/helpers/__test__/getWeekDayLabel.spec.js
```diff
@@ -0,0 +1,23 @@
...
+    it('should return empty string when day is not valid', () => {
+        const label = getWeekDayLabel('wrong-day-value');
+        expect(label).toBe('');
+    });
+    /*
+    it('should return day label in spanish when locale is es-MX', () => {
+        const locale = 'es-MX';
+        const label = getWeekDayLabel('sunday', locale);
+        expect(label).toBe('D');
+    });
```
LeandroTorresSicilia, line 21: no comment code

src/components/WeekDayPicker/helpers/__test__/getWeekDayLabel.spec.js
```diff
@@ -0,0 +1,23 @@
...
+        expect(label).toBe('S');
+    });
+    it('should return empty string when no day is sent', () => {
+        const label = getWeekDayLabel();
+        expect(label).toBe('');
+    });
+    it('should return empty string when day is not valid', () => {
+        const label = getWeekDayLabel('wrong-day-value');
+        expect(label).toBe('');
+    });
```
LeandroTorresSicilia, line 15: add test case for Spanish locale

LeandroTorresSicilia:
The last example must be like the design:
Screen Shot 2020-04-28 at 9 12 54 AM
It should be interactive with a language picker

integration/specs/WeekDayPicker/weekDayPicker-1.spec.js
```diff
@@ -0,0 +1,49 @@
...
+        browser.url('/#!/WeekDayPicker/1');
+    });
+    beforeEach(() => {
+        browser.refresh();
+        const component = $(WEEKDAY_PICKER);
+        component.waitForExist();
+    });
+    it('should check the option clicked', () => {
+        const weekDayPicker = new PageWeekDayPicker(WEEKDAY_PICKER);
+        weekDayPicker.clickOn('sunday');
```
LeandroTorresSicilia, line 17: let's use click to be consistent

integration/specs/WeekDayPicker/weekDayPicker-15.spec.js
```diff
@@ -0,0 +1,18 @@
...
+        browser.url('/#!/WeekDayPicker/15');
+    });
+    beforeEach(() => {
+        browser.refresh();
+        const component = $(WEEKDAY_PICKER);
+        component.waitForExist();
+    });
+    it('should have label localized', () => {
+        const weekDayPicker = new PageWeekDayPicker(WEEKDAY_PICKER);
+        expect(weekDayPicker.getInputLabel('monday')).toBe('L');
```
LeandroTorresSicilia, line 16: this should be part of a unit test, let's leave integrations tests for interactions

src/components/WeekDayPicker/helpers/__test__/getWeekDayLabel.spec.js
```diff
@@ -0,0 +1,16 @@
...
```
LeandroTorresSicilia, line 15: add test case for Spanish locale here since it is part of the unit test

src/components/WeekDayPicker/index.js
```diff
@@ -0,0 +1,148 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import valuePropValidation from './helpers/valuePropValidation';
+import getNormalizedValue from './helpers/getNormalizedValue';
+import { useUniqueIdentifier, useLocale } from '../../libs/hooks';
+import withReduxForm from './../../libs/hocs/withReduxForm';
```
LeandroTorresSicilia, line 6: instead of use withReduxForm let's use useReduxForm which is a new hook to integrate with redux-form, update with the master branch to get last changes and look at StrongPasswordInput component to see how it is used

src/components/WeekDayPicker/index.js
```diff
@@ -0,0 +1,148 @@
...
+
+WeekDayPicker.propTypes = {
+    /** The id of the outer element. */
+    id: PropTypes.string,
+    /** An identifier for the group of radio items. */
+    name: PropTypes.string,
+    /** The value of WeekDayPicker, a list of days selected. */
+    value: PropTypes.oneOfType([
+        PropTypes.arrayOf((propValue, key, componentName, location, propFullName) => {
+            return valuePropValidation(propValue, key, componentName, location, propFullName);
```
LeandroTorresSicilia, line 89: I think here you can use:
```PropTypes.arrayOf(valuePropValidation)```

File: src/components/WeekDayPicker/index.js [hunk 58]
```diff
@@ -0,0 +1,148 @@
...
+            'sunday',
+        ]),
+    ]),
+    /** The WeekDayPicker label. */
+    label: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
+    /** Shows the help message below the WeekDayPicker */
+    bottomHelpText: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
+    /** Specifies the available days from the week for selection. */
+    availableDates: PropTypes.arrayOf((propValue, key, componentName, location, propFullName) => {
+        return valuePropValidation(propValue, key, componentName, location, propFullName);
```
LeandroTorresSicilia, line 107: same ^^

src/components/WeekDayPicker/pageObject/index.js
```diff
@@ -0,0 +1,47 @@
...
+    /**
+     * Create a new PageWeekDayPicker page object.
+     * @constructor
+     * @param {string} rootElement - The selector of the PageWeekDayPicker root element.
+     */
+    constructor(rootElement) {
+        this.rootElement = rootElement;
+    }
+
+    clickOn(weekDay) {
```
LeandroTorresSicilia, line 17: these method should be documented since they are exposed

src/components/WeekDayPicker/pageObject/index.js
```diff
@@ -0,0 +1,47 @@
...
+        return this.getInputRef(weekDay).getText();
+    }
+
+    getInput(weekDay) {
+        const index = weekDays.findIndex(value => value === weekDay);
+        const elem = $(this.rootElement).$$('input');
+        return elem[index];
+    }
+
+    getInputRef(weekDay) {
```
LeandroTorresSicilia, line 40: only document those who are exposed

src/components/WeekDayPicker/pageObject/index.js
```diff
@@ -0,0 +1,47 @@
+const weekDays = ['sunday', 'monday', 'tuesday', 'wednesday', 'thursday', 'friday', 'saturday'];
```
LeandroTorresSicilia, line 1: you can reuse getWeekDays

src/components/WeekDayPicker/readme.md
```diff
@@ -0,0 +1,232 @@
...
+                    locale="ru-RU" 
+                    onChange={(day) => setDay2(day)} />
+            </div>
+            <div className="rainbow-m-vertical_x-large">
+                <WeekDayPicker 
+                    label="Japanese" 
+                    value={day3} 
+                    locale="ja-JP" 
+                    onChange={(day) => setDay3(day)} />
+            </div>
```
LeandroTorresSicilia, line 224: this example should be like design with a selector

src/components/WeekDayPicker/weekDayItems.js
```diff
@@ -0,0 +1,99 @@
...
+            />
+        );
+    });
+});
+
+WeekDayItems.propTypes = {
+    name: PropTypes.string,
+    value: PropTypes.oneOfType([
+        PropTypes.arrayOf((propValue, key, componentName, location, propFullName) => {
+            return valuePropValidation(propValue, key, componentName, location, propFullName);
```
LeandroTorresSicilia, line 62: same ^^

src/components/WeekDayPicker/weekDayItems.js
```diff
@@ -0,0 +1,99 @@
...
+            'tuesday',
+            'wednesday',
+            'thursday',
+            'friday',
+            'saturday',
+            'sunday',
+        ]),
+    ]),
+    availableDates: PropTypes.arrayOf((propValue, key, componentName, location, propFullName) => {
+        return valuePropValidation(propValue, key, componentName, location, propFullName);
```
LeandroTorresSicilia, line 75: same ^^

Issue #1490, opened 2020-04-21
feat: implement TimeSelectInput


TahimiLeonBravo:
#### Requirements
- [ ] Define API
- [ ] Implement component
- [ ] Add test

##### API
https://docs.google.com/document/d/15UoiGEEn5TwfPGn8WREJD_c0CciCcREIyvEk2cvbhKk/edit?usp=sharing

PR #1496, opened 2020-04-22
feat: new component TimeSelectInput

rafaelmg83:
fix: #1490 

- New Component TimeSelectInput

@nexxtway/react-rainbow


src/components/TimeSelectInput/index.js
```diff
@@ -0,0 +1,509 @@
...
+        let normalizedValue;
+
+        if (isNumber(value)) {
+            this.value = value;
+            if (Number(value) > 23 || this.isUpOrDownKeyPressed || this.hasPropValue) {
+                const newTypedValue = getSingleNewTypedValue(hour, value);
+                normalizedValue = normalizeHour(newTypedValue, hour24);
+                this.setState({
+                    hour: normalizedValue,
+                });
```
yvmunayev, line 118: call handleChangeTime({
hour: normalizedValue,
});

src/components/TimeSelectInput/index.js
```diff
@@ -0,0 +1,509 @@
...
+                const newTypedValue = getSingleNewTypedValue(minutes, value);
+                normalizedValue = normalizeMinutes(newTypedValue);
+                this.setState({
+                    minutes: normalizedValue,
+                });
+            } else {
+                normalizedValue = normalizeMinutes(value);
+                this.setState({
+                    minutes: normalizedValue,
+                });
```
yvmunayev, line 189: handleChangeTime({
minutes: normalizedValue,
});

src/components/TimeSelectInput/index.js
```diff
@@ -0,0 +1,509 @@
...
+        const { hour24 } = this.props;
+        let normalizedValue;
+
+        if (isNumber(value)) {
+            if (Number(value) > 60 || this.isUpOrDownKeyPressed) {
+                const newTypedValue = getSingleNewTypedValue(minutes, value);
+                normalizedValue = normalizeMinutes(newTypedValue);
+                this.setState({
+                    minutes: normalizedValue,
+                });
```
yvmunayev, line 184: handleChangeTime({
minutes: normalizedValue,
});

src/components/TimeSelectInput/index.js
```diff
@@ -0,0 +1,509 @@
...
+            minutes: '',
+        });
+    }
+
+    handleAmPmChange(value) {
+        this.setState({
+            inputFocusedIndex: -1,
+            ampm: value,
+        });
+        setTimeout(() => this.handleChangeTime(), 0);
```
yvmunayev, line 217: handleChangeTime({
inputFocusedIndex: -1,
            ampm: value,
});

src/components/TimeSelectInput/index.js
```diff
@@ -0,0 +1,509 @@
...
+            currentInputFocused.current.focus();
+        }
+    }
+
+    resetState() {
+        const { inputFocusedIndex } = this.state;
+        if (inputFocusedIndex === 0) {
+            this.setState({
+                hour: '',
+            });
```
yvmunayev, line 312: handleChangeTime({
hour: '',
});

src/components/TimeSelectInput/index.js
```diff
@@ -0,0 +1,509 @@
...
+            this.prevHour = '';
+        }
+        if (inputFocusedIndex === 1) {
+            this.setState({
+                minutes: '',
+            });
```
yvmunayev, line 318: handleChangeTime({
minutes: '',
});

src/components/TimeSelectInput/index.js
```diff
@@ -0,0 +1,509 @@
...
+            this.prevMinutes = '';
+        }
+        setTimeout(() => this.handleChangeTime(), 0);
```
yvmunayev, line 321: delete this call

src/components/TimeSelectInput/index.js
```diff
@@ -0,0 +1,509 @@
...
+    }
+
+    incrementHour() {
+        const { hour } = this.state;
+        const { hour24 } = this.props;
+        const hourValue = hour || this.prevHour;
+        this.setState({
+            hour: normalizeHour(getNextHour(hourValue, hour24), hour24),
+        });
+        setTimeout(() => this.handleChangeTime(), 0);
```
yvmunayev, line 345: handleChangeTime({
hour: normalizeHour(getNextHour(hourValue, hour24), hour24),
});

src/components/TimeSelectInput/index.js
```diff
@@ -0,0 +1,509 @@
...
+    }
+
+    decrementHour() {
+        const { hour } = this.state;
+        const { hour24 } = this.props;
+        const hourValue = hour || this.prevHour;
+        this.setState({
+            hour: normalizeHour(getPrevHour(hourValue, hour24), hour24),
+        });
+        setTimeout(() => this.handleChangeTime(), 0);
```
yvmunayev, line 355: same before

src/components/TimeSelectInput/index.js
```diff
@@ -0,0 +1,509 @@
...
+    }
+
+    incrementMinutes() {
+        const { minutes } = this.state;
+        const minutesValue = minutes || this.prevMinutes;
+        this.setState({
+            minutes: normalizeMinutes(getNextMinute(minutesValue)),
+        });
+        setTimeout(() => this.handleChangeTime(), 0);
```
yvmunayev, line 364: same before

src/components/TimeSelectInput/index.js
```diff
@@ -0,0 +1,509 @@
...
+    }
+
+    decrementMinutes() {
+        const { minutes } = this.state;
+        const minutesValue = minutes || this.prevMinutes;
+        this.setState({
+            minutes: normalizeMinutes(getPrevMinute(minutesValue)),
+        });
+        setTimeout(() => this.handleChangeTime(), 0);
```
yvmunayev, line 373: same before

src/components/TimeSelectInput/readme.md
```diff
@@ -0,0 +1,54 @@
...
+
+```js
+import React from 'react';
+import { TimeSelectInput } from 'react-rainbow-components';
+
+const containerStyles = {
+    display: "flex",
+    justifyContent: "center",
+};
+<div style={containerStyles}>
```
yvmunayev, line 11: use styled-component for Container

src/components/TimeSelectInput/readme.md
```diff
@@ -0,0 +1,54 @@
...
+```js
+import React from 'react';
+import { TimeSelectInput } from 'react-rainbow-components';
+
+const containerStyles = {
+    display: "flex",
+    justifyContent: "center",
+};
+
+<div style={containerStyles}>
```
yvmunayev, line 29: same before

src/components/TimeSelectInput/readme.md
```diff
@@ -0,0 +1,54 @@
...
```
yvmunayev, line 48: same before

src/components/TimeSelectInput/readme.md
```diff
@@ -3,52 +3,55 @@
...
-</div>
+</StyledContainer>
 ```
 
 ##### TimeSelectInput with initial value:
 
 ```js
 import React from 'react';
 import { TimeSelectInput } from 'react-rainbow-components';
```
yvmunayev, line 24: Same before

src/components/TimeSelectInput/readme.md
```diff
@@ -3,52 +3,55 @@
...
```
yvmunayev, line 6: External import before internal import always

src/components/TimeSelectInput/readme.md
```diff
@@ -3,52 +3,55 @@
...
-</div>
+</StyledContainer>
 ```
 
 ##### TimeSelectInput with 24 hour format:
 
 ```js
 import React from 'react';
 import { TimeSelectInput } from 'react-rainbow-components';
```
yvmunayev, line 44: Same before

src/components/TimeSelectInput/index.js
```diff
@@ -0,0 +1,503 @@
...
+        const { hour24 } = this.props;
+        let normalizedValue;
+
+        if (isNumber(value)) {
+            if (Number(value) > 60 || this.isUpOrDownKeyPressed) {
+                const newTypedValue = getSingleNewTypedValue(minutes, value);
+                normalizedValue = normalizeMinutes(newTypedValue);
+                this.setState({
+                    minutes: normalizedValue,
+                });
```
yvmunayev, line 185: I have here you shouldn't call handleChangeTime (newState) to call onChange ()

src/components/TimeSelectInput/readme.md
```diff
@@ -0,0 +1,57 @@
...
+
+const StyledContainer = styled.div`
+    display: flex;
+    justify-content: center;
+`;
+<StyledContainer>
+    <TimeSelectInput
+        id="time-select-input-1"
+    />;
+</StyledContainer>
```
yvmunayev, line 16: All Inputs use the "value" and "onChange ()" pattern. To do the examples it had to be like this.

`const TimeSelectInputExample = () => {
    const [time, setTime] = useState('08:00 AM');
    return (
        <TimeSelectInput
            id="time-select-input-2"
            value="08:00 AM"
            onChange={setTime}
        />
    );
}
<StyledContainer>
    <TimeSelectInputExample />
</StyledContainer>
`

src/components/TimeSelectInput/readme.md
```diff
@@ -0,0 +1,57 @@
...
+    display: flex;
+    justify-content: center;
+`;
+
+<StyledContainer>
+    <TimeSelectInput
+        id="time-select-input-2"
+        value="08:00 AM"
+    />;
+</StyledContainer>
```
yvmunayev, line 36: same before

src/components/TimeSelectInput/readme.md
```diff
@@ -0,0 +1,57 @@
...
+    display: flex;
+    justify-content: center;
+`;
+
+<StyledContainer>
+    <TimeSelectInput
+        id="time-select-input-2"
+        hour24={true}
+    />;
+</StyledContainer>
```
yvmunayev, line 56: same before

src/components/TimeSelectInput/index.js
```diff
@@ -0,0 +1,503 @@
...
+            minutes: currentMinutes,
+            ampm: currentAmPm,
+        });
+        this.setState(newState);
+
+        if (
+            (currentHour && currentMinutes && ampm && !hour24) ||
+            (currentHour && currentMinutes && hour24)
+        ) {
+            onChange(time);
```
yvmunayev, line 385: @rafaelmg83  The behavior of all our Input components is, that when you change the value you have to call the onChange method. You have to check that behavior well in this component, it is working fine.
I already told you to change the examples to reflect that behavior, so you can try adding a console.log (value) in the onChange and check that every time you change the date in the component it makes you console.log.

rafaelmg83: I don't know what you mean exactly, but the onChange event is called every time that any part of hour, minutes or am/pm changes except the value is incomplete where it has no logic to call it. Of course I tried the console.log test before push. Maybe you are seeing something that I am not, but I think that behavior is fine as it is.

src/components/TimeSelectInput/readme.md
```diff
@@ -0,0 +1,78 @@
...
+const StyledContainer = styled.div`
+    display: flex;
+    justify-content: center;
+`;
+const TimeSelectInputExample = () => {
+    const [time, setTime] = useState(null); 
+    return ( 
+        <TimeSelectInput 
+            id="time-select-input-2" 
+            value=""
```
LeandroTorresSicilia, line 17: since this component is an inputable it must be controlled, then the source of truth is the value prop, then if you don't pass the time state here as the value it should not work

src/components/TimeSelectInput/index.d.ts
```diff
@@ -0,0 +1,12 @@
+import { ComponentType } from 'react';
+import { BaseProps } from '../types';
+
+export interface TimeSelectInputProps extends BaseProps {
+    value?: string;
+    onChange?: (time: string) => void;
```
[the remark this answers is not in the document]
rafaelmg83: @yvmunayev I fixed the first error already, but, I don't understand what you mean with the second one, if it is the case you try to enter a number higher (6-9) than 5 [minutes: 0-59] it assumes that is the entire minute and focus the next input (am/pm). By ex. if you enter 7 the minutes will be: 07. Makes sense?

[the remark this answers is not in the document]
LeandroTorresSicilia: @yvmunayev We discussed this and we think it should be a string with the time as a string formatted and ready to store in a database or whatever place the user can use, right now the TimePicker send it in this way too (as a formatted string). But we can discuss it again to see your points.

src/components/TimeSelectInput/helpers/index.js
```diff
@@ -0,0 +1,39 @@
+import get12HourTime from './get12HourTime';
```
LeandroTorresSicilia, line 1: here you can:
```
export { default as get12HourTime } from './get12HourTime';
export { default as get24HourTime } from './get24HourTime';
...
```
and then it gets shorter importing and exporting in only one line each one

src/components/TimeSelectInput/icons/index.js
```diff
@@ -0,0 +1,4 @@
+import DownArrow from './downArrow';
+import UpArrow from './upArrow';
+
+export { DownArrow, UpArrow };
```
LeandroTorresSicilia, line 4: same ^^

LeandroTorresSicilia:
- when the hour input has a value and press delete key it does not clean its value, we should add test to this behavior

src/components/TimeSelectInput/index.js
```diff
@@ -0,0 +1,505 @@
...
+            this.handleLeftKeyPressed();
+        }
+        if (keyCode === UP_KEY) {
+            this.incrementHandler();
+        }
+        if (keyCode === DOWN_KEY) {
+            this.decrementHandler();
+        }
+        if (keyCode === DELETE_KEY) {
+            this.resetState();
```
LeandroTorresSicilia, line 238: this is not working properly since the hour input value does not reset, we should add test to this behavior

src/components/TimeSelectInput/index.js
```diff
@@ -0,0 +1,505 @@
...
+        const currentInputFocused = this.inputsMap[inputFocusedIndex];
+        if (currentInputFocused) {
+            currentInputFocused.current.focus();
+        }
+    }
+
+    resetState() {
+        const { inputFocusedIndex } = this.state;
+        if (inputFocusedIndex === 0) {
+            this.handleChangeTime()({
```
LeandroTorresSicilia, line 307: sure this is the problem resetting the hour value you are calling it twice and the first time without arguments:
```this.handleChangeTime()({ ... })```

File: src/components/TimeSelectInput/styled/index.js [hunk 29]
```diff
@@ -0,0 +1,301 @@
...
+export {
+    StyledContainer,
+    StyledDots,
+    StyledDownArrow,
+    StyledInputHidden,
+    StyledOptionLabel,
+    StyledSelectContent,
+    StyledSelectValue,
+    StyledUpArrow,
+    StyledVerticalButtonsContainer,
```
LeandroTorresSicilia, line 300: you can export directly in each const

LeandroTorresSicilia:
We also need to add test to index, currently we have test of this in `src/components/TimePicker/__test__/timeSelect.spec.js` and also add integration tests but you can wait to have the correct behavior to add these tests in order to make the right changes.

src/components/TimeSelectInput/helpers/normalizeValue.js
```diff
@@ -0,0 +1,56 @@
...
+    if (!hour24 && Number(hour) < 12) {
+        if (!ampm) ampm = 'AM';
+        if (hour === '00') hour = 12;
+    }
+
+    return {
+        hour: normalizeHour(hour, hour24),
+        minutes: normalizeMinutes(minutes, hour24),
+        ampm,
+    };
```
LeandroTorresSicilia, line 55: we also need to add test to this function

TahimiLeonBravo:
@rafaelmg83, when navigating to the component page all the TimeSelectInput examples, have focused the hour's element 
Screen Shot 2020-05-04 at 9 45 16 AM

src/components/TimeSelectInput/index.d.ts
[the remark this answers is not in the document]
yvmunayev: @rafaelmg83  try putting 12:12 on the keyboard several times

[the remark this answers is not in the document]
rafaelmg83: > @rafaelmg83 try putting 12:12 on the keyboard several times

fixed! Thanks

rafaelmg83:
> check that you have some lint problems

fixed! Thanks.

src/components/WeeklyCalendar/readme.md
```diff
@@ -96,7 +96,7 @@ const initialState = {
     currentWeek: new Date(),
 };
 
-<div className="rainbow-m-around_large">
+<div className="rainbow-m-around_large ">
```
yvmunayev, line 99: check this

LeandroTorresSicilia:
We need to be able to set the focus programmatically like when de component mount the first time:
```
  const ref = useRef();
  	useEffect(() => {
    	ref.current.focus()
    }, []);
    return (
        <TimeSelectInput
            id="time-select-input-1"
            value={time} 
            onChange={setTime}
            label="TimeSelectInput Label"
            ref={ref}
        />
```

LeandroTorresSicilia:
This should not happen, if value is `08:00 A` then the component should render the placeholder `--:-- --`
Screen Shot 2020-05-23 at 9 16 34 AM

LeandroTorresSicilia:
@TahimiLeonBravo check disabled and readOnly styles since there is no visual difference between them

src/components/TimeSelectInput/index.d.ts
```diff
@@ -0,0 +1,23 @@
...
+    hideLabel?: boolean;
+    required?: boolean;
+    bottomHelpText?: ReactNode;
+    error?: ReactNode;
+    readOnly?: boolean;
+    disabled?: boolean;
+    tabIndex?: number | string;
+    onClick?: (event: MouseEvent<HTMLInputElement>) => void;
+    onFocus?: (value: string) => void;
+    onBlur?: (value: string) => void;
```
LeandroTorresSicilia, line 17: There are inconsistencies here, currently, blur sends a string and it is fired only once like if it was only one input (good until here), but initially blur without any value it sends `null` instead of an empty string.
In the case of the focus, it behaves completely different, it fires an event object instead of the string and it fires every time for each input.
Let me check but I think we should send on both a string empty when no value or incomplete or the value when correct, and treat the input as only one, it means it should not send focus for each input

src/components/TimeSelectInput/index.js
```diff
@@ -94,6 +100,7 @@ export default class TimeSelectInput extends Component {
 
     handleClickOutside() {
         const { inputFocusedIndex } = this.state;
+        this.handleBlur();
```
LeandroTorresSicilia, line 103: this is causing the blur to get called every time for each input, this is not a correct behaviour since only focused element should be able to blur

src/components/TimeSelectInput/index.js
```diff
@@ -0,0 +1,633 @@
...
+        this.containerRef.current.focus();
+    }
+
+    /**
+     * Sets click on the element.
+     * @public
+     */
+    click() {
+        this.containerRef.current.click();
+    }
```
LeandroTorresSicilia, line 426: we need to add blur method too, like all other input components

src/components/TimeSelectInput/index.js
```diff
@@ -0,0 +1,633 @@
...
+                        onChange={this.handleChangeHour}
+                        onFocus={this.handleFocusHour}
+                        onBlur={this.handleBlurHour}
+                        inputMode="numeric"
+                        isFocused={inputFocusedIndex === 0}
+                        pattern="\d*"
+                        ref={this.hourInputRef}
+                        disabled={disabled}
+                        readOnly={readOnly}
+                        tabIndex={tabIndex || null}
```
LeandroTorresSicilia, line 499: There is a reason to use null?
If not I think we should simply do:
```tabIndex={tabIndex}```

File: src/components/TimeSelectInput/index.js [hunk 36]
```diff
@@ -408,11 +464,20 @@ export default class TimeSelectInput extends Component {
...
 
         return (
-            <article className={className} style={style} id={id}>
+            <StyledContainer className={className} style={style} id={id}>
+                <RenderIf isTrue={!!label}>
+                    <StyledLabel
+                        label={label}
+                        hideLabel={hideLabel}
+                        required={required}
+                        inputId="timeselectinput-label"
```
LeandroTorresSicilia, line 473: we need to research here what is the best approach for accessibility since this is for associate the input with the label, we can't simply pass a fixed text here, also normally this value is generated and unique since it can't be used by another element in the DOM

src/components/TimeSelectInput/readme.md
```diff
@@ -69,6 +71,92 @@ const TimeSelectInputExample = () => {
...
+
+const StyledContainer = styled.div`
+    display: flex;
+    justify-content: center;
+`;
+const TimeSelectInputExample = () => {
+    const [time, setTime] = useState(null); 
+    return ( 
+        <TimeSelectInput 
+            id="time-select-input-5"
```
LeandroTorresSicilia: ids must be unique, I see you are using `time-select-input-5` in each one

rafaelmg83:
> This should not happen, if value is `08:00 A` then the component should render the placeholder `--:-- --`
> Screen Shot 2020-05-23 at 9 16 34 AM

Fixed!

rafaelmg83:
> We need to be able to set the focus programmatically like when de component mount the first time:
> 
> ```
>   const ref = useRef();
>     useEffect(() => {
>     	ref.current.focus()
>     }, []);
>     return (
>         <TimeSelectInput
>             id="time-select-input-1"
>             value={time} 
>             onChange={setTime}
>             label="TimeSelectInput Label"
>             ref={ref}
>         />
> ```

Fixed!

rafaelmg83:
> @TahimiLeonBravo check disabled and readOnly styles since there is no visual difference between them

Fixed!

rafaelmg83:
> @rafaelmg83, when navigating to the component page all the TimeSelectInput examples, have focused the hour's element
> Screen Shot 2020-05-04 at 9 45 16 AM

Fixed!

src/components/TimeSelectInput/index.js
```diff
@@ -596,6 +608,8 @@ export default class TimeSelectInput extends Component {
     }
 }
 
+export default withForwardedRef(TimeSelectInput);
```
LeandroTorresSicilia, line 612: we need to export this component withReduxForm HOC

src/components/TimeSelectInput/helpers/normalizeValue.js
```diff
@@ -16,7 +16,9 @@ export default function normalizeValue(value, hour24) {
         ampm: '',
     };
 
-    if (!value) return invalidValue;
+    if (!value || hour24 === undefined || hour24 === null || typeof hour24 !== 'boolean') {
```
LeandroTorresSicilia, line 19: only checking for ```typeof hour24 !== 'boolean'``` is sufficent

src/components/TimeSelectInput/helpers/__test__/normalizeValue.spec.js
```diff
@@ -0,0 +1,137 @@
...
+        return '12';
+    }
+    if (+value < 10) {
+        return `0${+value}`;
+    }
+    if (+value > 12 && +value < 20 && !hour24) {
+        return `0${+value - 12}`;
+    }
+    return String(+value);
+});
```
LeandroTorresSicilia, line 71: we should not reimplement helpers in its mocks, since it is difficult to read and understand, also how you guaranties the right value returned, if we use this complex implementations in the mocks the we would need to test the mock itself, instead what you can do is to return a simple value like an empty string and then in each test where a specific value is needed, you can do:
```normalizeHour.mockReturnValue('12')```

LeandroTorresSicilia:
@TahimiLeonBravo check this styles, when readonly it has different colors
Screen Shot 2020-05-30 at 9 17 40 AM

rafaelmg83:
> @TahimiLeonBravo check this styles, when readonly it has different colors
> Screen Shot 2020-05-30 at 9 17 40 AM

Fixed!

PR #1498, opened 2020-04-23
feat: add the bottomHelpText prop to the Select component

rgah2107:
fix: #1313

- add the bottomHelpText prop to the Select component

@nexxtway/react-rainbow


src/components/Select/index.js
```diff
@@ -111,6 +116,8 @@ Select.propTypes = {
     onFocus: PropTypes.func,
     /** The action triggered when the element releases focus. */
     onBlur: PropTypes.func,
+    /** Text help for the select. */
```
yvmunayev, line 119: change to /** Shows the help message below the Select. */

src/components/Select/index.js
```diff
@@ -145,6 +152,7 @@ Select.defaultProps = {
     onClick: () => {},
     onFocus: () => {},
     onBlur: () => {},
+    bottomHelpText: undefined,
```
TahimiLeonBravo, line 155: in order to be more consistent, let's use `null` instead `undefined` as default prop

PR #1500, opened 2020-04-24
feat: application default locale with getBrowserLocale - first revision

blakmetall:
feat: default added to AppContext

----

notes:

### DIFF
--- a/src/components/Application/context.js
+++ b/src/components/Application/context.js
@@ -1,8 +1,4 @@
 import React from 'react';
-import getBrowserLocale from '../../libs/utils/getBrowserLocale';
 
-const defaultContext = {
-    locale: getBrowserLocale(),
-};
-export const AppContext = React.createContext(defaultContext);
+export const AppContext = React.createContext();
 export const { Provider, Consumer } = AppContext;
--- a/src/components/Application/context.js
+++ b/src/components/Application/context.js
@@ -1,4 +1,8 @@
 import React from 'react';
+import getBrowserLocale from '../../libs/utils/getBrowserLocale';
 
-export const AppContext = React.createContext();
+const defaultContext = {
+    locale: getBrowserLocale(),
+};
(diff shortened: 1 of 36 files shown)


src/components/Application/index.js
```diff
@@ -13,7 +13,9 @@ import normalizeTheme from '../../styles/helpers/normalizeTheme';
  */
 export default function Application(props) {
     const { children, className, style, locale, theme } = props;
-    const contextValue = { locale };
+
+    const context = useContext(AppContext);
+    const contextValue = locale ? { locale } : { locale: context.locale };
```
yvmunayev, line 18: Talk to Leo and we decide we're going to do this.
You have reversed the useLocale and AppContext changes.
If we do this it will already be:

`const { children, className, style, locale: localLocale, theme } = props;
    const locale = useLocale(localLocale);
    const contextValue = { locale };`

src/libs/hooks/useLocale.js
```diff
@@ -1,11 +1,7 @@
 import { useContext, useMemo } from 'react';
 import { AppContext } from '../../components/Application/context';
-import getBrowserLocale from '../utils/getBrowserLocale';
 
 export default function useLocale(localProp) {
     const context = useContext(AppContext);
-    return useMemo(() => localProp || (context && context.locale) || getBrowserLocale(), [
-        localProp,
-        context,
-    ]);
```
yvmunayev, line 10: reverse changes

src/components/Application/context.js
```diff
@@ -1,4 +1,8 @@
 import React from 'react';
+import getBrowserLocale from '../../libs/utils/getBrowserLocale';
 
-export const AppContext = React.createContext();
+const defaultContext = {
+    locale: getBrowserLocale(),
+};
+export const AppContext = React.createContext(defaultContext);
```
yvmunayev, line 7: reverse changes

Issue #1507, opened 2020-04-28
feat: implement FileSelector component


TahimiLeonBravo:
- [ ] define API
- [ ] Implement component
- [ ] add unit and integration tests

##### Define API
https://docs.google.com/document/d/1U2uJ6BVtBmRVD-n3vQ9m6LjvkF_9ZmYgSI70fZ8Mbow/edit?usp=sharing

##### Design
https://zpl.io/brZAy4e

##### Requirements
- [ ] It should accept the same attributes of the native input
- [ ] It should be work with redux-form
- [ ] It only should accept select or drag & drop files, not other elements.
- [ ] It only should allow select files, it not is possible to initialize your state with selected elements.

###### fileSelector
- [ ] should render an input with type file
- [ ] should set the same generated value to the label htmlFor and the input id
- [ ] should set the tabIndex passed in the input element
- [ ] should set disabled in the input element if it is passed
- [ ] should set required in the input element if it is passed
- [ ] should pass a generated id to the error element and set the same id to the aria-describedby for the input when a error is passed
- [ ] should pass the right props to the Label component
- [ ] should render HelpText when bottomHelpText is passed
- [ ] should fire change event when a file is picked
- [ ] should fire focus event
- [ ] should fire blur event
###### getText
- [ ] should return placeholder when files is not defined
- [ ] should return placeholder when files is an empty array
- [ ] should return the name of the file when files has only one item
- [ ] should return a string with the length of files
###### containsFiles
- [ ] should return false when there is types different than "Files"
- [ ] should return false when types array is empty
- [ ] should return false when types array is undefined
- [ ] should return true when all types are "Files"

PR #1516, opened 2020-05-02
feat: implement FileSelector component

HellWolf93:
fix: #1507

- implement FileSelector component
____


src/components/FileSelector/index.js
```diff
@@ -0,0 +1,320 @@
+import React, { useRef, useState, useEffect, useImperativeHandle } from 'react';
+import PropTypes from 'prop-types';
+import Label from '../Input/label';
+import useReduxForm from '../../libs/hooks/useReduxForm';
```
yvmunayev, line 4: put it together with the other global hooks

src/components/FileSelector/index.js
```diff
@@ -0,0 +1,320 @@
...
+            return <ErrorIcon />;
+        }
+        if (files && files.length === 1) {
+            return <FileIcon />;
+        }
+        if (files && files.length > 1) {
+            return <FilesIcon />;
+        }
+        return <UploadIcon />;
+    };
```
yvmunayev, line 172: you can put this in a helpers

src/components/FileSelector/index.js
```diff
@@ -0,0 +1,320 @@
...
+            return placeholder;
+        }
+        if (files.length === 0) {
+            return placeholder;
+        }
+        if (files.length === 1) {
+            return files[0].name;
+        }
+        return `${files.length} files`;
+    };
```
yvmunayev, line 185: same before

src/components/FileSelector/index.js
```diff
@@ -0,0 +1,320 @@
...
+                type="file"
+                multiple={multiple}
+                disabled={disabled}
+                onChange={handleChange}
+                tabIndex={tabIndex}
+                accept={accept}
+                onFocus={handleFocus}
+                onBlur={handleBlur}
+                ref={inputRef}
+            />
```
yvmunayev, line 254: you can use the HiddenElement look at CheckboxToggle

src/components/FileSelector/index.js
```diff
@@ -0,0 +1,320 @@
...
+        focus: () => {
+            inputRef.current.focus();
+        },
+        click: () => {
+            inputRef.current.click();
+        },
+        blur: () => {
+            inputRef.current.blur();
+        },
+    }));
```
yvmunayev, line 64: I think you don't need to define "inputRef", that you can the "ref" that comes by parameters

HellWolf93: I need to define inputRef because even when the developer does not pass a ref I need to send click events to the input and set the files

src/components/FileSelector/index.d.ts
```diff
@@ -0,0 +1,23 @@
...
+    bottomHelpText: ReactNode;
+    placeholder: string;
+    tabIndex: string | number;
+    required: boolean;
+    multiple: boolean;
+    disabled: boolean;
+    variant: string;
+    hideLabel: boolean;
+    accept: string;
+    onChange: (value: Array<File>) => void;
```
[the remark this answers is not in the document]
HellWolf93: File refers to base File interface defined in typescript lib, I think that as a predefined type we don't have to import it

src/components/FileSelector/__test__/fileSelector.spec.js
```diff
@@ -0,0 +1,82 @@
...
+            readOnly: false,
+            hideLabel: false,
+            inputId: expect.any(String),
+            id: expect.any(String),
+        });
+    });
+
+    it('should render HelpText when bottomHelpText is passed', () => {
+        const component = mount(<FileSelector bottomHelpText="help text" />);
+        expect(component.find(HelpText).length).toBe(1);
```
LeandroTorresSicilia, line 56: here you can:
```expect(component.find(HelpText).exists).toBe(true);```

File: src/components/FileSelector/index.d.ts [hunk 8]
```diff
@@ -0,0 +1,23 @@
...
+    onFocus: (event: FocusEvent<HTMLInputElement>) => void;
+    onBlur: (event: FocusEvent<HTMLInputElement>) => void;
```
LeandroTorresSicilia, line 20: all props must be optional

src/components/FileSelector/index.d.ts
```diff
@@ -0,0 +1,23 @@
...
+    name: string;
+    label: ReactNode;
+    error: ReactNode;
+    bottomHelpText: ReactNode;
+    placeholder: string;
+    tabIndex: string | number;
+    required: boolean;
+    multiple: boolean;
+    disabled: boolean;
+    variant: string;
```
LeandroTorresSicilia, line 15: we should be more explicit here:
```variant?: 'inline' | 'normal'```

File: src/components/FileSelector/index.js [hunk 10]
```diff
@@ -0,0 +1,302 @@
...
+    tabIndex: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
+    /** Specifies that an input field must be filled out before submitting the form.
+     * This value defaults to false. */
+    required: PropTypes.bool,
+    /** Specifies that multiple files can be picked. */
+    multiple: PropTypes.bool,
+    /** Specifies that an input element should be disabled. This value defaults to false. */
+    disabled: PropTypes.bool,
+    /** Specifies that the variant of the file selector. */
+    variant: PropTypes.oneOf(['inline', 'normal']),
```
LeandroTorresSicilia, line 268: let's use ```['inline', 'multiline']```

src/components/FileSelector/index.js
yvmunayev: @HellWolf93 You can try using the "ref" that comes by parameters, I am working on a component that is the same situation and it works for me with ref. It seems that forwardRef always sends a reference

src/components/FileSelector/index.d.ts
```diff
@@ -0,0 +1,23 @@
...
+    bottomHelpText?: ReactNode;
+    placeholder?: string;
+    tabIndex?: string | number;
+    required?: boolean;
+    multiple?: boolean;
+    disabled?: boolean;
+    variant?: 'inline' | 'multiline';
+    hideLabel?: boolean;
+    accept?: string;
+    onChange?: (value: Array<File>) => void;
```
LeandroTorresSicilia, line 18:
```suggestion
    onChange?: (value: FileList) => void;
```

src/components/FileSelector/index.js
```diff
@@ -0,0 +1,254 @@
...
+            inputRef.current.click();
+        },
+        blur: () => {
+            inputRef.current.blur();
+        },
+    }));
+
+    useEffect(() => {
+        inputRef.current.files = files;
+    }, [files]);
```
LeandroTorresSicilia, line 68: why do we need this?

PR #1524, opened 2020-05-04
feat: implement CodeInput component

blakmetall:
feat: implement codeInput component first revision


src/components/CodeInput/index.d.ts
```diff
@@ -0,0 +1,26 @@
...
+
+type Value = string | number;
+
+export interface CodeInputProps extends BaseProps {
+    id?: string;
+    name?: string;
+    value?: Value;
+    label?: ReactNode;
+    bottomHelpText?: ReactNode;
+    length: number;
```
yvmunayev, line 12: length is not required

src/components/CodeInput/index.js
```diff
@@ -0,0 +1,141 @@
...
+
+    return (
+        <StyledFieldset className={className} name={fieldsetName} style={style} id={inputId}>
+            <Label
+                label={label}
+                hideLabel={!label}
+                required={required}
+                readOnly={readOnly}
+                inputId={inputId}
+            />
```
yvmunayev, line 54: You have to check the inputId is putting it to the fieldset, you should put it to the first input and fieldset id is the id of the props.

src/components/CodeInput/inputItems.js
```diff
@@ -0,0 +1,142 @@
...
+                maxLength="1"
+            />
+        );
+    });
+
+    return <div>{inputs}</div>;
+});
+
+InputItems.propTypes = {
+    value: PropTypes.string,
```
yvmunayev, line 113: value: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),

blakmetall:
> I think that I should rethink your life cycle because it is very complex and makes you work harder. I think the first step should be to normalize "value" in an array and that array is core of your component. This is going to give you a few validations and checks now.
> 
> Try to get the functions out for the helpers and optimize using hook.

sure, thank you for your comments, applying all changes right away.

blakmetall:
Changes ready for a second revision...

About the length prop you said we don't, it is used to stablish the codeLength to be used for the component...

This is the file with the props definitions I have

https://docs.google.com/document/d/1PAFwTznCcUdz3_Vp9NJ5o0EFm3N4qcRY2GWQ1IXGl7Q/edit

-----

I changed the name from length to codeLength to be more explicit, if you think its best to be called different please let me know.

src/components/CodeInput/helpers/getLoopArray.js
```diff
@@ -0,0 +1,11 @@
...
+    const loopArray = [];
+
+    if (length > 0) {
+        for (let i = 0; i < length; i += 1) {
+            loopArray.push('');
+        }
+    }
+
+    return loopArray;
+}
```
LeandroTorresSicilia, line 11: remove this function

src/components/CodeInput/inputItems.js
```diff
@@ -0,0 +1,139 @@
...
+
+    const handleOnPaste = e => {
+        const clipboard = getValidValue(e.clipboardData.getData('Text'));
+        const copiedLength = clipboard.split('').length;
+        const newIndex = copiedLength < codeLength ? copiedLength : copiedLength - 1;
+        onChange(clipboard, 0);
+        setActiveIndex(newIndex);
+    };
+
+    const inputs = getLoopArray(codeLength).map((v, index) => {
```
LeandroTorresSicilia, line 78:
```
Array(length).fill('').map()
```

src/components/CodeInput/index.js
```diff
@@ -0,0 +1,142 @@
...
+        error,
+        tabIndex,
+        onClick,
+        onChange,
+        onFocus,
+        onBlur,
+        onKeyDown,
+        className,
+        style,
+    } = props;
```
LeandroTorresSicilia, line 36: use useReduxForm

src/components/CodeInput/helpers/getNormalizedValue.js
```diff
@@ -0,0 +1,10 @@
+export default function getNormalizedValue(inputValue, inputIndex, value, codeLength) {
+    const valueArray = value.split('');
+    for (let i = 0; i < codeLength; i += 1) {
+        if (i === inputIndex) {
+            valueArray[i] = inputValue;
+        }
+    }
```
LeandroTorresSicilia, line 7: let's try to avoid use for loops directly

src/components/CodeInput/helpers/getValidValue.js
```diff
@@ -0,0 +1,11 @@
+export default function getValidValue(value, codeLength) {
+    if (value) {
+        const extractedNumbers = value.match(/\d+/g);
+        const numbers = [];
+        for (let i = 0; i <= extractedNumbers.length; i += 1) {
+            numbers.push(extractedNumbers[i]);
+        }
```
LeandroTorresSicilia, line 7: same here not use of for

src/components/CodeInput/index.js
```diff
@@ -0,0 +1,142 @@
...
+
+    const valueProp = getValidValue(value, codeLength);
```
LeandroTorresSicilia, line 38: use a hook useNormalizedValue and inside the hook use useMemo

src/components/CodeInput/readme.md
```diff
@@ -0,0 +1,79 @@
...
+            </div>
+        </Card>
+    );
+};
+
+<div className="rainbow-align-content_center rainbow-m-around_xx-large">
+    <CodeInputBase  />
+</div>
+```
+ No newline at end of file
```
TahimiLeonBravo, line 77: remove this text

src/components/CodeInput/index.js
```diff
@@ -0,0 +1,142 @@
...
+
+    return (
+        <StyledFieldset className={className} name={name} style={style} id={id}>
+            <Label
+                label={label}
+                hideLabel={!label}
+                required={required}
+                readOnly={readOnly}
+                inputId={inputId}
+            />
```
LeandroTorresSicilia, line 54: since this is a group of inputs wrapped by a fieldset then here we need to use a legend tag instead of label then we can't reuse the Label component

src/components/CodeInput/inputItems.js
```diff
@@ -0,0 +1,139 @@
...
+        onKeyDown,
+    } = useReduxForm(props);
+
+    const arrayValue = value.split('');
+    const defaultActiveIndex = value ? arrayValue.length - 1 : 0;
+    const [activeIndex, setActiveIndex] = useState(defaultActiveIndex);
+
+    useEffect(() => {
+        setFocus(ref);
+    });
```
LeandroTorresSicilia, line 36: this will cause the first input to get the focus and we need to defer this to the user, the consumer is responsible for focus the input whenever he wants

src/components/CodeInput/inputItems.js
```diff
@@ -0,0 +1,139 @@
...
+
+    useEffect(() => {
+        setFocus(ref);
+    }, [ref, activeIndex]);
```
LeandroTorresSicilia, line 40: same ^^

blakmetall:
> I believe that the life cycle of this component should be this:
> 
> a variable values ​​that is an array that is calculated based on the prop value and lenght, with that array you have the data to paint in complete component. assume this "1234" = [1, 2, 3, 4], "1 34" = [1, "", 3, 4], "12" = [1, 2, "", ""] = and " "= [" "," "," "," "] put this in a custom hook and use useMemo
> 
> a focusedIndex state that controls which input the focus is on
> 
> handleChange (newValue, index) converts the array to string and calls onChange
> 
> make a custom hook that has all the logic of keyPress
> 
> To optimize the repainting of nodes you can use a context where the fucusedIndex state and the handleChange are passed.
> 
> in the render you make a values.map for InputItem
> 
> the InputItem component is simple
> 
> props value, index and get from context fucusedIndex and handleChange
> 
> a useEffect to put focus
> 
> use keyDownPress to only enter numbers
> 
> and the onChange of the input calls handleChange (value, index)
> 
> I think you can delete the InputItems, all that logic can be in index.js
> 
> try to optimize everything you can using useMeme and useCallback.

thank you for all your comments, they totally changed the way I was seeing the component...

src/components/CodeInput/index.d.ts
```diff
@@ -0,0 +1,25 @@
...
```
LeandroTorresSicilia, line 12: remember as @yvmunayev said length must be optional:
```length?: number;```

Comment (LeandroTorresSicilia) [inline src/components/CodeInput/index.d.ts:12]:
```suggestion
    length?: number;
```

src/components/CodeInput/index.d.ts
```diff
@@ -0,0 +1,25 @@
+import { ReactNode, MouseEvent, FocusEvent, KeyboardEvent } from 'react';
+import { BaseProps } from '../types';
+
+type Value = string | number;
+
+export interface CodeInputProps extends BaseProps {
+    id?: string;
+    name?: string;
+    value?: Value;
```
LeandroTorresSicilia, line 9: I think value should be only string

src/components/CodeInput/index.d.ts
```diff
@@ -0,0 +1,25 @@
...
+    label?: ReactNode;
+    bottomHelpText?: ReactNode;
+    length?: number;
+    disabled?: boolean;
+    required?: boolean;
+    readOnly?: boolean;
+    error?: ReactNode;
+    tabIndex?: number | string;
+    onClick?: (event: MouseEvent<HTMLInputElement>) => void;
+    onChange?: (value: Value) => void;
```
LeandroTorresSicilia, line 19: same here we only will send string

src/components/CodeInput/inputItem.js
```diff
@@ -0,0 +1,94 @@
...
+        <StyledInput
+            key={index}
+            value={value}
+            disabled={disabled}
+            required={required}
+            readOnly={readOnly}
+            error={error}
+            tabIndex={tabIndex}
+            onClick={e => onClick(e, index)}
+            onChange={e => handleOnChange(e, index)}
```
LeandroTorresSicilia, line 46: here you can direclty use:
```onChange={e => onChange(e.target.value, index)}```

File: src/components/CodeInput/inputItem.js [hunk 18]
```diff
@@ -0,0 +1,94 @@
...
+        id,
+        isActive,
+    } = props;
+
+    const handleOnChange = (e, inputIndex) => {
+        onChange(e.target.value, inputIndex);
+    };
+
+    const handleOnKeyDown = (e, inputIndex) => {
+        const shouldMovePrev = !value && e.key === 'Backspace';
```
LeandroTorresSicilia, line 29: key must be checked by key code and they are use in one place: `src/libs/constants.js`

src/components/CodeInput/inputItem.js
```diff
@@ -0,0 +1,94 @@
...
+            value={value}
+            disabled={disabled}
+            required={required}
+            readOnly={readOnly}
+            error={error}
+            tabIndex={tabIndex}
+            onClick={e => onClick(e, index)}
+            onChange={e => handleOnChange(e, index)}
+            onFocus={e => onFocus(e, index)}
+            onBlur={e => onBlur(e, index)}
```
LeandroTorresSicilia, line 48: we are sending event and index in the onBlur prop?
if so then fix this since onBlur should only send the native event

src/components/CodeInput/index.js
```diff
@@ -0,0 +1,160 @@
...
+    };
+
+    const handleOnPaste = e => {
+        onChange(getNumbersFromClipboard(e.clipboardData.getData('Text')));
+    };
+
+    return (
+        <StyledFieldset className={className} name={name} style={style} id={id}>
+            <RenderIf isTrue={!!label}>
+                <StyledLabel htmlFor={inputId}>
```
LeandroTorresSicilia, line 68: In this case, the input id does not get associated with the label because it is a legend element and not a label.

src/components/CodeInput/readme.md
```diff
@@ -0,0 +1,95 @@
...
+    useEffect(() => {
+        codeInputRef.current.focus();
+    }, []);
+
+    return (
+        <CodeInput
+            id="codeinput-1"
+            value={code}
+            label="Code"
+            required={true}
```
LeandroTorresSicilia, line 20: simply use ```required```

src/components/CodeInput/readme.md
```diff
@@ -0,0 +1,95 @@
...
+            onChange={(code) => { setCode(code); }}
```
LeandroTorresSicilia, line 21: you can ```onChange={setCode}```

src/components/CodeInput/readme.md
```diff
@@ -0,0 +1,95 @@
...
+const cardStyle = { padding: '40px' };
+const codeInputStyle = { marginBottom: '41px' };
+
+function CodeInputBase() {
+    const [code, setCode] = useState();
+    const [error, setError] = useState();
+    const codeInputRef = useRef();
+
+    const handleOnSubmit = () => {
+        const isValid = code.split('').length === 6;
```
LeandroTorresSicilia, line 61: I think you can check a length directly in a string

src/components/CodeInput/readme.md
```diff
@@ -0,0 +1,95 @@
...
+        <Card style={cardStyle}>
+            <CodeInput
+                id="codeinput-3"
+                value={code}
+                label={<MyLabel />}
+                length={6}
+                onChange={code => setCode(code)}
+                style={codeInputStyle}
+                error={error}
+                ref={codeInputRef}
```
LeandroTorresSicilia, line 79: there is no need to pass a ref since it is not used

src/components/CodeInput/inputItem.js
blakmetall: its ok, I added the index as an extra for the developer if they need the index of the input, might not be necessary at all for them, I'll removed those values.

src/components/CodeInput/readme.md
blakmetall: :O, I was only using it for the onClick, onFocus, and onEvent stuff... I see now

src/components/CodeInput/readme.md
blakmetall: definitely, that was a quick solution just for the readme, I'll check the one and replace it

src/components/CodeInput/inputItem.js
blakmetall: some of those onEvent are needed at the top of the component, inside we are calling the native event only, but we need in special cases the index...

onClick and onBlur are the only ones that can't be called directly

`
onClick={onClick}
onChange={e => handleOnChange(e.target.value, index)} 
onFocus={e => onFocus(e, index)}
onBlur={onBlur}
onKeyDown={e => handleOnKeyDown(e, index)}
`

I am going to change for now the onClick and onBlur, if we need to change the other ones differently please let me know.

src/components/CodeInput/inputItems.js
```diff
@@ -0,0 +1,92 @@
...
+        onFocus,
+        onBlur,
+        onKeyDown,
+        onPaste,
+        focusedIndex,
+        id,
+    } = props;
+
+    const inputs = value.map((inputValue, index) => {
+        const inputIndex = index;
```
LeandroTorresSicilia, line 24: ```const inputIndex = `code-input-${index}`;```

src/components/CodeInput/inputItem.js
```diff
@@ -0,0 +1,95 @@
...
+        const shouldMovePrev = !value && e.keyCode === DELETE_KEY;
+        if (shouldMovePrev) {
+            onChange('', inputIndex - 1);
+        }
+        onKeyDown(e);
+    };
+
+    return (
+        <StyledInput
+            key={index}
```
LeandroTorresSicilia, line 39: remove ```key={index}```

Issue #1519, opened 2020-05-03
feat: add prop size to VisualPicker component


TahimiLeonBravo:
#### Requirements
- [ ] it should have a be `size` prop with the values `small`, `medium` and `large`
- [ ] the default value should be `medium`

PR #1527, opened 2020-05-04
feat: add prop size to VisualPicker component

rgah2107:
fix: #1519

- add prop size to VisualPicker component

@nexxtway/react-rainbow

### DIFF
--- a/src/components/VisualPicker/index.d.ts
+++ b/src/components/VisualPicker/index.d.ts
@@ -1,5 +1,5 @@
 import { ReactNode, ComponentType } from 'react';
-import { BaseProps } from '../types';
+import { BaseProps, VisualPickerSize } from '../types';
 
 type Value = string[] | string;
 
@@ -13,7 +13,7 @@ export interface VisualPickerProps extends BaseProps {
     error?: ReactNode;
     children?: ReactNode;
     multiple?: boolean;
-    size?: string;
+    size?: VisualPickerSize;
 }
 
 declare const VisualPicker: React.ComponentType<VisualPickerProps>;
--- a/src/components/VisualPicker/index.d.ts
+++ b/src/components/VisualPicker/index.d.ts
@@ -13,6 +13,7 @@ export interface VisualPickerProps extends BaseProps {
     error?: ReactNode;
     children?: ReactNode;
     multiple?: boolean;
+    size?: string;
 }
 
 declare const VisualPicker: React.ComponentType<VisualPickerProps>;
--- a/src/components/VisualPickerOption/styled/content.js
+++ b/src/components/VisualPickerOption/styled/content.js
@@ -1,8 +1,9 @@
 import styled from 'styled-components';
 import attachThemeAttrs from '../../../styles/helpers/attachThemeAttrs';
 
+const sizeMap = { large: '210px', medium: '142px', small: '100px' };
 const StyledContent = attachThemeAttrs(styled.span)`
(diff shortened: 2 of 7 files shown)


src/components/VisualPicker/index.d.ts
```diff
@@ -13,6 +13,7 @@ export interface VisualPickerProps extends BaseProps {
     error?: ReactNode;
     children?: ReactNode;
     multiple?: boolean;
+    size?: string;
```
LeandroTorresSicilia, line 16: be more especific here like:
```size?: 'small' | 'medium' | 'large';```

File: src/components/VisualPickerOption/styled/content.js [hunk 2]
```diff
@@ -13,6 +13,21 @@ const StyledContent = attachThemeAttrs(styled.span)`
...
+    ${props =>
+        props.size === 'medium' &&
+        `
+            height: 142px;
+        `};
+    ${props =>
+        props.size === 'small' &&
+        `
+            height: 100px;
+        `};
```
LeandroTorresSicilia, line 30: here you can use a map like:
```
const sizeMap = {
    large: '210px',
    medium: '142px',
    small: '100px',
};
```
and then:
```height: ${props => sizeMap[props.size] || 142px}```

**Comment by TahimiLeonBravo** [comment]
@rgah2107 let's add 2 examples more with the new sizes on the component

File: src/components/VisualPickerOption/styled/content.js [hunk 3]
```diff
@@ -1,8 +1,9 @@
 import styled from 'styled-components';
 import attachThemeAttrs from '../../../styles/helpers/attachThemeAttrs';
 
+const sizeMap = { large: '210px', medium: '142px', small: '100px' };
 const StyledContent = attachThemeAttrs(styled.span)`
-    height: 142px;
+    height: ${props => sizeMap[props.size] || '142px'};
```
yvmunayev, line 6: size is always defined because in defaultProps you put medium

TahimiLeonBravo:
@rgah2107 we need to have the same width and the height on the different variants

Screen Shot 2020-05-05 at 9 32 47 PM

Screen Shot 2020-05-05 at 9 32 41 PM

src/components/VisualPickerOption/styled/content.js
LeandroTorresSicilia: @yvmunayev
I think we need do a fallback to `|| '142px'` since the user can pass a wrong size value and the component should render nicely

yvmunayev: check here, I'm wrong and if necessary put:

`height: ${props => sizeMap[props.size] || '142px'};`

src/components/VisualPickerOption/styled/content.js
```diff
@@ -1,9 +1,10 @@
 import styled from 'styled-components';
 import attachThemeAttrs from '../../../styles/helpers/attachThemeAttrs';
 
+const sizeMap = { large: '210px', medium: '142px', small: '100px' };
 const StyledContent = attachThemeAttrs(styled.span)`
-    height: 142px;
-    width: 100%;
+    height: ${props => sizeMap[props.size]};
+    width: ${props => sizeMap[props.size]};
```
LeandroTorresSicilia, line 7: here we need to do a fallback to a default size since the user can pass a wrong size where `sizeMap[props.size]` will return undefined

Issue #1467, opened 2020-04-08
feat: implement PhoneInput component


TahimiLeonBravo:
As a customer, I want to inter phone number

#### Requirements
- [x] Define API
- [ ] Implement component
- [ ] Add Test

##### Define API
https://docs.google.com/document/d/17K7Qnph7ZvbC4g_LBY0Te37U6enj-OGRwTNYHFcBEQ0/edit

##### Design
https://zpl.io/V4Bpjzy

PR #1533, opened 2020-05-08
feat: implement phone input component

yvmunayev:
fix: #1467

- feat: implement PhoneInput component
- fix: add useCallback to useOutsideClick

@nexxtway/react-rainbow


src/components/PhoneInput/styled/indicator.js
```diff
@@ -0,0 +1,22 @@
+import styled from 'styled-components';
+
+const StyledIndicator = styled.span`
+    z-index: 2;
```
HellWolf93, line 4: z-index has no effect in `static` positioned elements

LeandroTorresSicilia:
@yvmunayev There is a conflict with the `src/components/Picklist/menuArrowButton.js` since we removed this component from the Picklist, you should create a new one inside the PhoneInput

LeandroTorresSicilia:
When pass an empty array to the countries prop it gets broken, when fix it also add a test for this situation
Screen Shot 2020-06-01 at 9 38 52 AM

src/components/PhoneInput/styled/index.js
```diff
@@ -0,0 +1,20 @@
+import styled from 'styled-components';
+import attachThemeAttrs from '../../../styles/helpers/attachThemeAttrs';
+
+export const StyledCountryCode = attachThemeAttrs(styled.div)`
```
LeandroTorresSicilia, line 4: since the theme is not used in this styled then remove attachThemeAttrs

src/components/PhoneInput/hooks/useHandleFocus.js
```diff
@@ -0,0 +1,16 @@
...
+export default function useHandleFocus(focusIndex, onFocus, setFocusIndex) {
+    return useCallback(
+        (event, index) => {
+            if ((index === 0 && focusIndex === 1) || index === focusIndex) {
+                return;
+            }
+            if (focusIndex === -1) {
+                onFocus(event);
+            }
+            setFocusIndex(index);
```
LeandroTorresSicilia, line 12: leave this for now, but keep in mind that passing setState to external functions is an antipattern

src/components/PhoneInput/index.d.ts
```diff
@@ -0,0 +1,35 @@
...
+export interface PhoneInputProps extends BaseProps {
+    id?: string;
+    name?: string;
+    label?: ReactNode;
+    hideLabel?: boolean;
+    placeholder?: string;
+    icon?: ReactNode;
+    maxLength?: number;
+    minLength?: number;
+    pattern?: string;
```
LeandroTorresSicilia, line 19: in this case, the pattern prop does not make sense, also let's remove maxLength and minLength for now until a use case appears

src/components/PhoneInput/countriesDropdown/countriesList.js
```diff
@@ -0,0 +1,62 @@
...
+import PropTypes from 'prop-types';
+import RenderIf from '../../RenderIf';
+import Icon from '../../PicklistOption/icon';
+import StyledIconContainer from '../../PicklistOption/styled/iconContainer';
+import StyledCheckmarkIcon from '../../PicklistOption/styled/checkmarkIcon';
+import { StyledItem, StyledCountryCodeItem } from './styled';
+import { useSimulatedLoading } from './hooks';
+
+const CountriesList = memo(props => {
+    const { countries, country, itemsRef, handleCountryChange, handleActiveChange } = props;
```
LeandroTorresSicilia, line 11: normally handle is used to define the function that handles the callback, instead this should be called onCountryChange and onActiveChange, I think it is better simply call them onChange and onHover

src/components/PhoneInput/countriesDropdown/countriesList.js
```diff
@@ -0,0 +1,62 @@
...
+            <StyledItem
+                key={isoCode}
+                ref={el => {
+                    itemsRef.current[index] = el;
+                    return itemsRef.current[index];
+                }}
+                onClick={() => handleCountryChange(value)}
+                onMouseEnter={() => handleActiveChange(index)}
+                role="option"
+                aria-selected={false}
```
LeandroTorresSicilia, line 28: aria-selected will be always false?

src/components/PhoneInput/countriesDropdown/helpers/filterCountries.js
```diff
@@ -0,0 +1,39 @@
...
+                    return false;
+                }
+                return true;
+            }
+            return false;
+        }) || [];
+    if (isFilteredCountry) {
+        filteredCountries.unshift(country);
+    }
+    return filteredCountries;
```
LeandroTorresSicilia, line 38: take look here: `src/components/InternalDropdown/helpers/searchFilter.js`
this is a filter function that we use in other projects, maybe we can do similar and simplify this

src/components/PhoneInput/countriesDropdown/index.js
```diff
@@ -0,0 +1,136 @@
...
+    const handleCountryChange = useHandleCountryChange(scrollableRef, onCountryChange, setQuery);
+    const handleActiveChange = useKeyboardNavigation(
+        country,
+        countriesFiltered,
+        searchRef,
+        scrollableRef,
+        itemsRef,
+        handleCountryChange,
+        setFocusIndex,
+    );
```
LeandroTorresSicilia, line 49: for now leave this, but we should avoid pass to much arguments to a function, instead, when they are more than two we pass an object (only one argument with named parameters)

src/components/PhoneInput/countriesDropdown/styled/index.js
```diff
@@ -1,6 +1,6 @@
 import styled from 'styled-components';
 import attachThemeAttrs from '../../../../styles/helpers/attachThemeAttrs';
-import MenuArrowButton from '../../../Picklist/menuArrowButton';
+import CheckmarkIcon from '../../../Option/checkmark';
```
LeandroTorresSicilia, line 3: I think we should not get this icon from other component folders since changes to Options could break PhoneInput, in this case it is better to repeat the icon to be inside PhoneInput folder, we are working in a solution in order to not repeat icons any more

PR #1538, opened 2020-05-11
fix: codeInput implement specs, code improvement and new readme examples

blakmetall:
fix: codeInput accessibility, a few code improvements and readme examples added


src/components/CodeInput/index.js
```diff
@@ -82,6 +83,8 @@ const CodeInput = React.forwardRef((props, ref) => {
                     <RequiredAsterisk required={required} />
                     {label}
                 </StyledLabel>
+
+                <StyledHiddenLabel id={labelId}>{label}</StyledHiddenLabel>
```
LeandroTorresSicilia, line 87: Take a look here: https://www.w3.org/WAI/tutorials/forms/custom-controls/
In order to have the right a11y we should have an label for each input but since we don't have a clear understanding of how a11y should be implemented here we should not do this for now

src/components/CodeInput/inputItems.js
```diff
@@ -46,6 +48,7 @@ const InputItems = React.forwardRef((props, ref) => {
                 ref={inputRef}
                 id={inputId}
                 isActive={isActive}
+                ariaLabelledBy={ariaLabelledBy}
```
LeandroTorresSicilia, line 51: same ^^ let's not do this for now

src/components/CodeInput/inputItem.js
```diff
@@ -54,6 +51,8 @@ const InputItem = React.forwardRef((props, ref) => {
             maxLength="1"
             id={id}
             isActive={isActive}
+            aria-label="Enter a number"
+            aria-labelledby={ariaLabelledBy}
```
LeandroTorresSicilia, line 55: same ^^

src/components/CodeInput/index.js
blakmetall: sure, definitely, I applied changes from my a11y perspective, might be diffuse, I understand completely... however it was quite a new learning and good exercise.

src/components/CodeInput/inputItems.js
blakmetall: roger that

src/components/CodeInput/inputItem.js
blakmetall: roger that again haha

integration/specs/CodeInput/codeInput-1.spec.js
```diff
@@ -0,0 +1,109 @@
...
+    beforeAll(() => {
+        browser.url('/#!/CodeInput/1');
+    });
+    beforeEach(() => {
+        browser.refresh();
+        const component = $(CODEINPUT);
+        component.waitForExist();
+    });
+    /*
+    it('should have focus on the first input on load', () => {
```
blakmetall, line 15: fixed: uncommented code in next commit

integration/specs/CodeInput/codeInput-1.spec.js
```diff
@@ -0,0 +1,107 @@
...
+        const component = $(CODEINPUT);
+        component.waitForExist();
+    });
+    it('should have focus on the first input on load', () => {
+        const codeInput = new PageCodeInput(CODEINPUT);
+        expect(codeInput.getFocusedIndex()).toBe(0);
+    });
+    it('should change focus to the second input when the first input its filled with a number', () => {
+        const codeInput = new PageCodeInput(CODEINPUT);
+        browser.keys(['', '2']);
```
LeandroTorresSicilia, line 21: why we need to pass ```['', '2']```?

integration/specs/CodeInput/codeInput-1.spec.js
```diff
@@ -0,0 +1,107 @@
...
+        expect(codeInput.getFocusedIndex()).toBe(1);
+    });
+    it('should remove value from first input when first input is selected, has a value and "Backspace" key is pressed', () => {
+        const codeInput = new PageCodeInput(CODEINPUT);
+        browser.keys(['', '2']);
+        expect(codeInput.getInputValueByIndex(0)).toBe('2');
```
LeandroTorresSicilia, line 27: ```getInputValueAtIndex```

integration/specs/CodeInput/codeInput-1.spec.js
```diff
@@ -0,0 +1,107 @@
...
+        browser.keys(DELETE_KEY);
+        expect(codeInput.getInputValueByIndex(0)).toBe('');
+    });
+    it('should return focus to the first input and clear his value when the first input has a value, the second input is empty, is focused and "Backspace" key is pressed', () => {
+        const codeInput = new PageCodeInput(CODEINPUT);
+        browser.keys(['', '2']);
+        expect(codeInput.getInputValueByIndex(0)).toBe('2');
+        expect(codeInput.getFocusedIndex()).toBe(1);
+        browser.keys(DELETE_KEY);
+        expect(codeInput.getInputValueByIndex(0)).toBe('');
```
LeandroTorresSicilia, line 37: in the test name say `should return focus to the first input`, then we should add a expect here:
```expect(codeInput.getFocusedIndex()).toBe(0);```

File: integration/specs/CodeInput/codeInput-1.spec.js [hunk 8]
```diff
@@ -0,0 +1,107 @@
...
+        browser.keys(['', '2']);
+        browser.keys(['', '2']);
+        browser.keys(['', '2']);
+        expect(codeInput.getFocusedIndex()).toBe(3);
+    });
+    it('should keep focus in the current input when we type a letter', () => {
+        const codeInput = new PageCodeInput(CODEINPUT);
+        expect(codeInput.getFocusedIndex()).toBe(0);
+        browser.keys(['', 'A']);
+        expect(codeInput.getFocusedIndex()).toBe(0);
```
LeandroTorresSicilia, line 58: add another test like this that check for the input to be empty in order to prove that it does not accepts other values than numbers

integration/specs/CodeInput/codeInput-1.spec.js
```diff
@@ -0,0 +1,107 @@
...
+    });
+    it('should keep current input focus when unfocused inputs are clicked', () => {
+        const codeInput = new PageCodeInput(CODEINPUT);
+        expect(codeInput.getFocusedIndex()).toBe(0);
+        codeInput.clickInputByIndex(2);
```
LeandroTorresSicilia, line 63: clickInputAtIndex

integration/specs/CodeInput/codeInput-1.spec.js
```diff
@@ -0,0 +1,107 @@
...
+        browser.keys(['', '2']);
+        expect(codeInput.getFocusedIndex()).toBe(1);
+        codeInput.clickInputByIndex(0);
+        expect(codeInput.getFocusedIndex()).toBe(1);
+    });
+    it('should leave focus on all inputs when "Tab" key is pressed', () => {
+        const codeInput = new PageCodeInput(CODEINPUT);
+        expect(codeInput.getFocusedIndex()).toBe(0);
+        browser.keys(TAB_KEY);
+        expect(codeInput.getFocusedIndex()).toBe(-1);
```
LeandroTorresSicilia, line 77: let's use undefined to indicate that no input is focused

src/components/CodeInput/__test__/codeInput.spec.js
```diff
@@ -0,0 +1,159 @@
...
+    });
+    it('should call useValueState with the right value and length', () => {
+        useValueState.mockImplementation(() => ['1', '', '', '']);
+        mount(<CodeInput value="1" />);
+        expect(useValueState).toHaveBeenCalledWith('1', 4);
+    });
+    it('should call useFocusedIndexState with the right value and length', () => {
+        useFocusedIndexState.mockImplementation(() => 1);
+        mount(<CodeInput value="1" />);
+        expect(useValueState).toHaveBeenCalledWith('1', 4);
```
LeandroTorresSicilia, line 146: when testing with what was called a mocked function you need to call mockReset before since in the last test useValueState  was called with this same value then this is not testing really that now it was called with ```'1', 4```

integration/specs/CodeInput/codeInput-1.spec.js
blakmetall: My bad: I though it was necessary the first value, it refers to Metadata, I changed to key only and seems to be working...

- - - - -

Reverted to browsers.keys('2'); all of them in next PR update

-

src/components/CodeInput/__test__/inputItems.spec.js
```diff
@@ -0,0 +1,74 @@
...
+                    .prop('value'),
+            ).toBe(value);
+        });
+    });
+    it('should have unique ids on each input identified by base id and input index', () => {
+        const values = ['', '', '', ''];
+        const id = 'code-input-base';
+        const component = mount(<InputItems value={values} id={id} />);
+        values.forEach((value, index) => {
+            const singleId = `${id}-${index}`;
```
LeandroTorresSicilia, line 49: we should construct this id using our uniqueId hook instead of use the index since it will assure that the id will be unique, remember that id can't be twice in the entire web page

src/components/CodeInput/__test__/inputItems.spec.js
```diff
@@ -0,0 +1,74 @@
...
+        const input0 = component
+            .find(InputItem)
+            .at(0)
+            .getDOMNode();
+        const input1 = component
+            .find(InputItem)
+            .at(1)
+            .getDOMNode();
+        expect(input0).not.toBe(ref.current);
+        expect(input1).toBe(ref.current);
```
LeandroTorresSicilia, line 72: to compare object values you must use toEqual

integration/specs/CodeInput/codeInput-1.spec.js
blakmetall: I have 3 methods returning -1, 

should I apply undefined return value on all of them or just for gestFocusedIndex ?

integration/specs/CodeInput/codeInput-1.spec.js
blakmetall: definitely... that was the main test for that block

src/components/CodeInput/hooks/__test__/usePreviousIndex.spec.js
```diff
@@ -0,0 +1,17 @@
...
+        const { result } = renderHook(() => usePreviousIndex());
+        expect(result.current).toBe(undefined);
+    });
+    it('should return 1 when focusedIndex is 1', () => {
+        const { result, rerender } = renderHook(initialValue => usePreviousIndex(initialValue));
+        expect(result.current).toBe(undefined);
+        rerender(2);
+        setTimeout(() => {
+            expect(result.current).toEqual(2);
+        }, 1000);
```
LeandroTorresSicilia, line 15: why use setTimeout? if rerender is async then use async await in order to wait for the result

src/components/CodeInput/index.js
```diff
@@ -113,8 +116,6 @@ const CodeInput = React.forwardRef((props, ref) => {
 CodeInput.propTypes = {
     /** The id of the outer element. */
     id: PropTypes.string,
-    /** An identifier for the CodeInput element. */
-    name: PropTypes.string,
```
LeandroTorresSicilia, line 117: remember to remove name prop in index.d.ts

src/components/CodeInput/inputItems.js
```diff
@@ -20,11 +21,19 @@ const InputItems = React.forwardRef((props, ref) => {
...
+            onFocus(e);
+        }
+    };
+
     const inputs = value.map((inputValue, index) => {
         const inputIndex = index;
         const isActive = inputIndex === focusedIndex;
         const inputRef = isActive ? ref : null;
-        const inputId = isActive ? id : null;
+        const inputId = `${id}-${index}`;
```
LeandroTorresSicilia, line 36: as said before input id should be generated with uniqueId, but since we will not associate it with no label then it is not needed then, for now, let's not pass id to the input

src/components/CodeInput/inputItems.js
```diff
@@ -20,11 +21,19 @@ const InputItems = React.forwardRef((props, ref) => {
         id,
     } = props;
 
+    const handleOnFocus = e => {
+        if (readOnly || disabled) {
+            ref.current.blur();
```
LeandroTorresSicilia, line 26: check this, because I think that when you pass disabled or readonly to the input itself then it does not get the focus so you don't need to call blur programatically

src/components/CodeInput/__test__/inputItems.spec.js
blakmetall: For this case, we are using the uniqueID hook in the parent

`const inputId = useUniqueIdentifier('code-input');`

Then it is passed to InputItems which renders unique ids for inputs based on parentID and index, setting there a unique id for each input.

For my that solution work, if not for learning purposes if possible can you share why?

--

src/components/CodeInput/pageObject/index.js
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    getFocusedInput() {
+        const inputs = $(this.rootElement).$$('input');
+        return inputs.filter(input => input.isFocused()).shift();
+    }
+
+    getInputByIndex(inputIndex) {
+        const inputs = $(this.rootElement).$$('input');
+        return inputs.filter((elem, index) => index === inputIndex).shift();
```
LeandroTorresSicilia, line 83: I think you can take access the input by index directly, look at other implementations, thus we avoid the filter

src/components/CodeInput/styled/index.js
```diff
@@ -83,6 +89,11 @@ const StyledInput = attachThemeAttrs(styled.input)`
     `}
 `;
 
+const StyledInputItems = attachThemeAttrs(styled.div)`
+    position: relative;
+    margin-bottom: -12px;
```
LeandroTorresSicilia, line 94: @TahimiLeonBravo check this negative margin

integration/specs/CodeInput/codeInput-1.spec.js
```diff
@@ -0,0 +1,120 @@
+const PageCodeInput = require('../../../src/components/CodeInput/pageObject');
+const { DELETE_KEY, TAB_KEY } = require('../../constants');
+
+const CODEINPUT = '#codeinput-1';
+
+const awaitForUpdate = () => {
+    browser.pause(99);
+};
```
LeandroTorresSicilia, line 8: we should avoid use pause, it is not deterministic

src/components/CodeInput/__test__/inputItems.spec.js
blakmetall: test removed for now because we are removing id's over the inputs for now; until we fully integrate accesibility.

src/components/CodeInput/styled/index.js
blakmetall: Screen Shot 2020-05-14 at 16 56 14

that was because in a previous commit I added to inputs a 12px margin-bottom to prepare for responsiveness,  so I added to the container a negative margin bottom to bring back moved the position.

Screen Shot 2020-05-14 at 16 57 47

src/components/CodeInput/inputItems.js
blakmetall: I believe that when I did this test my used example was wrong, that make me think I should disable programmatically... this is misterious hahaha...

Now I confirmed it is not needed

integration/specs/CodeInput/codeInput-1.spec.js
```diff
@@ -80,8 +72,7 @@ describe('CodeInput base example', () => {
     });
     it('should keep focus on the second input when any input is clicked and the first input has a number already', () => {
         const codeInput = new PageCodeInput(CODEINPUT);
-        browser.keys(['', '2']);
-        awaitForUpdate();
+        browser.keys('2');
         expect(codeInput.getFocusedIndex()).toBe(1);
         codeInput.clickInputByIndex(0);
```
LeandroTorresSicilia, line 77: clickInputByIndex  => clickInputAtIndex

TahimiLeonBravo:
@blakmetall  I added a design for readOnly appearance, let's follow these styles
https://zpl.io/bPjBNym
Also, add the value to the example when it is readOnly

src/components/CodeInput/readme.md
```diff
@@ -88,6 +88,109 @@ function CodeInputBase() {
 };
 
 <div className="rainbow-align-content_center rainbow-m-around_xx-large">
-    <CodeInputBase  />
+    <CodeInputWithLength  />
+</div>
+```
+
+##### CodeInput events:
```
TahimiLeonBravo, line 95: I think we don't need this example

src/components/CodeInput/styled/index.js
TahimiLeonBravo: @blakmetall @LeandroTorresSicilia I think we should not add this negative margin since this a bad practice if we want to separate the bottomHelpText or the errorText from the input we should remove the `margin-top`

src/components/CodeInput/inputItems.js
```diff
@@ -44,18 +43,16 @@ const InputItems = React.forwardRef((props, ref) => {
...
                 onPaste={onPaste}
                 ref={inputRef}
-                id={inputId}
                 isActive={isActive}
             />
         );
     });
 
-    return <div>{inputs}</div>;
+    return <StyledInputItems>{inputs}</StyledInputItems>;
```
TahimiLeonBravo, line 51: I think we should not need this wrapper here, instead, we should use a `Fragment`

TahimiLeonBravo:
@blakmetall we need fix some tests here in order to merge this PR

src/components/CodeInput/pageObject/index.js
```diff
@@ -12,6 +12,13 @@ class PageCodeInput {
         this.rootElement = rootElement;
     }
 
+    typeKey(key) {
```
LeandroTorresSicilia, line 15: let's use type

blakmetall:
- [x] type method from pageObject updated
- [x] spec removed for readOnly verification (not needed anymore)
- [x] readOnly styled configuration update
- [x] update over onClick event
- [x] removal of unnecessary condition related to readOnly (not needed anymore)

blakmetall:
Note: integration tests failed over another Component unrelated tu CodeInput

Screen Shot 2020-05-21 at 14 10 28

Issue #1542, opened 2020-05-13
feat: add the `variant` prop to the `Picklist` component


TahimiLeonBravo:
#### Requirements
- [ ] It should have a `variant` prop 
 ```variant: PropTypes.oneOf(['default', 'shaded']),```
- [ ] It should have as defaultProp the variant `default`
- [ ] The variant shaded should looks like the last example on the design
https://zpl.io/agdABDD
- [ ] The variant default should looks like as the current design

Let's add a variant prop to the GlobalHeader component

PR #1549, opened 2020-05-14
feat: add the variant prop to the Picklist component

rgah2107:
fix: #1542

- add the `variant` prop to the `Picklist` component

@nexxtway/react-rainbow


TahimiLeonBravo:
@rgah2107 remember we need to add the variant prop to the Picklist

TahimiLeonBravo:
@rgah2107 I think we should redefine the variant values on the GlobalHeader component because the appearance of this was designed as shaded initially and now we can add a variant that is not shaded, maybe we should name it `neutral` or `flat` ...

Issue #1553, opened 2020-05-16
feat: add the `selectionType` prop to the Calendar component


TahimiLeonBravo:
#### Requirements
- [x] It should have a `selectionType` prop with the `'single' | 'range'` values
- [x] It should have by default the `single` value
- [x] it should send `onChange value = [Date, Date]` if `selectionStyle === 'range'`
- [x] it should sendon the `onChange`
`[Date, undefined] | [Date, Date] | []`

#### Design
https://zpl.io/a7vLJZK
example 5
We should follow the selection behavior like this
https://www.airbnb.com/

PR #1563, opened 2020-05-19
feat: add selectionType prop to Calendar component

wildergd:
fix: #1553 

- add selectionType prop to Calendar component

@nexxtway/react-rainbow


src/components/Calendar/hooks/useDaysBuilder.js
```diff
@@ -0,0 +1,55 @@
+import React, { useMemo } from 'react';
+import { addDays, shouldDateBeSelected, isDateWithinRange, isSameDay } from '../helpers';
+
+export default function useDaysBuilder(Component, props) {
```
yvmunayev, line 4: you don't need to pass Component, it's better to import Day here, just pass all the props to it. We are also trying to make everything functional. You can change something like this:

`
function useDaysBuilder(
    startDate,
    endDate,
    firstDayMonth,
    lastDayMonth,
    minDate,
    maxDate,
    selectionType,
    value,
    selectedRange,
    currentRange,
    onChange,
    ) {
    return useMemo(() => {

        const daysDiff = endDate.getDate() - startDate.getDate();
        return Array.from(Array(daysDiff), (_, index) => {
            const date = addDays(startDate, index);
            return (
                <Day
                    date={date}
                    firstDayMonth={firstDayMonth}
                    key={date.getTime()}
                    minDate={minDate}
                    maxDate={maxDate}
                    onChange={onChange}
                    isSelected={shouldDateBeSelected(date, value, selectionType, selectedRange)}
                    isWithinRange={isDateWithinRange(date, currentRange)}
                    isFirstDayOfWeek={index === 0 || isSameDay(date, firstDayMonth)}
                    isLastDayOfWeek={index === 6 || isSameDay(date, lastDayMonth)}
                />
            );
        });
    }, [
        currentRange,
        endDate,
        firstDayMonth,
        lastDayMonth,
        maxDate,
        minDate,
        onChange,
        selectedRange,
        selectionType,
        startDate,
        value,
    ]);
}
`

src/components/Calendar/day.js
```diff
@@ -5,17 +5,42 @@ import { Consumer } from './context';
...
     const isDisabled = compareDates(date, maxDate) > 0 || compareDates(date, minDate) < 0;
     const tabIndex = isSameDay(focusedDate, date) ? 0 : -1;
     const buttonRef = useRef();
 
+    const isRangeStartDate = !isEmptyRange(currentRange) ? isSameDay(date, currentRange[0]) : false;
+    const isRangeEndDate =
+        !isEmptyRange(currentRange) && currentRange.length >= 2
+            ? isSameDay(date, currentRange[1])
+            : false;
+
```
yvmunayev, line 43: Custom hook and useMemo for this

src/components/Calendar/hooks/useCurrentDateFromValue.js
```diff
@@ -0,0 +1,9 @@
+import { useMemo } from 'react';
+
+export default function useCurrentDateFromValue(value) {
+    return useMemo(() => {
+        if (!Array.isArray(value)) return value;
+        const [rangeStart, rangeEnd] = value;
+        return rangeEnd || rangeStart;
```
yvmunayev, line 7: the value must be normalized here, to avoid wrong values

src/components/Calendar/hooks/useRangeFromValue.js
```diff
@@ -0,0 +1,9 @@
+import { useMemo } from 'react';
+
+export default function useRangeFromValue(value, selectionType) {
+    return useMemo(() => {
+        if (selectionType === 'single') return undefined;
+        if (!!value && !Array.isArray(value)) return [value];
+        return value;
```
yvmunayev, line 7: the value must be normalized here, to avoid wrong values

src/components/Calendar/index.js
```diff
@@ -1,357 +1,49 @@
 /* eslint-disable react/prop-types */
```
yvmunayev, line 1: the PropTypes is defined

src/components/Calendar/week.js
```diff
@@ -1,41 +1,41 @@
...
+        startDate,
+        endDate,
+        minDate,
+        maxDate,
+        firstDayMonth,
+        lastDayMonth,
+        onChange,
+    } = props;
+    const { currentRange, selectionType, selectedRange } = useContext(CalendarContext);
+    const Days = useDaysBuilder(Day, {
```
yvmunayev, line 19: You don't need to pass the Day, just props

src/components/Calendar/index.js
```diff
@@ -1,357 +1,49 @@
...
+import { useLocale } from '../../libs/hooks';
+import SingleCalendar from './singleCalendar';
+import { buildNewRangeFromValue } from './helpers';
+import { useCurrentDateFromValue, useRangeFromValue } from './hooks';
 
 /**
  * Calendar provide a simple way to select a single date.
  */
-export default function Calendar({ locale, ...rest }) {
+export default function Calendar({ locale, selectionType, value, onChange, ...rest }) {
```
yvmunayev, line 12: `export default function Calendar(props) {
const { locale, selectionType, value, onChange, ...rest } = props;
}`

src/components/Calendar/day.js
```diff
@@ -5,17 +5,42 @@ import { Consumer } from './context';
...
+        firstDayMonth,
+        isSelected,
+        minDate,
+        maxDate,
+        onChange,
+        isWithinRange,
+        isFirstDayOfWeek,
+        isLastDayOfWeek,
+    } = props;
+    const {
```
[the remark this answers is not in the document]
wildergd: this is to split component props from context props... it can be one sentence

src/components/Calendar/hooks/useCurrentDateFromValue.js
wildergd: no need to normalize... value can be a date or an array... if it is a date it is passed to SingleCalendar component ... if it is an array, it takes one of the items and pass it... and SingleCalendar is the one that normalize the value

src/components/Calendar/helpers/buildNewRangeFromValue.js
```diff
@@ -0,0 +1,27 @@
+import isDateWithinRange from './isDateWithinRange';
+import compareDates from './compareDates';
+
+let rangeIndex = 0;
```
LeandroTorresSicilia, line 4: we should avoid this kind of side effects and instead embrace functional programming best practice

src/components/Calendar/hooks/useDaysBuilder.js
```diff
@@ -0,0 +1,50 @@
...
+                    firstDayMonth={firstDayMonth}
+                    key={date.getTime()}
+                    minDate={minDate}
+                    maxDate={maxDate}
+                    onChange={onChange}
+                    isSelected={shouldDateBeSelected(date, value, selectionType, selectedRange)}
+                    isWithinRange={isDateWithinRange(date, currentRange)}
+                    isFirstDayOfWeek={index === 0 || isSameDay(date, firstDayMonth)}
+                    isLastDayOfWeek={index === 6 || isSameDay(date, lastDayMonth)}
+                />
```
LeandroTorresSicilia, line 34: I think this should be a component since hooks are meant to deal with data and not visual logic

src/components/Calendar/hooks/useRangeStartDate.js
```diff
@@ -0,0 +1,9 @@
+import { useMemo } from 'react';
+import { isSameDay, isEmptyRange } from '../helpers';
+
+export default function useRangeStartDate(date, currentRange) {
+    return useMemo(() => (!isEmptyRange(currentRange) ? isSameDay(date, currentRange[0]) : false), [
```
LeandroTorresSicilia, line 5: to improve readability we need to avoid negate conditions instead we could:
```isEmptyRange(currentRange) ? false : isSameDay(date, currentRange[0])```

File: src/components/Calendar/hooks/useWeeksBuilder.js [hunk 12]
```diff
@@ -0,0 +1,34 @@
...
+        const week = addDays(date, -dayOfWeek);
+
+        return Array.from(Array(totalWeeks), (_, index) => {
+            const startDate = addDays(week, 7 * index);
+            const endDate = addDays(startDate, 6);
+
+            startDate.setHours(0, 0, 0, 0);
+            endDate.setHours(11, 59, 59, 999);
+            return (
+                <Week
```
LeandroTorresSicilia, line 20: same ^^ this should be a component that makes the iteration

src/components/Calendar/index.d.ts
```diff
@@ -2,10 +2,12 @@ import { ComponentType } from 'react';
 import { BaseProps } from '../types';
 
 export interface CalendarProps extends BaseProps {
+    id?: string;
```
LeandroTorresSicilia, line 5: good job!!!

LeandroTorresSicilia:
There is an issue with the dark theme:
Screen Shot 2020-05-21 at 11 40 17 PM

TahimiLeonBravo:
> There is an issue with the dark theme:
> Screen Shot 2020-05-21 at 11 40 17 PM

@wildergd I think we should use colors with opacity here like an RGBA with a value on the opacity different of 1

src/components/Calendar/day.js
```diff
@@ -31,18 +51,29 @@ function DayComponent(props) {
...
+                isFirstDayOfWeek={isFirstDayOfWeek}
+                isLastDayOfWeek={isLastDayOfWeek}
             >
-                {day}
-            </StyledDayButton>
+                <StyledDayButton
+                    ref={buttonRef}
+                    tabIndex={tabIndex}
+                    onClick={() => onChange(new Date(date))}
+                    onMouseEnter={() => privateOnHover(new Date(date))}
```
maxxgreene, line 65: why we need to create a new instance here, if we need to avoid date mutation the clone should happen inside the privateOnHover or onChange

src/components/Calendar/helpers/isDateWithinRange.js
```diff
@@ -0,0 +1,7 @@
+import compareDates from './compareDates';
+
+export default function isDateWithinRange(date, range) {
+    if (!date || !Array.isArray(range) || range.length <= 1) return false;
```
maxxgreene, line 4: I guess in the UCLV we learn to exit quickly in a function :) ... almost every code that I review of UCLV folks it's like that ... but actually it's difficult to read ... try it like this and check if makes sense for you

```js
if (data && Array.isArray(range) && range.length > 1) {
...
}
return false;
```

as a rule of thumbs try to avoid negations(!) when possible on if conditions

maxxgreene:
The implementation looks good to me, I would say the helpers should be written avoiding (!) in conditions.

Issue #1552, opened 2020-05-15
fix: the last Table column is not calculated when the columns have a width fixed


TahimiLeonBravo:
#### Description
When the `columns` have a `width` the last column without `width` on the `Table`, don't grow to cover the width.

#### Steps to repro
1. Go to the `Table` component page or any project when you use the Table
2. Add 4 `Columns` components
3. Add the `width={50}` prop to the first 3 `Column` components 

#### Current Behavior
The last column doesn't grow to cover  the width available
<img width="1348" alt="Screen Shot 2020-05-18 at 10 03 20 PM" src="https://user-images.githubusercontent.com/32083556/82280264-18c3a900-9954-11ea-8436-b5c060ad79fd.png">

#### Expected Behavior
The last column should grow to cover  the width available

PR #1564, opened 2020-05-19
fix: the last Table column is not calculated when the columns have a …

rgah2107:
fix: #1552

- the last Table column is not calculated when the columns have a width fixed

@nexxtway/react-rainbow


src/components/Table/index.js
```diff
@@ -405,7 +405,7 @@ export default class Table extends Component {
         };
         const maxRowSelection = this.getMaxRowSelection();
         const minColWidth = Number(minColumnWidth) || 50;
-        const maxColWidth = Number(maxColumnWidth) || 1000;
+        const maxColWidth = Number(maxColumnWidth) || 5000;
```
TahimiLeonBravo, line 408: Let's use the `Infinity` instead `5000`

PR #1568, opened 2020-05-19
fix: fileSelector styled don't grow properly base on container

blakmetall:
fix: fileSelector styled don't grow properly base on container

It was simply needed to expand using width: 100%

- [x] tested using a file with a long pdf name
- [x] tested growing the container


blakmetall:
Screen Shot 2020-05-19 at 13 07 59
Screen Shot 2020-05-19 at 13 08 11

Issue #1566, opened 2020-05-19
feat: add remove files action on FileSelector  when multiline


TahimiLeonBravo:
- [x] add remove files action on FileSelector when multiline
https://zpl.io/VkRoE0v

blakmetall:
- [x] add remove files action on FileSelector when multiline

PR #1571, opened 2020-05-19
feat: add remove files action on file selector when multiline

blakmetall:
fix: #1566  

- add remove files action on file selector when multiline

@nexxtway/react-rainbow


TahimiLeonBravo:
@blakmetall let's show in the multiline variant the cancel button only when you hover
https://zpl.io/VkRoE0v

blakmetall:
- [x] show in the multiline variant the cancel button only when you hover

TahimiLeonBravo:
@blakmetall review the cancel icon position, it should look like the design

Group 236

src/components/FileSelector/styled/index.js
```diff
@@ -186,6 +213,17 @@ export const StyledButtonIcon = styled(ButtonIcon)`
     svg {
         pointer-events: none;
     }
+
+    ${props =>
+        props.variantValue === 'multiline' &&
+        `
+        position: absolute;
+        top: -14px;
```
LeandroTorresSicilia, line 221: why we need to pass `-14px`?
It is a bad practice since it makes the code more complex and difficult for developer's understanding

what's happen with the button if on the container styles example we add ` height: 500,`
`const containerStyles = {
    height: 500,
};`

@TahimiLeonBravo please, review it

blakmetall: ok, never try here again negative values

TahimiLeonBravo: we can't use this `top: -14px` here because when we change the height of the container we get it 
Screen Shot 2020-05-21 at 8 19 52 PM
@blakmetall  we need find a better solution here, we should follow always the best-practices and be consitent

TahimiLeonBravo:
@blakmetall please review the conflicts here

blakmetall:
> @blakmetall please review the conflicts here

ready

src/components/FileSelector/index.js
```diff
@@ -138,6 +138,20 @@ const FileSelector = React.forwardRef((props, ref) => {
                 disabled={disabled}
                 error={error}
             >
+                <RenderIf isTrue={variant === 'multiline'}>
```
HellWolf93, line 141: We should avoid using logic inside JSX. You can declare
`const shouldRenderCancelIcon = variant === 'multiline'`
and then 
`<RenderIf isTrue={shouldRenderCloseIcon}>`

Issue #1544, opened 2020-05-13
feat: add the `selectedIcon` prop to the `SidebarItem` component


TahimiLeonBravo:
As a customer, I want to show a different icon when an item is selected.

#### Requirements
- [ ] it should have a `selectedIcon` prop to show a different icon on the selected item.
- [ ] it should use the `icon` as `selectedIcon` as fallback when the `selectedIcon` wasn't passed
- [ ] it should be tested

PR #1574, opened 2020-05-20
feat: add the selectedIcon prop to the SidebarItem component

rgah2107:
fix: #1544

- add the `selectedIcon` prop to the `SidebarItem` component

@nexxtway/react-rainbow


src/components/SidebarItem/itemContent.js
```diff
@@ -5,10 +5,10 @@ import StyledIcon from './styled/icon';
 import StyledLabel from './styled/label';
 
 export default function ItemContent(props) {
-    const { label, icon, isSelected } = props;
+    const { label, icon, selectedIcon, isSelected } = props;
     return (
         <Fragment>
-            <StyledIcon>{icon}</StyledIcon>
+            {isSelected ? <StyledIcon>{selectedIcon}</StyledIcon> : <StyledIcon>{icon}</StyledIcon>}
```
yvmunayev, line 11: normally we use RenderIf for all internal conditionals of the lib

src/components/SidebarItem/__test__/sidebarItem.spec.js
```diff
@@ -36,6 +36,22 @@ describe('<SidebarItem when href is passed />', () => {
...
+                    icon={<FontAwesomeIcon icon={faUser} />}
+                    selectedIcon={<FontAwesomeIcon icon={faClock} />}
+                    name="sidebaritem-test-4"
+                    label="sidebaritem-test-4"
+                />
+            </Sidebar>,
+        );
+        const sidebarLink = component.find('a');
+        sidebarLink.simulate('click');
+        expect(component.find({ 'data-icon': 'clock' }).exists()).toBe(true);
```
LeandroTorresSicilia, line 53: This is something internal from FontAwesome? If true then we need to change this assertion since we can't make introspection inside what we could not control, font awesome can change this behaviour and then if we update then our test take the risk to fail, instead, we should find the FontAwesomeIcon component in the tree of react components and check that the icon prop passed is faClock

Issue #1569, opened 2020-05-19
styles: the colors on the FileSelector should be customized with the theme


TahimiLeonBravo:
#### Requirements
- [x] The backgrounds colors when drag&drop and error should be rgba format like the design
<img width="373" alt="Screen Shot 2020-05-19 at 2 11 04 PM" src="https://user-images.githubusercontent.com/32083556/82368173-9631fc80-99da-11ea-85f1-a1607866bde9.png">

- [x] The `error`, `cancel` icons colors should be customized with the theme
- [x] The upload icon should be rendered on the component when dragging over the component when one or many files are selected

PR #1575, opened 2020-05-21
styles: the colors on the FileSelector should be customized with the theme

blakmetall:
fix: #1569  

- fix colors on FileSelector

@nexxtway/react-rainbow


blakmetall:
I am not completely sure if the implementation I applied to ErrorIcon is the best, 

... however, in the solution, I added a control that won't affect the current usage of the ErrorIcon svg component

src/components/FileSelector/icons/error.js
```diff
@@ -1,6 +1,10 @@
 import React from 'react';
+import PropTypes from 'prop-types';
+
+function ErrorIcon({ inheritFillColor }) {
+    const fillCircleColor = inheritFillColor ? 'currentColor' : '#FE4849';
```
TahimiLeonBravo, line 5: we don't need to do this check here instead we should
on
`circle fill="currentColor"`
and on
`path fill="#000" fillOpacity=".102"`

src/components/FileSelector/styled/index.js
```diff
@@ -60,7 +61,7 @@ export const StyledDropzone = attachThemeAttrs(styled.div)`
             cursor: not-allowed;
             border-color: ${props.palette.text.disabled}
         }
-        `}
+    `}
 
     ${props =>
         props.error &&
```
TahimiLeonBravo, line 67: the `background-color` when `error` also should be transparent, we should use anything like that:

`background-color: ${replaceAlpha(props.palette.error.main, 0.1)};`

src/components/FileSelector/icons/error.js
```diff
@@ -13,15 +13,15 @@ export default function ErrorIcon() {
...
                                 cx="11.9766316"
                                 cy="11.9766316"
                                 r="11.9766316"
                             />
                             <path
                                 d="M23.9532105,11.9766316 C23.9532105,18.5637895 18.5637368,23.9532632 11.9765789,23.9532632 C8.15905263,23.9532632 4.79063158,22.2316316 2.61984211,19.4620526 C4.64089474,21.1088421 7.26078947,22.0819474 10.1052105,22.0819474 C16.6923684,22.0819474 22.0818421,16.6924737 22.0818421,10.1053158 C22.0818421,7.26089474 21.1087368,4.641 19.4619474,2.61994737 C22.2315789,4.79063158 23.9532105,8.15905263 23.9532105,11.9766316 Z"
-                                id="Shape"
-                                fill="#EA4243"
+                                id="OvalPartialShape"
+                                fill="currentColor"
```
TahimiLeonBravo, line 24: let's use  `fill="#000" fillOpacity=".102"` on the `path` element

src/components/FileSelector/styled/index.js
```diff
@@ -172,8 +180,20 @@ export const StyledIconContainer = attachThemeAttrs(styled(IconContainer))`
...
             font-size: 64px !important;
+
+            ${props.error &&
+                `
+                height: 40px !important;
+                width: 40px !important;
+                font-size: 40px !important;
+                color: ${props.palette.error.main};
+                
+                #OvalPartialShape {
```
TahimiLeonBravo, line 191: we should not overcomplicate this if we use these styles `fill="#000" fillOpacity=".102"` on the `path` element we don't need to use this.
using the above styles we guaranty these color always will be darker

blakmetall: sure, no probs, keep it simple

src/components/FileSelector/helpers/getIcon.js
```diff
@@ -1,15 +1,17 @@
...
-    if (error) {
-        return <ErrorIcon />;
-    }
-    if (files && files.length === 1) {
-        return <FileIcon />;
-    }
-    if (files && files.length > 1) {
-        return <FilesIcon />;
+export default function getIcon(files, error, isDragOver) {
+    if (!isDragOver) {
```
TahimiLeonBravo, line 5: Let's avoid the negation here and say

PR #1582, opened 2020-05-23
feat: implement enumerate rows to table component

blakmetall:
feat: implement enumerate rows to table component


src/components/Table/helpers/columns/getColumns.js
```diff
@@ -41,14 +54,37 @@ export default function getColumns(params) {
...
+    };
+
+    if (showRowNumberColumn) {
+        extraColumns.withEnumerable = {
+            type: WITH_ENUMERABLE,
+            rowNumberOffset,
+            width: getEnumerableWidth(rowNumberOffset),
+        };
+    }
+
```
yvmunayev, line 69: crea que seria mas sencillos hacerla esto

`if (showRowNumberColumn) {
        columnsData .push({
            type: WITH_ENUMERABLE,
            rowNumberOffset,
            width: getEnumerableWidth(rowNumberOffset),
        });
    }

if (showCheckboxColumn) {
         columnsData .push({
            type: SELECTABLE_CHECKBOX,
            width: 52,
        });
    }
return columnsData;

`

LeandroTorresSicilia:
If you inspect an example with selection you will see this:
Screen Shot 2020-05-25 at 9 37 43 AM
it means that the first element in the row is a td tag with role="gridcell" (the column with the checkbox) and the second is a th tag with scope="row" (the first column with real content data)
In the case of the number column we should do the same

src/components/Table/index.d.ts
```diff
@@ -11,6 +11,8 @@ export interface TableProps extends BaseProps {
     minColumnWidth?: string | number;
     maxColumnWidth?: string | number;
     showCheckboxColumn?: boolean;
+    showRowNumberColumn: boolean;
+    rowNumberOffset: number;
```
LeandroTorresSicilia, line 15: remember no props are required, this is very important since in this way typescript user will must pass showRowNumberColumn and rowNumberOffset and then the current usages of the table would break, instead:
```
showRowNumberColumn?: boolean;
rowNumberOffset?: number;
```

src/components/Table/body/enumerableCell.js
```diff
@@ -0,0 +1,26 @@
...
+export default function EnumerableCell({ rowNumberOffset, rowIndex }) {
+    const rowNumber = rowNumberOffset + rowIndex;
+    return (
+        <StyledCell scope="row" tabIndex={-1}>
+            <StyledCellContent>
+                <StyledEnumerable title={rowNumber}>{rowNumber}</StyledEnumerable>
+            </StyledCellContent>
+        </StyledCell>
+    );
+}
```
LeandroTorresSicilia, line 16: take a look here:
https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Lists_and_Counters/Using_CSS_counters
https://css-tricks.com/almanac/properties/c/counter-increment/
I think we should use this since it will be more simple and much more performance

src/components/Table/helpers/columns/getColumns.js
```diff
@@ -41,14 +71,9 @@ export default function getColumns(params) {
...
-        return [
-            {
-                type: SELECTABLE_CHECKBOX,
-                width: 52,
-            },
-            ...columnsData,
-        ];
+    if (configColumns.length) {
+        return configColumns.concat(columnsData);
     }
```
LeandroTorresSicilia, line 76: we should add new test cases for this

blakmetall:
> If you inspect an example with selection you will see this:
> Screen Shot 2020-05-25 at 9 37 43 AM
> it means that the first element in the row is a td tag with role="gridcell" (the column with the checkbox) and the second is a th tag with scope="row" (the first column with real content data)
> In the case of the number column we should do the same
> We should add unit test to this, to ensure that in the case where you use two columns before the data like using numbers and selection columns, then the third element in the row is which use the th tag and the scope="row" and the others are td elements with role="gridcell"

src/components/Table/index.d.ts
blakmetall: sure

blakmetall:
> If you inspect an example with selection you will see this:
> Screen Shot 2020-05-25 at 9 37 43 AM
> it means that the first element in the row is a td tag with role="gridcell" (the column with the checkbox) and the second is a th tag with scope="row" (the first column with real content data)
> In the case of the number column we should do the same
> We should add unit test to this, to ensure that in the case where you use two columns before the data like using numbers and selection columns, then the third element in the row is which use the th tag and the scope="row" and the others are td elements with role="gridcell"

...updating markup and adding test cases on this

src/components/Table/body/enumerableCell.js
blakmetall: applying....

src/components/Table/helpers/columns/getColumns.js
blakmetall: on the way...

LeandroTorresSicilia:
@blakmetall why you close this pr?

blakmetall:
> @blakmetall why you close this pr?
> we shouldn't close PRs unnecessarily

by mistake I believe, I wanted to delete a comment

blakmetall:
#### body/cell.js

- - - - <Cell /> when isFirst is not passed - - - -
- [ ] should render the enumerableCell component when columnType is "WITH_ENUMERABLE"

- - - - <Cell /> when isFirst is is passed - - - -
- [ ] should render the SelectableCell component when columnType is "WITH_ENUMERABLE"

#### body/enumerableCell.js (not sure if this one should or needs to be tested)


#### helpers/getColumns.js 

**current cases:**

- [x] should return null when children is null, showCheckboxColumn is false
- [x] should return an empty array when children is not passed
- [x] should return an empty array when children is an array with falsy values and showCheckboxColumn is false
- [x] should return an array with the columns props when showCheckboxColumn is false
- [x] should return an array with the columns props plus the selectable column when showCheckboxColumn is true
- [x] should return an array with the right columns props when one column is type "action"
- [x] should return an array with the right columns props when defaultWidth and width are passed
- [x] should return an array with the right columns props when defaultWidth is passed and is out of range

- - - - - - - - 

**proposed cases:**

- [ ] should return null when children is null, showCheckboxColumn is false and showRowNumberColumn is false (UPDATE)
- [ ] should return an empty array when children is not passed, showCheckboxColumn is false 
 and showRowNumberColumn is false (UPDATE)
- [ ] should return an empty array when children is an array with falsy values, showCheckboxColumn is false and showRowNumberColumn is false (UPDATE)
- [ ] should return an array with the columns props when showCheckboxColumn is false and showRowNumberColumn is false (UPDATE)
- [ ] should return an array with the columns props plus the enumerable and the selectable column when showCheckboxColumn is true and showRowNumberColumn is true (NEW)
- [x] should return an array with the columns props plus the selectable column when showCheckboxColumn is true (UNCHANGED)
- [ ] should return an array with the columns props plus the enumerable column when showRowNumberColumn is true (NEW)
- [ ] should return an array with the enumerable column containing the right rowNumberOffset value when showRowNumberColumn is true and rowNumberOffset is passed (NEW)
- [x] should return an array with the right columns props when one column is type "action" (UNCHANGED)
- [x] should return an array with the right columns props when defaultWidth and width are passed (UNCHANGED)
- [x] should return an array with the right columns props when defaultWidth is passed and is out of range (UNCHANGED)

- - - - - - - - 

#### body/cell.js

... filling spec

src/components/Table/__test__/table.spec.js
```diff
@@ -169,6 +223,53 @@ describe('<Table />', () => {
...
+                width: 50,
+            },
+            {
+                field: 'name',
+                header: 'Name',
+                sortable: false,
+                computedWidth: 50,
+                type: 'text',
+            },
+        ]);
```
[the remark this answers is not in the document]
LeandroTorresSicilia: Yeah, I will be working on this

Issue #1554, opened 2020-05-16
feat: add the `variant` prop to the Calendar component


TahimiLeonBravo:
#### Requirements
- [x] It should have a variant `prop` with the `'single' | 'doble'` values
- [x] It should have by default the `single` value

#### Design Guidelines
https://zpl.io/bzEjZ7E

#### Design Example
https://zpl.io/a7vLJZK
example 4

PR #1583, opened 2020-05-23
feat: add the variant prop to the Calendar component

wildergd:
fix: #1554 

- add the variant prop to the Calendar component

@nexxtway/react-rainbow


src/components/Calendar/doubleCalendar/yearSelect.js
```diff
@@ -0,0 +1,64 @@
...
+        if (editMode) {
+            listenOutsideClick();
+        }
+
+        return () => {
+            if (editMode) {
+                stopListeningOutsideClick();
+            }
+        };
+    }, [editMode, listenOutsideClick, stopListeningOutsideClick]);
```
yvmunayev, line 29: @maxxgreene @LeandroTorresSicilia @TahimiLeonBravo  here this same logic of useOutsideClick I think we should change fast.

I think we should do an internal component with the logic of the editMode

src/components/Calendar/readme.md
```diff
@@ -97,3 +97,57 @@ const calendarContainerStyles = {
     </div>
 </div>
 ```
+
+##### Calendar with variant double:
+
+```js
+import React from 'react';
+import { Card, Calendar, Avatar } from 'react-rainbow-components';
```
yvmunayev, line 105: delete Card and Avatar

src/components/Calendar/readme.md
```diff
@@ -97,3 +97,57 @@ const calendarContainerStyles = {
...
+        </div>
+    </div>
+</div>
+```
+
+##### Calendar with variant double and range selection:
+
+```js
+import React from 'react';
+import { Card, Calendar, Avatar } from 'react-rainbow-components';
```
yvmunayev, line 129: Same before

src/components/Calendar/doubleCalendar/yearSelect.js
```diff
@@ -0,0 +1,64 @@
...
+import PropTypes from 'prop-types';
+import { useOutsideClick } from '../../../libs/hooks';
+import RenderIf from '../../RenderIf';
+// import Select from '../../Select';
+import StyledYearLabel from './styled/yearLabel';
+import StyledSelect from './styled/select';
+
+export default function YearSelect(props) {
+    const { currentYear, yearsRange, onYearChange } = props;
+    const selectRef = useRef();
```
yvmunayev, line 11: I think it is more descriptive containerRef

src/components/Calendar/doubleCalendar/yearSelect.js
```diff
@@ -0,0 +1,64 @@
+import React, { useRef, useState, useEffect } from 'react';
+import PropTypes from 'prop-types';
+import { useOutsideClick } from '../../../libs/hooks';
+import RenderIf from '../../RenderIf';
+// import Select from '../../Select';
```
yvmunayev, line 5: no comment code

src/components/Calendar/doubleCalendar/monthHeader.js
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { useFormattedMonth } from './hooks';
+import StyledHeaderContainer from './styled/headerContainer';
+import StyledMonth from './styled/month';
+import YearSelect from './yearSelect';
+
+export default function MonthHeader(props) {
+    const { id, month, locale, currentYear, yearsRange, onYearChange } = props;
+    const formattedLabel = useFormattedMonth(month, locale);
```
yvmunayev, line 10: I think that in these situations the best option is to make useMemo or useCallback to the props and functions that are going to be passed in the parent components and the child components to do React.memo (). This ensures that the child components run only when necessary.

src/components/Calendar/doubleCalendar/index.js
```diff
@@ -0,0 +1,350 @@
...
+                const keyHandler = altKey ? keyHandlerMapAlt : keyHandlerMap;
+                if (keyHandler[keyCode]) {
+                    event.preventDefault();
+                    event.stopPropagation();
+                    keyHandler[keyCode]();
+                }
+            }
+        },
+        [enableNavKeys, keyHandlerMap, keyHandlerMapAlt],
+    );
```
yvmunayev, line 230: I think all keyboard logic should be in a custom hook useHandleKeyDown

src/components/Calendar/doubleCalendar/index.js
```diff
@@ -0,0 +1,350 @@
...
+            const [currentRangeStart] = currentRange;
+
+            if (isDateBelowLimit(date, rangeStart)) {
+                setCurrentRange([currentRangeStart]);
+            } else {
+                setCurrentRange([rangeStart, date]);
+            }
+        },
+        [currentRange, selectedRange, selectionType],
+    );
```
yvmunayev, line 97: It is good practice in large components like this, put all hook calls in custom hook with descriptive names so that the code is much more readable.

src/components/Calendar/doubleCalendar/index.js
```diff
@@ -0,0 +1,350 @@
...
+    const secondMonthLabelId = useUniqueIdentifier('second-month');
+    const currentYear = currentMonth.getFullYear();
+    const rightCalendarYear = rightCalendarMonth.getFullYear();
+    const yearsRange = useYearsRange(minDate, maxDate, currentMonth);
+    const lastYearItem = yearsRange[yearsRange.length - 1];
+    const minSelectableDate = minDate || new Date(yearsRange[0].value, 0, 1);
+    const maxSelectableDate = maxDate || new Date(lastYearItem.value, 11, 31);
+    const lastDayMonth = useLastDayMonth(currentMonth);
+    const disableNextMonth = addMonths(rightCalendarMonth, 1) > maxSelectableDate;
+    const disablePreviousMonth = lastDayMonth < minSelectableDate;
```
yvmunayev, line 74: I think you should put all this in a custom Hook useDisabledControls that returns [isDisableNextMonth, isDisablePreviousMonth] and useLastDayMonth is not necessary because it will only be used inside useDisabledControls

src/components/Calendar/doubleCalendar/index.js
```diff
@@ -0,0 +1,350 @@
...
+    useMoveFocusedDayFunction,
+    useMoveFocusedMonthFunction,
+} from './hooks';
+import StyledControlsContainer from '../styled/controlsContainer';
+import StyledArrowButton from '../styled/arrowButton';
+import StyledMonthsContainer from './styled/monthsContainer';
+import StyledTable from '../styled/table';
+import StyledContainer from './styled/container';
+import StyledCalendar from './styled/calendar';
+import StyledDivider from './styled/divider';
```
yvmunayev, line 42: I think you should put an index.js in the styled folders and export all this styled component

src/components/Calendar/doubleCalendar/hooks/useMoveFocusedDayFunction.js
```diff
@@ -0,0 +1,43 @@
...
+    addMonths,
+    isSameMonth,
+    getFirstDayMonth,
+    isDateBelowLimit,
+    isDateBeyondLimit,
+    getCalendarBounds,
+    getSign,
+} from '../../helpers';
+
+export default function useMoveFocusedDayFunction(focusedDate, currentMonth, minDate, maxDate) {
```
yvmunayev, line 13: i think it's better to useMoveFocusedMonth

src/components/Calendar/doubleCalendar/hooks/useMoveFocusedMonthFunction.js
```diff
@@ -0,0 +1,32 @@
+import { useMemo } from 'react';
+import {
+    addMonths,
+    isDateBelowLimit,
+    isDateBeyondLimit,
+    getFirstDayMonth,
+    getCalendarBounds,
+} from '../../helpers';
+
+export default function useMoveFocusedMonthFunction(focusedDate, minDate, maxDate) {
```
yvmunayev, line 10: same before

src/components/Calendar/doubleCalendar/styled/select.js
```diff
@@ -0,0 +1,21 @@
+import styled from 'styled-components';
+import { FONT_SIZE_HEADING_MEDIUM, FONT_SIZE_TEXT_LARGE } from '../../../../styles/fontSizes';
+import attachThemeAttrs from '../../../../styles/helpers/attachThemeAttrs';
+import Select from '../../../Select';
+
+const StyledSelect = attachThemeAttrs(styled(Select))`
```
TahimiLeonBravo, line 6: Since this select looks and work different to our Select component, I think we don't would use our Select component, we should style the native select
This way we avoid doing introspection in the Select component

TahimiLeonBravo:
@wildergd we need to review the `selectionType` the `onChange` should send on the first value with the time on 12:00 am and on the second value with the time 11:59:59 pm in order to include the complete information of the second day
thus the component use is easier

src/components/Calendar/doubleCalendar/styled/select.js
```diff
@@ -0,0 +1,37 @@
+import styled from 'styled-components';
+import { FONT_SIZE_HEADING_MEDIUM, FONT_SIZE_TEXT_LARGE } from '../../../../styles/fontSizes';
+import { PADDING_MEDIUM } from '../../../../styles/paddings';
+import attachThemeAttrs from '../../../../styles/helpers/attachThemeAttrs';
+import Select from '../../../Select';
+
+const StyledSelect = attachThemeAttrs(styled(Select))`
+    select {
```
TahimiLeonBravo, line 8: Here we are doing bad practices, we have to avoid doing introspection because it could give us problems. whats happen if in the future the Select change? this could break
Since we are changing almost all the Select styles then it is not feasible to use our Select component instead we use the native select.

src/components/Calendar/doubleCalendar/hooks/useDisabledControls.js
```diff
@@ -0,0 +1,22 @@
...
+import useLastDayMonth from './useLastDayMonth';
+
+export default function useDisabledControls(
+    yearsRange,
+    minDate,
+    maxDate,
+    currentMonth,
+    rightCalendarMonth,
+) {
+    const lastDayMonth = useLastDayMonth(currentMonth);
```
yvmunayev, line 12: const lastDayMonth = getLastDayMonth(addMonths(currentMonth, -1)) inside the useMemo and you don't need this hook anymore.

src/components/Calendar/doubleCalendar/styled/container.js
```diff
@@ -0,0 +1,10 @@
+import styled from 'styled-components';
+
+const StyledContainer = styled.div`
```
LeandroTorresSicilia, line 3: Now we put all styled in one file, look here: `src/components/FileSelector/styled/index.js`

src/components/Calendar/helpers/buildNewRangeFromValue.js
```diff
@@ -2,37 +2,34 @@ import isDateWithinRange from './isDateWithinRange';
 import compareDates from './compareDates';
 
 export default function buildNewRangeFromValue(value, currentRange, currentUpdatePosition = 0) {
-    if (!currentRange || currentRange.length === 0)
-        return {
-            range: [value],
-        };
+    if (currentRange && currentRange.length > 0) {
+        const [rangeStart, rangeEnd] = currentRange;
+        rangeStart.setHours(0, 0, 0, 0);
```
LeandroTorresSicilia, line 7: be carefull doing this since these date methods make mutations and can lead to unexpected behaviors and hard to track

Issue #1585, opened 2020-05-24
feat: improve the Spinner component


TahimiLeonBravo:
#### Requirements
- [ ] It should have a `type` prop with the `'circle' | 'arc'` values, by default the `circle` value
- [ ] Add a new `size="x-large"`
- [ ] It should have a `children` prop that will be rendered on the center of the component.

#### Design 
https://zpl.io/agNWMr1

PR #1586, opened 2020-05-27
feat: improve the Spinner component

rgah2107:
fix: #1585

- improve the Spinner component

@nexxtway/react-rainbow


src/components/Spinner/readme.md
```diff
@@ -115,3 +115,37 @@ const Loading = styled.h1.attrs(props => {
     </Loading>
 </InverseContainer>
 ```
+
+##### list loading using spinner - brand - large - arc
```
HellWolf93, line 119: I think this example names are missleading, there is nothing related to lists here. I am thinking about using something like `Arc Spinner large brand`; same for the rest of the exampe headers

src/components/Spinner/readme.md
```diff
@@ -115,3 +115,37 @@ const Loading = styled.h1.attrs(props => {
...
+    <GlobalHeader src="images/user/user3.jpg" />
+    <div className="rainbow-p-vertical_xx-large">
+        <div className="rainbow-align-content_center rainbow-position_relative rainbow-p-vertical_xx-large">
+            <Spinner size="large" type="arc" variant="brand" />
+        </div>
+    </div>
+</div>
+```
+
+##### list loading using spinner - brand - x-large - arc - with logo
```
HellWolf93, line 135: ^^ same

HellWolf93:
There is a bug when spinner size is smaller than children, I believe we should limit the size of the icon to the inside of the circle.
Screenshot_20200527_215726
Screenshot_20200527_215752

src/components/Spinner/index.js
```diff
@@ -6,23 +6,42 @@ import StyledArcSpinner from './styled/arcSpinner';
 import StyledSpinnerContainer from './styled/spinnerContainer';
 import StyledChildContainer from './styled/childContainer';
 
+const sizeMap = {
```
HellWolf93, line 9: since you are using the same size map in StyledArcSpinner i think you could put it in a separate file and import it from both places

TahimiLeonBravo:
+1

src/components/Spinner/index.js
```diff
@@ -1,18 +1,46 @@
...
  */
 export default function Spinner(props) {
-    const { className, style, assistiveText, isVisible, size, variant } = props;
+    const { className, style, assistiveText, isVisible, size, variant, type, children } = props;
+    const currentSize = getSizeValue(size);
 
     if (isVisible) {
+        if (type === 'arc') {
+            return (
+                <StyledSpinnerContainer>
```
LeandroTorresSicilia, line 20: we need to pass className and style prop to the container, all component containers receive these props in order to pass styles to positioning, margin and other styles they want to pass

src/components/Spinner/helpers/getSizeValue.js
```diff
@@ -0,0 +1,15 @@
...
+    small: '26',
+    'x-small': '20',
+    'xx-small': '16',
+};
+
+export default function getSizeValue(size) {
+    if (size) {
+        return sizeMap[size] || sizeMap.medium;
+    }
+    return sizeMap.medium;
```
yvmunayev, line 14: I think the if is too much with "return sizeMap [size] || sizeMap.medium;" the case that size is undefine is covered

Issue #1548, opened 2020-05-14
feat: implement DatePickerModal component


TahimiLeonBravo:
#### Requirements 
- [x] define API
https://docs.google.com/document/d/16zTKrI1Zfqld_ZV8I8MlxanXDdgQ4koQvL9IrACWL1s/edit?usp=sharing
- [x] implement component
- [ ] add tests

PR #1601, opened 2020-06-03
feat: implement DatePickerModal component

wildergd:
fix: #1548 

- implement DatePickerModal component
- fix minor issues in Calendar component

@nexxtway/react-rainbow


src/components/DatePickerModal/index.d.ts
```diff
@@ -0,0 +1,7 @@
+import { BaseProps } from '../types';
+
+export interface DatePickerModalProps extends BaseProps {
+    id?: string;
```
LeandroTorresSicilia, line 4: let's put all props here

src/components/DatePickerModal/index.js
```diff
@@ -0,0 +1,99 @@
...
+    /** The ending of a range of valid dates. The range includes the endDate.
+     * The default value is current date + 100 years. */
+    maxDate: PropTypes.instanceOf(Date),
+    /** The beginning of a range of valid dates. The range includes the startDate.
+     * The default value is current date - 100 years. */
+    minDate: PropTypes.instanceOf(Date),
+    /** The type of the selection. It can be a single date or a range. The default value is 'single'. */
+    selectionType: PropTypes.oneOf(['single', 'range']),
+    /** The calendar type. Defaults to 'single' */
+    calendarType: PropTypes.oneOf(['single', 'double']),
```
LeandroTorresSicilia, line 66: use variant instead of calendarType as we did in Calendar

src/components/DatePickerModal/index.js
```diff
@@ -0,0 +1,99 @@
...
+        PropTypes.arrayOf(PropTypes.oneOfType([PropTypes.instanceOf(Date), PropTypes.string])),
+    ]),
+    /** Controls whether the DatePickerModal is opened or not. */
+    isOpen: PropTypes.bool,
+    /** The action triggered when the component request to close
+     *  (e.g click close button, press esc key or click outside). */
+    onRequestClose: PropTypes.func,
+    /** The title can include text or another component,
+     * and is displayed in the header of the component. */
+    title: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
```
TahimiLeonBravo, line 82: The title in the modal will be shown according to the selected date, which is the use case we have now. 
We will not have this prop for now if in the future the use case appears where you need to pass it a customizable title then we will add the prop.

src/components/DatePickerModal/index.js
```diff
@@ -0,0 +1,99 @@
...
+        <StyledModal
+            id={id}
+            className={className}
+            style={style}
+            isOpen={isOpen}
+            onRequestClose={onRequestClose}
+        >
+            <Header>
+                <HeaderTitle>{title}</HeaderTitle>
+            </Header>
```
LeandroTorresSicilia, line 32: let's put this inside a RenderIf in order to avoid render markup when the title is not passed

Issue #1596, opened 2020-06-02
feat: add `multiple` prop to the `InternalDropdown` component


TahimiLeonBravo:
#### Requirements
- add multi-selection logic

PR #1605, opened 2020-06-04
feat: add `multiple` prop to the `InternalDropdown` component

HellWolf93:
fix: #1596

- Add `multiple` prop to InternalDropdown component


src/components/InternalDropdown/index.js
```diff
@@ -208,15 +231,35 @@ const InternalDropdown = forwardRef((props, reference) => {
...
+                    }
+                    return onChange(value.concat([option]));
+                }
+                if (value) {
+                    return onChange([value, option]);
+                }
+                return onChange([option]);
+            }
+            return onChange(option);
+        },
```
reiniergs, line 250: this function and the one above are kinds of the same, can we reuse it somehow?

src/components/InternalDropdown/helpers/getValueNames.js
```diff
@@ -0,0 +1,10 @@
+export default function getValueNames(value) {
+    if (Array.isArray(value)) {
+        const names = value.map(v => v.name);
+        return names.filter(v => !!v);
```
reiniergs, line 4: I that the options are going to be few and this code will run lightning-fast, but :) ... you are iterating twice ... you can use reduce here and make 1 iteration e.g.

```js
return value.reduce((seed, item) => {
    if (item.name) {
        seed.push(name);
    }
    return seed;
}, []);
```

reiniergs:
@HellWolf93 I don't see test for the component, are you gonna write integration tests for multiple selection??

/cc @TahimiLeonBravo @LeandroTorresSicilia

src/components/InternalDropdown/helpers/getValueNames.js
HellWolf93: you are right, thanks.

Issue #1562, opened 2020-05-19
feat: implement TagsInput component


TahimiLeonBravo:
#### Requirements
- [ ] define API
https://docs.google.com/document/d/1DmWbRSap7QbX1NWmg9ZM1aG3cKRqalDWcb0pBvNo4Y8/edit?usp=sharing
- [ ] Implement component
- [ ] add tests

#### Design
https://zpl.io/2pGjZBE

PR #1606, opened 2020-06-04
feat: implement MultiSelect component

HellWolf93:
fix: #1562

- Implement MultiSelect component
- Add tests to MultiSelect component


src/components/MultiSelect/index.js
```diff
@@ -0,0 +1,240 @@
...
+        handleOutsideClick,
+    );
+
+    const chips = getChips(value, chipVariant, handleDelete);
+    const haveChips = chips && chips.length > 0;
+
+    return (
+        <StyledContainer className={className} style={style}>
+            <Label label={label} hideLabel={hideLabel} required={required} inputId={inputId} />
+            <StyledInput
```
TahimiLeonBravo, line 113: Remember we have to follow the accessibility standards and since this component will be used on forms it needs to be an inputable.
Also, It needs to work with redux-form.
We need to review the w3c accessibility standards here in order to use the right HTML elements and attributes.

src/components/MultiSelect/readme.md
```diff
@@ -0,0 +1,176 @@
...
+
+const MultiSelectExample = props => {
+    const [value, setValue] = useState([]);
+
+    const handleChange = val => {
+        setValue(val);
+    }
+
+    return (
+        <MultiSelect
```
TahimiLeonBravo, line 19: The example should be work as a real use case

src/components/MultiSelect/icons/plus.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const PlusIcon = props => {
+    const { className, style } = props;
+
+    return (
+        <svg className={className} style={style} width="14" height="14" viewBox="0 0 14 14">
+            <path
+                fill="#01B6F5"
```
TahimiLeonBravo, line 10: In order to do this icon customized with the theme let's use `currentColor` instead of a fixed color

src/components/MultiSelect/readme.md
HellWolf93: sorry, just had to update the branch with the InternalDropdown multiple changes, examples should work as expected now

src/components/MultiSelect/index.js
yvmunayev: @TahimiLeonBravo  I have a question about the operation. This component will allow the input to be a filter and when the keyprees delete key the options are deleted

src/components/MultiSelect/helpers/getChips.js
```diff
@@ -0,0 +1,21 @@
...
+        return value.map(val => (
+            <StyledChip
+                key={val.name}
+                label={val.label}
+                variant={variant}
+                onDelete={() => deleteCallback(val)}
+            />
+        ));
+    }
+    return (
```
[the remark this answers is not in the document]
LeandroTorresSicilia: Yes, in order to be more declarative we should follow this rule, if a helper returns a component it should be a component

src/components/MultiSelect/helpers/__test__/getChips.spec.js
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
+import { mount } from 'enzyme';
+import getChips from '../getChips';
+import { StyledChip } from '../../styled';
+
+describe('getChips', () => {
+    test('should return null when value is falsey', () => {
```
LeandroTorresSicilia, line 7: let's use it to be consistent

src/components/MultiSelect/helpers/getChips.js
```diff
@@ -0,0 +1,21 @@
...
+    if (!value) {
+        return null;
+    }
+    if (Array.isArray(value)) {
+        return value.map(val => (
+            <StyledChip
+                key={val.name}
+                label={val.label}
+                variant={variant}
+                onDelete={() => deleteCallback(val)}
```
LeandroTorresSicilia, line 14: I think it is more readable if we use onDelete

src/components/MultiSelect/icons/plus.js
LeandroTorresSicilia: yes, we should use:
```fill="currentColor"```
in order to get the color from a parent element

File: src/components/MultiSelect/index.js [hunk 7]
```diff
@@ -0,0 +1,240 @@
...
+        if (event.target !== inputRef.current) return;
+        if (event.keyCode === ENTER_KEY || event.keyCode === SPACE_KEY) {
+            event.preventDefault();
+            if (disabled || readOnly) return;
+            setIsOpen(!isOpen);
+            if (!isOpen) {
+                // eslint-disable-next-line no-use-before-define
+                startListeningOutsideClick();
+            }
+        }
```
LeandroTorresSicilia, line 89: since this pattern is being used twice, I think we should create a handler for this like:
```
const toggleDropdown = () => {
 if (disabled || readOnly) return;
        setIsOpen(!isOpen);
        if (!isOpen) {
            // eslint-disable-next-line no-use-before-define
            startListeningOutsideClick();
        }
}
```

src/components/MultiSelect/index.js
```diff
@@ -0,0 +1,240 @@
...
+            setIsOpen(false);
+        }
+    };
+    const [startListeningOutsideClick, stopListeningOutsideClick] = useOutsideClick(
+        dropdownRef,
+        handleOutsideClick,
+    );
+
+    const chips = getChips(value, chipVariant, handleDelete);
+    const haveChips = chips && chips.length > 0;
```
LeandroTorresSicilia, line 108: hasChips

src/components/MultiSelect/index.js
```diff
@@ -0,0 +1,240 @@
...
+        readOnly,
+        tabIndex,
+        variant,
+        chipVariant,
+        value,
+        onChange,
+        onFocus,
+        onBlur,
+        children,
+    } = props;
```
LeandroTorresSicilia, line 41: since this is an inputable component it should use useReduxForm

src/components/MultiSelect/index.js
```diff
@@ -0,0 +1,240 @@
...
+            <StyledInput
+                variant={variant}
+                error={error}
+                disabled={disabled}
+                onFocus={onFocus}
+                onBlur={onBlur}
+                onKeyDown={handleKeyDown}
+                tabIndex={tabIndex}
+                ref={inputRef}
+            >
```
LeandroTorresSicilia, line 122: we should follow accessibility from here:
https://www.lightningdesignsystem.com/components/combobox/#Multiple-selections
take a careful look to the markup used and all the aria, role and other attributes used

src/components/index.js
```diff
@@ -40,6 +40,7 @@ export { default as MenuDivider } from './MenuDivider';
 export { default as MenuItem } from './MenuItem';
 export { default as Modal } from './Modal';
 export { default as MonthlyCalendar } from './MonthlyCalendar';
+export { default as MultiSelect } from './MultiSelect';
```
LeandroTorresSicilia, line 43: remember export it in the index.d.ts too

src/components/index.js
```diff
@@ -41,6 +41,7 @@ export { default as MenuDivider } from './MenuDivider';
...
```
LeandroTorresSicilia, line 44: Define component API in index.d.ts file

src/components/MultiSelect/index.d.ts
```diff
@@ -0,0 +1,27 @@
...
+    disabled?: boolean;
+    readOnly?: boolean;
+    variant?: 'default' | 'bare';
+    chipVariant?: 'base' | 'neutral' | 'outline-brand' | 'brand';
+    hideLabel?: boolean;
+    value?: MultiSelectOption | Array<MultiSelectOption>;
+    onChange?: (value: Array<MultiSelectOption>) => void;
+    onFocus?: (event: FocusEvent<HTMLInputElement>) => void;
+    onBlur?: (event: FocusEvent<HTMLInputElement>) => void;
+}
```
LeandroTorresSicilia, line 25: add children prop

src/components/MultiSelect/index.d.ts
```diff
@@ -0,0 +1,27 @@
...
+    error?: ReactNode;
+    bottomHelpText?: ReactNode;
+    tabIndex?: string | number;
+    required?: boolean;
+    disabled?: boolean;
+    readOnly?: boolean;
+    variant?: 'default' | 'bare';
+    chipVariant?: 'base' | 'neutral' | 'outline-brand' | 'brand';
+    hideLabel?: boolean;
+    value?: MultiSelectOption | Array<MultiSelectOption>;
```
LeandroTorresSicilia, line 21: in this component, we will use an array with strings as the value and for the onChange like:
```
value?: string[];
onChange?: (value: string[]) => void;
```
where the string is the name that identifies the option

src/components/InternalDropdown/index.js
```diff
@@ -73,6 +73,7 @@ const InternalDropdown = forwardRef((props, reference) => {
             }
             return containerRef.current.focus();
         },
+        contains: element => containerRef.current.contains(element),
```
LeandroTorresSicilia, line 76: Remember we need to be careful in what we expose here.
In this case why we need this?

HellWolf93: It is needed in useOutsideClick hook, otherwise throws the error 'ref.current.contains' is not a function,  it is used there to check if the target of the click event is a descendant of the element being rendered by the InternalOverlay component

LeandroTorresSicilia:
- when selecting an option with the keyboard it does not close the menu.
- the menu should close when pressing ESC and return focus to trigger element (like in the Picklist)

src/components/MultiSelect/helpers/__test__/hasChips.spec.js
```diff
@@ -0,0 +1,19 @@
...
+        const values = [
+            undefined,
+            null,
+            0,
+            true,
+            [],
+            { label: 'Label', name: 'Name' },
+            [{ label: 'Label', name: 'Name' }],
+        ];
+        const expected = [false, false, false, false, false, true, true];
```
LeandroTorresSicilia, line 14: we should add one test case for when false and other for when true

src/components/MultiSelect/helpers/__test__/normalizeValue.spec.js
```diff
@@ -0,0 +1,27 @@
...
+            [{ label: 'Label', name: 'Name', value: 'Value', icon: 'Icon' }],
+        ];
+        const expected = [
+            [],
+            [],
+            [],
+            [],
+            [],
+            [{ label: 'Label', name: 'Name', value: 'Value' }],
+            [{ label: 'Label', name: 'Name', value: 'Value' }],
```
LeandroTorresSicilia, line 21: same ^^ separate in two tests, one for empty array result and other for ```[{ label: 'Label', name: 'Name', value: 'Value' }]```

src/components/MultiSelect/helpers/normalizeValue.js
```diff
@@ -0,0 +1,18 @@
...
+    }
+    if (Array.isArray(value)) {
+        if (value.length === 0) {
+            return [];
+        }
+        return value.map(v => {
+            const newValue = v;
+            delete newValue.icon;
+            return newValue;
+        });
```
LeandroTorresSicilia, line 13: let's stick to functional programming, here you are coping the v reference to newValue and then when doing ```delete newValue.icon;``` there is a mutation of the newValue and thus to the original v object, then instead let's do:
```
return value.map((item) => {
    const { value, label, name  } = item;  // here we control exactly what we want to pass
    return {
        value,
        label,
        name,
    };
});
```

src/components/MultiSelect/helpers/normalizeValue.js
```diff
@@ -0,0 +1,18 @@
...
+    }
+    const newValue = value;
+    delete newValue.icon;
```
LeandroTorresSicilia, line 16: same here you are mutating the original value object, which is not a good practice

src/components/MultiSelect/index.d.ts
```diff
@@ -0,0 +1,29 @@
...
+    error?: ReactNode;
+    bottomHelpText?: ReactNode;
+    tabIndex?: string | number;
+    required?: boolean;
+    disabled?: boolean;
+    readOnly?: boolean;
+    variant?: 'default' | 'bare';
+    chipVariant?: 'base' | 'neutral' | 'outline-brand' | 'brand';
+    hideLabel?: boolean;
+    value?: MultiSelectOption | MultiSelectOption[];
```
LeandroTorresSicilia, line 22: vale must be only an array of MultiSelectOption

src/components/MultiSelect/index.js
```diff
@@ -0,0 +1,297 @@
...
+    /** Specifies the variant for the input. */
+    variant: PropTypes.oneOf(['default', 'bare']),
+    /** Specifies the variant for the chips. */
+    chipVariant: PropTypes.oneOf(['base', 'neutral', 'outline-brand', 'brand']),
+    /** Specifies the value of an input element. */
+    value: PropTypes.oneOfType([
+        PropTypes.shape({
+            name: PropTypes.string,
+            label: PropTypes.string,
+        }),
```
LeandroTorresSicilia, line 255: same here we should document this only as an array

src/components/MultiSelect/pageObject/index.js
```diff
@@ -0,0 +1,47 @@
...
+        this.rootElement = rootElement;
+    }
+
+    /**
+     * Clicks the component
+     * @method
+     */
+    click() {
+        $(this.rootElement)
+            .$('div')
```
LeandroTorresSicilia, line 21: I think we should be more specific in this selector like using the role or other attribute

src/components/MultiSelect/pageObject/index.js
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    /**
+     * Returns true when the combobox element has focus.
+     * @method
+     * @returns {bool}
+     */
+    hasFocus() {
+        return $(this.rootElement)
+            .$('div')
```
LeandroTorresSicilia, line 42: same ^^

src/components/index.js
LeandroTorresSicilia, line 44: you forget to export in `components/index.d.ts`

src/components/MultiSelect/helpers/normalizeValue.js
```diff
@@ -0,0 +1,26 @@
...
+    if (Array.isArray(value)) {
+        if (value.length === 0) {
+            return [];
+        }
+        return value.map(item => {
+            const { label, name, value: itemValue } = item;
+            return {
+                label,
+                name,
+                itemValue,
```
LeandroTorresSicilia, line 14: why use itemValue here?
this means that the onChange will send this instead of value

Issue #1590, opened 2020-05-29
feat: add the possibility to show the value over the bar on the Chart component


TahimiLeonBravo:
It should be looks like the last design
https://zpl.io/V0JRXex

PR #1613, opened 2020-06-06
feat: add the possibility to show the value over the bar on the Chart…

rgah2107:
fix: #1590

- add the possibility to show the value over the bar on the Chart component

@nexxtway/react-rainbow

### DIFF
--- a/src/components/Chart/resolveLabelsOnBars.js
+++ b/src/components/Chart/resolveLabelsOnBars.js
@@ -1,0 +1,26 @@
+/* eslint-disable no-underscore-dangle */
+import ChartJS from 'chart.js';
+import defaultTheme from '../../styles/defaultTheme';
+
+const palette = defaultTheme.palette;
+export default function resolveLabelsOnBars() {
+    const chart = this.chart;
+    const ctx = chart.ctx;
+
+    ctx.font = ChartJS.helpers.fontString(
+        ChartJS.defaults.global.defaultFontSize,
+        ChartJS.defaults.global.defaultFontStyle,
+        ChartJS.defaults.global.defaultFontFamily,
+    );
+    ctx.fillStyle = palette.text.label;
+    ctx.textAlign = 'center';
+    ctx.textBaseline = 'bottom';
+
+    this.data.datasets.forEach((dataset, i) => {
+        const meta = chart.controller.getDatasetMeta(i);
+        meta.data.forEach((bar, index) => {
+            const dataValue = dataset.data[index];
+            ctx.fillText(dataValue, bar._model.x, bar._model.y - 5);
+        });
+    });
+}
--- a/src/components/Chart/resolveLabelsOnBars.js
+++ b/src/components/Chart/resolveLabelsOnBars.js
@@ -1,8 +1,6 @@
 /* eslint-disable no-underscore-dangle */
 import ChartJS from 'chart.js';
-import defaultTheme from '../../styles/defaultTheme';
 
-const palette = defaultTheme.palette;
 export default function resolveLabelsOnBars() {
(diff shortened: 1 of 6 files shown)


src/components/Chart/resolveOptions.js
```diff
@@ -30,7 +35,7 @@ export default function resolveOptions(conditions) {
     };
 
     let options = {
-        maintainAspectRatio,
+        maintainAspectRatio: maintainAspectRatio,
```
LeandroTorresSicilia, line 38: why this change? I think it is better using the object shorthand

src/components/Chart/resolveOptions.js
```diff
@@ -92,15 +105,54 @@ export default function resolveOptions(conditions) {
...
+                                  meta.data.forEach((bar, index) => {
+                                      const dataValue = dataset.data[index];
+                                      ctx.fillText(dataValue, bar._model.x, bar._model.y - 5);
+                                  });
+                              });
+                          }
+                        : null,
+            },
+        };
+    }
```
LeandroTorresSicilia, line 147: we should add test cases for this changes

src/components/Chart/resolveOptions.js
```diff
@@ -1,3 +1,7 @@
+/* eslint-disable func-names */
+/* eslint-disable object-shorthand */
+/* eslint-disable no-underscore-dangle */
```
LeandroTorresSicilia, line 3: we should avoid adding eslint exceptions, only in some situations, keep in mind that they encourage best practices

src/components/Chart/resolveOptions.js
```diff
@@ -92,15 +105,54 @@ export default function resolveOptions(conditions) {
...
+            hover: {
+                animationDuration: disableAnimations
+                    ? 0
+                    : ChartJS.defaults.global.hover.animationDuration,
+            },
+            animation: {
+                duration: disableAnimations ? 0 : ChartJS.defaults.global.animation.duration,
+                onComplete:
+                    showLabelsOnBars && type === 'bar'
+                        ? function() {
```
LeandroTorresSicilia, line 123: instead of using ```function() {}``` here, you can use an arrow function:
```? () => {}```

File: src/components/Chart/resolveOptions.js [hunk 5]
```diff
@@ -92,15 +105,54 @@ export default function resolveOptions(conditions) {
...
+
+                              this.data.datasets.forEach((dataset, i) => {
+                                  const meta = chart.controller.getDatasetMeta(i);
+                                  meta.data.forEach((bar, index) => {
+                                      const dataValue = dataset.data[index];
+                                      ctx.fillText(dataValue, bar._model.x, bar._model.y - 5);
+                                  });
+                              });
+                          }
+                        : null,
```
LeandroTorresSicilia, line 144: we should move this complex function to another file in order to test it in isolation

src/components/Chart/resolveLabelsOnBars.js
```diff
@@ -0,0 +1,26 @@
+/* eslint-disable no-underscore-dangle */
+import ChartJS from 'chart.js';
+import defaultTheme from '../../styles/defaultTheme';
+
+const palette = defaultTheme.palette;
```
yvmunayev, line 5: You can't just use the defaultTheme, I think a possible solution is to pass the parameter palette to this function when you call it in resolveOptions.js

rgah2107: I tried that at first but the function loses the instance of the chart when I pass some parameter, however check my next commit, I made a change regarding this, anyway tell me if you see a better solution

Issue #1607, opened 2020-06-04
feat: implement ButtonHelp component


TahimiLeonBravo:
#### Requirements
- [ ] Review accessibility issues
https://www.w3.org/WAI/WCAG21/Understanding/content-on-hover-or-focus.html
- [ ] Define API
https://docs.google.com/document/d/10XYsBDGFRhOwb01vQXQfYyCfGNVTstrah2p5vwCT3_4/edit?usp=sharing
- [ ] Implement component
- [ ] Add tests

https://www.w3.org/TR/wai-aria-practices-1.1/#tooltip

#### Design
https://zpl.io/VkRBvpW

PR #1627, opened 2020-06-12
feat: implement HelpText component

yvmunayev:
fix: #1607 

- feat: implement ButtonHelp component

@nexxtway/react-rainbow


src/components/HelpText/index.js
```diff
@@ -0,0 +1,127 @@
...
+    /** A CSS class for the outer element, in addition to the component's base classes. */
+    className: PropTypes.string,
+    /** An object with custom style applied to the outer element. */
+    style: PropTypes.object,
+    /** Displayed the title of component. */
+    title: PropTypes.node,
+    /** Displayed the help message */
+    text: PropTypes.node.isRequired,
+    /** The icon to customize the button. If undefined it is used by default based on the variant. */
+    icon: PropTypes.node,
```
TahimiLeonBravo, line 109: we are not add the`icon` prop right now

src/components/HelpText/index.js
```diff
@@ -0,0 +1,127 @@
...
+                ariaLabelledby={helpTextId}
+            >
+                {icon}
+            </StyledButton>
+            <InternalOverlay
+                isVisible={isOpen}
+                render={() => {
+                    return (
+                        <StyledTooltip id={helpTextId} role="tooltip">
+                            <RenderIf isTrue={!!title}>
```
TahimiLeonBravo, line 79: The `title` always should be a `string` type, if the title is passed we render the text with the icon and if is not passed return null.

src/components/HelpText/styled/index.js
```diff
@@ -0,0 +1,47 @@
+import styled from 'styled-components';
+import attachThemeAttrs from '../../../styles/helpers/attachThemeAttrs';
+
+export const StyledButton = styled.button`
```
TahimiLeonBravo, line 4: I think we should use our `Button` component,
@reiniergs and @LeandroTorresSicilia whats do you think?

src/components/HelpText/styled/index.js
```diff
@@ -0,0 +1,47 @@
...
+    border: none;
+    background: transparent;
+
+    &:focus,
+    &:active {
+        outline: 0;
+    }
+`;
+export const StyledTooltip = attachThemeAttrs(styled.div)`
+    width: 244px;
```
TahimiLeonBravo, line 14: we should not add a fixed width, the user should control the width of the content they want to show, instead maybe we can use min-width

src/components/HelpText/styled/index.js
```diff
@@ -0,0 +1,47 @@
...
+    color: ${props => props.variantColor};
+    align-items: center;
+`;
+
+export const StyledIconContainer = styled.span`
+    margin-right: 8px;
+`;
+
+export const StyledText = attachThemeAttrs(styled.div)`
+    color: ${props => props.palette.text.main}
```
TahimiLeonBravo, line 46: also, we can add a `font-size` and `line-height`

src/components/HelpText/index.js
```diff
@@ -0,0 +1,178 @@
...
+            event.preventDefault();
+            closeOverlay();
+        }
+    };
+
+    const icon = iconMap[variant] || iconMap.info;
+    const inverseIcon = inverseIconMap[variant] || inverseIconMap.info;
+
+    return (
+        <div id={id} className={className} style={style}>
```
TahimiLeonBravo, line 106: I think we can't need to use a `div` wrapper here, instead we can use a Fragment `<>` and add the `className` `style` and `id` to the `StyledButton` element

src/components/HelpText/index.js
```diff
@@ -0,0 +1,181 @@
...
+    question: <QuestionInverseIcon />,
+    info: <InfoInverseIcon />,
+    error: <ErrorInverseIcon />,
+    warning: <WarningInverseIcon />,
+};
+
+/**
+ * HelpText is a popup that displays information related to an element.
+ */
+const HelpText = props => {
```
LeandroTorresSicilia, line 43: to be consistent use function

Issue #1628, opened 2020-06-13
refactor: useOutsideClick hook


TahimiLeonBravo:
#### Requirements
- [x] Refactor `useOutsideClick`  hooks
- [x] Refactor all components using `useOutsideClick hooks`
- [ ] Add documentation for the `useOutsideClick hooks`

```
const ref = userRef();
const [isOpen, setIsOpen] = useState(false);
...
// variant # 1
useOutsideClick(
    ref,
    () => setIsOpen(false),
    isOpen
);

// variant # 2
useOutsideClick(ref, () => setIsOpen(false));
```

PR #1637, opened 2020-06-16
refactor: useOutsideClick hook

yvmunayev:
fix: #1628 

- refactor: refactor useOutsideClick hooks
- refactor: refactor all components using useOutsideClick hooks

@nexxtway/react-rainbow

### DIFF
--- a/src/libs/hooks/useOutsideClick.js
+++ b/src/libs/hooks/useOutsideClick.js
@@ -1,8 +1,10 @@
-import { useEffect, useState, useCallback } from 'react';
+import { useEffect, useCallback } from 'react';
 
-export default function useOutsideClick(ref, handler) {
-    const [isListening, setIsListening] = useState(false);
-
+/**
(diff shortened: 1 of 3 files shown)


src/libs/hooks/useOutsideClick.js
```diff
@@ -1,8 +1,10 @@
...
-export default function useOutsideClick(ref, handler) {
-    const [isListening, setIsListening] = useState(false);
+import { useEffect, useCallback } from 'react';
 
+/**
+ * This component implements the positioning of a component (inserted in the DOM at the body level) based on a trigger DOM element. By the way of example, you can think of the use case of a Menu Options, Tooltip, Popup that should be floating on top of all the elements and it should be correctly positioned based on the component/element that triggers the show/open action.
+ * @category Internal
+ */
+export default function useOutsideClick(ref, handler, isListening = true) {
     const listener = useCallback(
```
reiniergs, line 8: it should depend on the handler, because it will be use like
```js
useOutsideClick(ref, () => {
}, isOpen);
```
then the handlers will be different every time the function executes but in reality, it will be the same

reiniergs: same in use effect

reiniergs:
```js
event => {
    if (ref && isDomElement(ref.current) && !ref.current.contains(event.target)) {
        return handler(event);
    }
}
```

Issue #1626, opened 2020-06-12
feat: allow add plugins to the Chart and Dataset components


TahimiLeonBravo:
### Specs
```
import datalabels from 'chartjs-plugin-datalabels';

const plugins = { datalabels };

<Chart plugins={plugins} datalabels={ { color: '#AAA' } }>
    // customization at the dataset level
    <Dataset datalabels={ { color: '#CCC' } } /> 
</Charts>
```

#### Requirements
-[ ] Implement new feature
- [ ] add tests

PR #1638, opened 2020-06-17
feat: allow add plugins to the Chart and Dataset components

rgah2107:
fix: #1626

- allow add plugins to the Chart and Dataset components

@nexxtway/react-rainbow


package.json
```diff
@@ -33,6 +33,7 @@
     "dependencies": {
         "autosize": "^4.0.2",
         "chart.js": "2.7.3",
+        "chartjs-plugin-datalabels": "^0.7.0",
```
reiniergs, line 36: this dependency should be only for examples purposes, move it to devDependency

src/components/Chart/index.d.ts
```diff
@@ -17,6 +17,8 @@ export interface ChartProps extends BaseProps {
     disableXAxisTickLabels?: boolean;
     disableYAxisTickLabels?: boolean;
     maintainAspectRatio?: boolean;
+    plugins?: Array<any>;
+    datalabels?: object;
```
reiniergs, line 21: the plugins feature is for any possible plugin, right? which means there will be more props depending on the plugins passed in `plugins` prop, I would say we need to find some dynamic manner of defining those props. 

i.e. a plugin named `foo` will receive it's config data through `foo` prop
```js
import foo from 'chart-plugin-foo';

<Chart plugins={{ foo }} foo={fooConfigAtChartLevel}>
    <Dataset foo={fooConfigAtDatasetLevel}>
</Chart>
```

src/components/Chart/index.js
```diff
@@ -46,7 +46,10 @@ export class Chart extends Component {
     }
 
     renderChart() {
-        const { type, labels, ...conditions } = this.props;
+        const { type, labels, plugins, ...conditions } = this.props;
+        if (plugins) {
+            ChartJS.plugins.unregister(plugins);
```
reiniergs, line 51: why we have to unregister the plugins?? I know the labels one needs to be unregistered ... but I guess that's not the case for all the plugins, right?

Again this feature should a mechanism to inject any possible chart plugin

src/components/Chart/index.js
```diff
@@ -55,6 +59,7 @@ export class Chart extends Component {
                 labels,
                 datasets: this.datasets,
             },
+            plugins: plugins || null,
             options: resolveOptions({ type, ...conditions }),
```
yvmunayev, line 63: you should add the plugins variable here, because with the line

`const { type, labels, plugins, ...conditions } = this.props;`

conditions no longer contains plugins

src/components/Chart/readme.md
```diff
@@ -888,4 +888,65 @@ const containerStyles = {
...
+const pluginsChartConf = {datalabels: datalabelsLevelChart, deferred: deferredChart};
+
+const datalabelsLebelDataSet_0 = {color: '#ff6837'};
+const datalabelsLebelDataSet_2 = {color: '#01b6f5'};
+const pluginsDatasetConf_0 = {datalabels: datalabelsLebelDataSet_0};
+const pluginsDatasetConf_2 = {datalabels: datalabelsLebelDataSet_2};
+
+const containerStyles = {
+    maxWidth: 600,
+};
```
yvmunayev, line 916: check lint

yvmunayev, line 916: check lint rules

src/components/Chart/index.js
yvmunayev, line 63: you should add the plugins variable here, because with the line
`const { type, labels, plugins, ...conditions } = this.props;`
conditions no longer contains plugins

src/components/Chart/index.js
yvmunayev: I was looking at this.

The problem of globally registered plugins is solved like this:

`ChartJS.plugins.clear();`

this must be done in everything always, because when it is done

```
import datalabels from 'chartjs-plugin-datalabels';
import deferred from 'chartjs-plugin-deferred';
```

All ChartJS already have that plugin registered.

yvmunayev: all global plugins can be used inline

`ChartJS.plugins.clear();`

is the way to keep ChartJS plugins isolated

src/components/Chart/index.js
```diff
@@ -49,7 +48,7 @@ export class Chart extends Component {
     renderChart() {
         const { type, labels, plugins, ...conditions } = this.props;
         if (plugins) {
-            ChartJS.plugins.unregister(getGlobalsPlugins(plugins));
+            ChartJS.plugins.clear();
```
yvmunayev, line 51: this must always be done in order to isolate the component so that the global plugins do not affect it.

src/components/Chart/index.d.ts
```diff
@@ -17,6 +17,9 @@ export interface ChartProps extends BaseProps {
     disableXAxisTickLabels?: boolean;
     disableYAxisTickLabels?: boolean;
     maintainAspectRatio?: boolean;
+    plugins?: Array<object>;
+    pluginsChartConf?: object;
+    pluginsDatasetConf?: object;
```
yvmunayev, line 22: remove this props from the API. the component has to be like this:

https://github.com/nexxtway/react-rainbow/issues/1626

src/components/Chart/helpers/getPluginsConf.js
```diff
@@ -0,0 +1,10 @@
+export default function getPluginsConf(rest, plugins) {
+    const pluginsConf = {};
+    const keys = plugins.map(plugin => plugin.id);
+    keys.forEach(key => {
+        if (rest[key]) {
+            pluginsConf[key] = rest[key];
+        }
+    });
```
yvmunayev, line 8: that's easier

```
const pluginsConf = {};
    plugins.forEach(plugin => {
        const { id } = plugin;
        if (rest[id]) {
            pluginsConf[id] = rest[id];
        }
    });
```

src/components/Chart/index.js
```diff
@@ -46,7 +47,8 @@ export class Chart extends Component {
     }
 
     renderChart() {
-        const { type, labels, ...conditions } = this.props;
+        unregisterGlobalPLugins(ChartJS);
```
maxxgreene, line 50: `unregisterGlobalP(L)ugins` => `unregisterGlobalPlugins`

maxxgreene:
@rgah2107 @yvmunayev it looks good to me, please change the typo in other PR, I will merge it

Issue #1618, opened 2020-06-07
docs: add interactiveExamples to DatePickerModal component


TahimiLeonBravo:
#### Requirements
- [ ] It should look like the design
https://zpl.io/bAJkN88

PR #1644, opened 2020-06-22
docs: add interactive Examples to DatePickerModal

amontalvof:
fix: #1618 

-

@nexxtway/react-rainbow


src/components/DatePicker/readme.md
```diff
@@ -240,5 +240,76 @@ const containerStyles = {
...
+    <div className="rainbow-flex rainbow-align_center rainbow-m-vertical_medium">
+        <div>
+            <StyledAvatar src="images/user/avatar-5.svg" assistiveText={name} title={name} />
+        </div>
+        <div>
+            <Name>Ana Doe</Name>
+            <Email>ana@gmail.com</Email>
+        </div>
+    </div>
+    <div className="rainbow-m-vertical_medium" style={{ width: '400px' }}>
```
TahimiLeonBravo, line 300: We should avoid use styles inline, instead use a const

src/components/DatePicker/readme.md
```diff
@@ -240,5 +240,76 @@ const containerStyles = {
...
+            id="datePicker-1"
+            placeholder="Select your DOB"
+            value={state.date}
+            onChange={value => setState({ date: value })}
+            label="Birthday"
+            formatStyle="medium"
+            locale={state.locale.name}
+        />
+    </div>
+    <div className="rainbow-m-vertical_medium" style={{ width: '400px' }}>
```
TahimiLeonBravo, line 311: same ^^

src/components/DatePicker/readme.md
```diff
@@ -240,5 +240,76 @@ const containerStyles = {
...
+        <DatePicker
+            id="datePicker-1"
+            placeholder="Select your DOB"
+            value={state.date}
+            onChange={value => setState({ date: value })}
+            label="Birthday"
+            formatStyle="medium"
+            locale={state.locale.name}
```
LeandroTorresSicilia, line 308: since this locale is fixed and it will not change then we can put it here directly instead of using a state:
```locale="'en-US"```
then remove it from initialState

File: src/components/DatePicker/readme.md [hunk 4]
```diff
@@ -61,7 +61,7 @@ initialState = { date: new Date() };
         label="DatePicker Label"
         onChange={value => setState({ date: value })}
     />
-</div>
+</div>;
```
LeandroTorresSicilia, line 64: I think all these semicolons after the div is not needed

src/components/DatePickerModal/readme.md
```diff
@@ -1,72 +1,381 @@
...
+                    .getDate()
+                    .toString()
+                    .padStart(2, '0') +
+                '/' +
+                dates[0].getFullYear();
+            setState({ dates: rangeDate });
+        }
+    }
+    setState({ isOpen: false });
+};
```
LeandroTorresSicilia, line 265: this is too complex, I think we don't need all this

src/components/DatePickerModal/readme.md
```diff
@@ -1,72 +1,381 @@
...
+                    .getDate()
+                    .toString()
+                    .padStart(2, '0') +
+                '/' +
+                endDate.getFullYear();
+
+            setState({ selection: { name: 'This Week', label: startDate + ' - ' + endDate } });
+        }
+    }
+}
```
LeandroTorresSicilia, line 88: this is too complex, I think we don't need all this

src/components/DatePickerModal/readme.md
```diff
@@ -1,72 +1,381 @@
...
+                    .getDate()
+                    .toString()
+                    .padStart(2, '0') +
+                '/' +
+                dates[0].getFullYear();
+            setState({ selection: { name: 'Custom', label: rangeDate } });
+        }
+    }
+    setState({ isOpen: false });
+}
```
LeandroTorresSicilia, line 128: same ^^

src/components/DatePickerModal/readme.md
```diff
@@ -1,72 +1,381 @@
...
+                        <MenuItem
+                            label="United State"
+                            onClick={() => {
+                                setLocation('United State');
+                            }}
+                        />
+                    </ButtonMenu>
+                </div>
+            </Column>
+            <Vl />
```
TahimiLeonBravo, line 309: let's avoid the use of abbreviations, we need to make the code more descriptive and readable

src/components/DatePickerModal/readme.md
```diff
@@ -1,72 +1,381 @@
...
+`;
+
+const Column = styled.div`
+    min-width: 30%;
+    padding: 0.9rem;
+`;
+
+const Vl = styled.div`
+    margin: 1rem 0;
+    border-left: 2px solid ${props => props.theme.rainbow.palette.text.header};
```
TahimiLeonBravo, line 186: use `...border.divider`, we should use the text colors only for texts elements

src/components/DatePicker/readme.md
```diff
@@ -96,7 +96,7 @@ initialState = { date: new Date() };
             onChange={value => setState({ date: value })}
         />
     </div>
-</div>
+</div>;
```
LeandroTorresSicilia, line 99: let's remove this semicolons

src/components/DatePickerModal/readme.md
```diff
@@ -1,72 +1,312 @@
...
+};
+
+<div className="rainbow-m-bottom_xx-large rainbow-p-bottom_xx-large">
+    <GlobalHeader src="images/user/user3.jpg">
+        <div className="rainbow-flex rainbow-align_right">
+            <Picklist
+                value={state.selection}
+                style={containerStyles}
+                placeholder="Select Date"
+                onChange={value => setDates(value)}
```
LeandroTorresSicilia, line 57: here you can use:
```onChange={setDates}```

File: src/components/DatePickerModal/readme.md [hunk 12]
```diff
@@ -1,72 +1,312 @@
...
+        } else if (value.name === 'Yesterday') {
+            setState({ selection: { name: 'Yesterday', label: 'Yesterday' } });
+        } else if (value.name === 'This Week') {
+            setState({ selection: { name: 'This Week', label: 'This Week' } });
+        }
+    }
+}
+
+function formatDates(date) {
+    if (date) {
```
LeandroTorresSicilia, line 34: use dates instead of date

src/components/DatePickerModal/readme.md
```diff
@@ -1,72 +1,312 @@
...
+        if (date.length > 1) {
+            startDay = new Intl.DateTimeFormat().format(date[0]);
+            endDay = new Intl.DateTimeFormat().format(date[1]);
+            formatedDate = startDay + ' - ' + endDay;
```
LeandroTorresSicilia, line 38: we should use const, otherwise, it will pollute the global scope

src/components/DatePickerModal/readme.md
```diff
@@ -1,72 +1,312 @@
...
+        } else {
+            formatedDate = new Intl.DateTimeFormat().format(date[0]);
+        }
+        return formatedDate;
+    }
+}
```
LeandroTorresSicilia, line 44: I think you can do something like:
```
const startDay = new Intl.DateTimeFormat().format(date[0]);
if (date.length > 1) {
    const endDay = new Intl.DateTimeFormat().format(date[1]);
    return startDay + ' - ' + endDay;
}
return startDay;
```

src/components/DatePickerModal/readme.md
```diff
@@ -1,72 +1,312 @@
...
+                      selection: { name: 'Custom', label: formatDates(value) },
+                      title: formatDates(value),
+                      isOpen: false,
+                  })
+                : setState({
+                      value,
+                      selection: { name: 'Custom', label: formatDates(value) },
+                      title: formatDates(value),
+                  });
+        }}
```
LeandroTorresSicilia, line 85: it is more readable if we use:
```onChange={handleDatePickerChange}```
and then above:
```
function handleDatePickerChange(value) {
    if (value.length > 1) {
       setState({
                      value,
                      selection: { name: 'Custom', label: formatDates(value) },
                      title: formatDates(value),
                      isOpen: false,
                  });
    } else {
        setState({
                      value,
                      selection: { name: 'Custom', label: formatDates(value) },
                      title: formatDates(value),
                  });
    }
}
```

File: src/components/DatePickerModal/readme.md [hunk 16]
```diff
@@ -1,72 +1,312 @@
...
+            </Picklist>
+        </div>
+    </GlobalHeader>
     <DatePickerModal
         title={state.title}
         isOpen={state.isOpen}
-        value={getDate(state.value)}
+        variant="double"
+        value={getDates(state.value)}
         selectionType={state.selectionType}
```
LeandroTorresSicilia, line 71: let's use selection type fixed like:
```selectionType="range"```
then you don't need to maintain this state also remove the getDates and you can pass value directly from state

File: src/components/DatePickerModal/readme.md [hunk 17]
```diff
@@ -1,72 +1,312 @@
...
-            label="Click to select a date"
-            onClick={() => setState({ isOpen: true, selectionType: 'single', value: null, title: 'Select a date' })} />
-    </div>
-    <div className="rainbow-p-horizontal_medium">
-        <Button
-            label="Click to select a range of dates"
-            onClick={() => setState({ isOpen: true, selectionType: 'range', value: null, title: 'Select range of dates' })} />
-    </div>
+function setDates(value) {
+    if (value) {
```
LeandroTorresSicilia, line 15: sometimes we need to do this kind of checks but in this case, the Picklist always will send a value in the onChange then there is no need to do this check here

src/components/DatePickerModal/readme.md
```diff
@@ -1,72 +1,312 @@
...
+                value: null,
+                title: 'Select a Date',
+            });
+        } else if (value.name === 'Today') {
+            setState({ selection: { name: 'Today', label: 'Today' } });
+        } else if (value.name === 'Yesterday') {
+            setState({ selection: { name: 'Yesterday', label: 'Yesterday' } });
+        } else if (value.name === 'This Week') {
+            setState({ selection: { name: 'This Week', label: 'This Week' } });
+        }
```
LeandroTorresSicilia, line 29: I think you should only care about the custom since other values come from picklist as they should be stored:
```
if (value.name === 'Custom') {
           return setState({
                isOpen: true,
                value: null,
                title: 'Select a Date',
            });
}
return setState({ selection: value });
```

src/components/DatePickerModal/readme.md
```diff
@@ -1,72 +1,312 @@
...
+    DatePickerModal,
+    ButtonIcon,
+    ButtonGroup,
+    ButtonMenu,
+    MenuItem,
+} from 'react-rainbow-components';
+import styled from 'styled-components';
+
+const Row = styled.div`
+    /*background-color: ${props => props.theme.rainbow.palette.background.main};*/
```
LeandroTorresSicilia, line 105: remove commented style

src/components/DatePickerModal/readme.md
```diff
@@ -1,72 +1,312 @@
...
+function formatDates(date) {
+    if (date) {
+        if (date.length > 1) {
+            startDay = new Intl.DateTimeFormat().format(date[0]);
+            endDay = new Intl.DateTimeFormat().format(date[1]);
+            formatedDate = startDay + ' - ' + endDay;
+        } else {
+            formatedDate = new Intl.DateTimeFormat().format(date[0]);
+        }
+        return formatedDate;
```
LeandroTorresSicilia, line 181: same ^^

src/components/DatePickerModal/readme.md
```diff
@@ -1,72 +1,312 @@
...
+                      dates: formatDates(value),
+                      title: formatDates(value),
+                      isOpen: false,
+                  })
+                : setState({
+                      value,
+                      dates: formatDates(value),
+                      title: formatDates(value),
+                  });
+        }}
```
LeandroTorresSicilia, line 308: same ^^

src/components/DatePickerModal/readme.md
```diff
@@ -1,72 +1,312 @@
...
+                ? setState({
+                      value,
+                      dates: formatDates(value),
+                      title: formatDates(value),
+                      isOpen: false,
+                  })
+                : setState({
+                      value,
+                      dates: formatDates(value),
+                      title: formatDates(value),
```
LeandroTorresSicilia, line 306: also here you should call formatDates only once and store it in a const

Issue #481, opened 2018-12-26
fix: the badge of the GitHub stars is render on the page of each component


TahimiLeonBravo:
- Current behavior:
When change between pages in the library (e.g when you are in components page and select a specific component) the GitHub Stars component (`library/styleguideComponents/SectionHeading/GithubStarsWrapper/index.js`) is recreated and thus it fetches the stars again.
- Expected behavior:
It should fetch GitHub stars only once (when the library first render)

MarcosVillacanas:
Hi! Is this still causing trouble? If so, Could you please give me more context of the problem?

LeandroTorresSicilia:
@MarcosVillacanas I just added a description of the bug above
Hint: since this library is based on react-styleguidist, their SectionHeading component is rerendered each time when you navigate, there is a RibbonRenderer component which is meant for this, then we must replace this component with ours.

LeandroTorresSicilia:
We have this already done here
https://github.com/reiniergs/react-rainbow-styleguide

MarcosVillacanas:
Okay, I'm gonna take it :)
I will let you know as soon as I have got something

MarcosVillacanas:
I think I have solved the issue: https://github.com/nexxtway/react-rainbow/pull/1653
Could you please check it?

PR #1653, opened 2020-06-25
fix: github Stars badge is now rendered just once

MarcosVillacanas:
fix: Github Stars badge is now rendered just once
issue: #481 

- Section Heading was included on SectionRenderer, which led to Github Stars Badge renderization each time the user navigated through the webpage.
- Now, the header is used before any page renders its content, getting the same result as before but avoiding continuously rendering.
- It was just a matter of how the header component header was called, inside SectionRenderer caused the problem, now it is treated as a normal component, which solves the issue. 
- Parameters received by SectionHeader (only from SectionRenderer, and not from ComponentsPage, which has to be an error), were not used and that's why there is no conflict removing them.
- I have manually tested my approach to check it solves the problem.

- [ x ] I have followed (at least) the [PR section of the contributing guide](https://github.com/nexxtway/react-rainbow/blob/master/CONTRIBUTING.md#submitting-a-pull-request).

@nexxtway/react-rainbow


MarcosVillacanas:
There are three commits because:
-I thought there was a format error due to the commit error, so reverted the commit which first letter was capitalized
-Just after that, I realized the problem was the PR name, it is correctly formatted now.

LeandroTorresSicilia:
The GitHub stars continue rerendering when you are in a specific component page like: `/#/Accordion` and then go back to all components page (`/#/Components`) and then go to another component page like `/#/Button`.
What we need to do is substitute the RibbonRenderer from styleguidist like we do in https://github.com/reiniergs/react-rainbow-styleguide
Take a look at the `styleguide.config.js` file in the root, there is a block:
```
    styleguideComponents: {
        StyleGuideRenderer: path.join(__dirname, 'library/styleguideComponents/StyleGuide'),
        SectionHeading: path.join(__dirname, 'library/styleguideComponents/SectionHeading'),
        ReactComponentRenderer: path.join(__dirname, 'library/styleguideComponents/ReactComponent'),
        TableRenderer: path.join(__dirname, 'library/styleguideComponents/PropsTable'),
        PlaygroundRenderer: path.join(__dirname, 'library/styleguideComponents/Playground'),
        TabButtonRenderer: path.join(__dirname, 'library/styleguideComponents/CodeEditorButton'),
        ToolbarButtonRenderer: path.join(__dirname, 'library/styleguideComponents/ToolbarButton'),
        TableOfContents: path.join(__dirname, 'library/styleguideComponents/TableOfContents'),
        ComponentsListRenderer: path.join(__dirname, 'library/styleguideComponents/ComponentsList'),
        Editor: path.join(__dirname, 'library/styleguideComponents/Editor'),
        PathlineRenderer: path.join(__dirname, 'library/styleguideComponents/Pathline'),
        Wrapper: path.join(__dirname, 'library/styleguideComponents/Wrapper'),
        SectionRenderer: path.join(__dirname, 'library/styleguideComponents/SectionRenderer'),
        Examples: path.join(__dirname, 'library/styleguideComponents/Examples'),
    },
```
here we need to use:
```RibbonRenderer: path.join(__dirname, './components/RibbonRenderer'),```
then we can substitute the RibbonRenderer with the our, this component should have the desired behavior since it will render only once

MarcosVillacanas:
Thank you for your comment and explanation :)
Should I completely change SectionHeading to RibbonRenderer? 
Or do they have to co-exist?

In my opinion, it is a complete change, but I would like to have your confirmation

Last question, Can I use useEffect hook for the purpose of just rendering once inside RibbonRenderer? Or should I keep it consistent and use componentDidMount?

Thank you again, I am learning a lot with this issue, perfect for beginners.

LeandroTorresSicilia:
Yes, you are right I think it is a complete change, note that there is no need to use header tag anymore since now it will be the right side of the header and not a header itself.
About using hooks, it is great, you can use it, right now we are using hooks for everything, note that this code was written some years ago.

MarcosVillacanas:
I have removed SectionHeading as we discussed yesterday, and added your RibbonRenderer instead. I have just changed the name to RibbonHeader, and some HTML and CSS lines to keep it consistent. I have also added it to styleguide.config.js.

Just changing the components wasn't enough to avoid multi-rendering, and I have had to keep the main idea of my first approach loading the Header in the main div of styleGuide index.

I have tested it going from Accordion to Components and back to Button, and it works.

LeandroTorresSicilia:
Sorry, for some missing explanations.
We must maintain RibbonRenderer inside library/styleguideComponents folder, it is a simple substitute for the RibbonRenderer of react-styleguidist, when you put it in the `styleguide.config.js` file you are telling styleguidist that you will use your custom RibbonRenderer instead of default one. Visit https://react-styleguidist.js.org/ to understand what styleguidist is.
`src/components` folder is where we put the components that we build as part of the library to others use them as an npm package and RibbonRenderer is not a library component, but a styleguidist component that only is internal for make the UI layout, it is not meant to be a component to be used when you use our react-rainbow-components npm package.
Let me know if you need more context.

LeandroTorresSicilia:
Keep in mind that you need to use RibbonRenderer name and you will not need to use it in the main div of styleGuide index.

MarcosVillacanas:
Okay, now I understand what you wanted to tell me, I have relocated RibbonRenderer to use it as a substitution of the original one.
But at this point, the stars barge keeps rendering every time user navigate through Components Page. Just let me know if this is where I should be to keep looking for the solution.

LeandroTorresSicilia:
Hi, @MarcosVillacanas I made a commit with some fixes that achieve the desired behavior, you can take a look here:
https://github.com/nexxtway/react-rainbow/pull/1653/commits/34aaf1c01926e76d4a2ce812478aa3dd04753444
Thanks for your contribution

Issue #1645, opened 2020-06-23
fix: improve the MultiSelect component


TahimiLeonBravo:
#### Requirements
- [x] Fix UI when readOnly
- [x] Some improvements
- [x] Add Tests

#### Design
https://zpl.io/awROBBK

#### Specs
- [x] it should not show the cancel icons on the Chips when the component is `readOnly`
- [x] it should not show the plus button when the component is `readOnly`
- [x] it should not be focusable when the component is `readOnly`
- [x] The `variant` prop should receive the values `default` | `chip` with the `defaultValue` to `default` (the default variant should look like the example 4)
- [x] Add a `isBare` prop `boolean` that should look like the example 6
- [x] When the `variant` is `default` it should render the text only on one line and show `...` when it is larger than the container
- [x] When the `variant` is `default` it should render a counter with the amount selected
- [x] When `hover` or `active` one options selected it should render an `X` instead of the `✔️`
- [x] When `clicking` or press the `enter` key on a selected option it should unselect the option and close the dropdown

#### Tests
(We need to add tests for the new implementation)

PR #1659, opened 2020-06-25
fix: improve the MultiSelect component

HellWolf93:
fix: #1645

- Add isBare prop
- Change variant prop behavior
- Hide buttons when readonly


src/components/MultiSelect/index.js
```diff
@@ -181,20 +199,20 @@ const MultiSelect = React.forwardRef((props, ref) => {
                     <RenderIf isTrue={!shouldRenderChips}>
                         <StyledPlaceholder>{placeholder}</StyledPlaceholder>
                     </RenderIf>
-                    <RenderIf isTrue={shouldRenderChips}>
-                        <Chips value={value} variant={chipVariant} onDelete={handleDelete} />
-                    </RenderIf>
+                    <RenderIf isTrue={shouldRenderChips}>{content}</RenderIf>
```
yvmunayev, line 204: I think I should rethink here because it works but it is difficult to understand. Something like that would be easier to understand.

```
const content = variant === 'default' && getContent(value);
const chips = variant === 'chip' && hasChips(value);
```

```
<RenderIf isTrue={!content && !chips}>
    <StyledPlaceholder>{placeholder}</StyledPlaceholder>
</RenderIf>
<RenderIf isTrue={!!chips}>
    <Chips value={value} variant={chipVariant} onDelete={handleDelete} />
</RenderIf>

<RenderIf isTrue={!!content}><StyledText>{content}</StyledText></RenderIf>
```

src/components/MultiSelect/chips.js
```diff
@@ -3,19 +3,24 @@ import PropTypes from 'prop-types';
...
-        return value.map(val => (
-            <StyledChip
-                key={val.name}
-                label={val.label}
-                variant={variant}
-                onDelete={() => onDelete(val)}
-            />
-        ));
+        return value.map(val => {
+            const onDeleteCallback = disabled || readOnly ? null : () => onDelete(val);
```
yvmunayev, line 13: I think this logic works best in the handleDelete declaration in the parent component. It is only once and you don't have to be passing disable and readonly to this component.

src/components/MultiSelect/index.js
LeandroTorresSicilia: we should be careful here since `!content && !chips` is hard to read because of the negates

LeandroTorresSicilia: also, the use of const like shouldRenderChips indicates explicitly what is the intention

src/components/MultiSelect/index.js
```diff
@@ -181,20 +201,23 @@ const MultiSelect = React.forwardRef((props, ref) => {
...
-                    title="Add"
-                    variant="neutral"
-                    size="small"
-                    icon={<PlusIcon />}
-                    onClick={handleTriggerClick}
-                    disabled={disabled}
-                    ref={triggerRef}
-                    tabIndex="-1"
-                />
+                <RenderIf isTrue={!!value && selectedCount > 0}>
```
LeandroTorresSicilia, line 206: here we should use a const that explains the intention

src/components/MultiSelect/index.js
```diff
@@ -141,8 +146,23 @@ const MultiSelect = React.forwardRef((props, ref) => {
...
+            <Chips
+                value={value}
+                variant={chipVariant}
+                readOnly={readOnly}
+                disabled={disabled}
+                onDelete={handleDelete}
+            />
+        ) : (
+            <StyledText>{getContent(value)}</StyledText>
+        );
```
LeandroTorresSicilia, line 164: this can be another component

src/components/MultiSelect/index.js
LeandroTorresSicilia: what can be done is change shouldRenderChips to shouldRenderContent or something like that since there is not only chips now

src/components/MultiSelect/index.js
```diff
@@ -141,8 +145,22 @@ const MultiSelect = React.forwardRef((props, ref) => {
         dropdownRef,
         handleOutsideClick,
     );
+    useWindowResize(() => setIsOpen(false), isOpen);
 
     const shouldRenderChips = hasChips(value);
+    const tabIndex = disabled || readOnly ? '-1' : tabIndexInProps;
```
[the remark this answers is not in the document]
HellWolf93: you are right, it is just a piece of code left behind after a few changes, right now when disabled or readonly the button is not rendered so there is no need to modify it tabIndex in order to skip focus

src/components/MultiSelect/index.js
```diff
@@ -164,7 +182,7 @@ const MultiSelect = React.forwardRef((props, ref) => {
                 onFocus={onFocus}
                 onBlur={onBlur}
                 onKeyDown={handleKeyDown}
-                tabIndex={tabIndex}
+                tabIndex="-1"
                 ref={comboboxRef}
                 aria-labelledby={labelId}
```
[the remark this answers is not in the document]
HellWolf93: I will remove that tabIndex since it is no longer necessary as the only focusable element in the component is the Add button.

src/components/MultiSelect/index.js
HellWolf93: Done, renamed shouldRenderChips to shouldRenderContent, also moved all other RenderIf conditions to shouldRender* const variables

src/components/MultiSelect/chips.js
HellWolf93: Each chip should call handleDelete with its own value, and the only way of doing it that occurred to me is this way, I will be thankful if you could explain to me how do you think it will be better

src/components/MultiSelect/index.js
HellWolf93: Done

src/components/MultiSelect/index.js
HellWolf93: Done

src/components/MultiSelect/helpers/getContent.js
```diff
@@ -0,0 +1,9 @@
+export default function getContent(value) {
+    if (!value || typeof value !== 'object') {
+        return null;
+    }
+    if (Array.isArray(value)) {
+        return value.map(item => item.label).join(', ');
+    }
+    return value.label;
```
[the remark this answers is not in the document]
HellWolf93: Done

src/components/MultiSelect/index.js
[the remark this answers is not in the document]
HellWolf93: Done.

src/components/MultiSelect/index.js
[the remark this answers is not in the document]
HellWolf93: Done

src/components/MultiSelect/content.js
```diff
@@ -0,0 +1,51 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { getContent } from './helpers';
+import { StyledText } from './styled';
+import Chips from './chips';
+
+const Content = props => {
+    const { variant, value, chipVariant, readOnly, disabled, onDelete } = props;
+    const onDeleteCallback = disabled || readOnly ? null : onDelete;
```
LeandroTorresSicilia, line 9: this logic is here and in the chips twice, then we have it repeated three times, can we put it in only one place?

src/components/MultiSelect/readme.md
```diff
@@ -121,6 +162,71 @@ const MultiSelectExample = props => {
...
+
+const containerStyles = {
+    maxWidth: 400,
+};
+
+const MultiSelectExample = props => {
+    const value = [{
+        label: 'All buildings',
+        name: 'all',
+    }];
```
LeandroTorresSicilia, line 179: you can put this outside

src/components/MultiSelect/chips.js
LeandroTorresSicilia: @HellWolf93 I think whats @yvmunayev meant is to use this logic in parent component and then pass below a prop which could be null or a function then here you not need to use disabled and readOnly and also repeat the logic

src/components/MultiSelect/content.js
HellWolf93: Done

src/components/MultiSelect/readme.md
HellWolf93: Done

src/components/Option/index.js
```diff
@@ -94,9 +95,12 @@ class OptionItem extends Component {
     }
 
     renderRightIcon() {
-        const { name, currentValues, icon, iconPosition } = this.props;
+        const { name, currentValues, icon, iconPosition, activeOptionName } = this.props;
         const hasRightIcon = !!(icon && iconPosition === 'right');
         if (currentValues && currentValues.includes(name)) {
+            if (activeOptionName === name) {
+                return <StyledUncheckIcon />;
+            }
```
LeandroTorresSicilia, line 103: this will affect the piclklist too, how we can avoid this?

LeandroTorresSicilia: I think we could pass down the multiple prop through context and then only render the uncheck icon when is multiple thus simple usages like in Picklist will be not affected

HellWolf93: Done

Issue #1641, opened 2020-06-18
feat: implement MarkdownOutput component


TahimiLeonBravo:
#### Description
- It should be as simple as rendering the on MD the data passed

#### Acceptance Criteria
 - [x] design & review specs (API)
 - [x] implementation
 - [x] unit testing
 - [ ] integration testing
 - [x] docs/examples

#### Refernces
- https://github.com/rexxars/react-markdown

- it should be use the https://remark.js.org/

PR #1662, opened 2020-06-27
feat: implement MarkdownOutput component

wildergd:
fix: #1641 

- add MarkdownOutput component 

@nexxtway/react-rainbow


wildergd:
There are some commented lines ... this is still WIP....

src/components/MarkdownOutput/renderers/styled/index.js
```diff
@@ -0,0 +1,233 @@
...
+    background-color: ${props => props.palette.background.main};
+    border-spacing: 0;
+    width: 100%;
+    box-sizing: border-box;
+    border-radius: 10px;
+    border: 1px solid ${props => props.palette.border.divider};
+`;
+
+export const TableHead = attachThemeAttrs(styled.thead)`
+`;
```
yvmunayev, line 88: check this

src/components/MarkdownOutput/index.d.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export interface MarkdownOutputProps extends BaseProps {
+    id?: string;
+    renderers?: object;
+    linkTarget?: string;
+    transformLinkUri?: () => void;
+    transformImageUri?: () => void;
+    plugins?: object[];
+    parserOptions?: object;
+}
```
yvmunayev, line 12: add source prop

yvmunayev:
Screen Shot 2020-07-01 at 4 00 55 AM
check this

yvmunayev:
markdown
check this

maxxgreene:
@wildergd check this code sandbox I think we can reduce considerably the complexity of the component 
https://codesandbox.io/s/stoic-leaf-sk4ks?file=/src/App.js

wildergd:
I tried that... and yest is quite simple and that's true.. 2 lines of code.. the main problem i found  in `remark-react` is that `options.remarkReactComponents` is not quite well documented so it is not clear what props are passed to react components... 
about markdown abstract-syntax-tree ... there is a lot of documentation about it... you can easily parse it and map element attributes (alignment, heading levels, etc) to component props... renderers are based on component types instead of html tag...

maxxgreene:
keep it simple with `remark-react`, check the `HTML` generated and provide a component for each element, by default `remark-react` will generate plain `HTML` element ... just write styled-components for each element that obey our theme implementation.

wildergd:
Ok.. code refactored... more simple now.. but styling is a bit more complicated now... for example... <li>.... you cannot differentiate if is is a listItem in a simple list.. or an ordered one... or a task list... what if we want to hide the listitem marker on a task list?... the listitem is blind.... it doesn't know the list type... and css does not allow you to style a parent from a child... so you cannot use the checkbox and from there set style on it's parent listitem ot list...  ...

package.json
```diff
@@ -36,7 +36,12 @@
         "clipboard-copy": "^2.0.0",
         "prop-types": "^15.6.2",
         "react-async-script-loader": "^0.3.0",
+        "remark": "^12.0.0",
+        "remark-parse": "^8.0.2",
+        "remark-react": "^7.0.1",
         "styled-components": "^4.3.2",
+        "unified": "^9.0.0",
+        "unist-util-visit": "^2.0.2",
```
reiniergs, line 44: remove `remark` and `unist-util-visit` since we don't need it anymore

src/components/MarkdownOutput/index.d.ts
```diff
@@ -0,0 +1,12 @@
+import { ReactNode } from 'react';
+import { BaseProps } from '../types';
+
+export interface MarkdownOutputProps extends BaseProps {
+    id?: string;
+    renderers?: object;
+    plugins?: object[];
+    parserOptions?: object;
+    source: string;
```
reiniergs, line 9: - source => value
- remove renderers , plugins, parserOptions(we wont allow that for now) 
- source is not required(by default empty string)

src/components/MarkdownOutput/index.js
```diff
@@ -0,0 +1,63 @@
...
+        id,
+        className,
+        style,
+        source: sourceInProps,
+        renderers: customRenderers,
+        parserOptions,
+        plugins,
+    } = props;
+
+    const source = sourceInProps || '';
```
reiniergs, line 17: use defaultProps

src/components/MarkdownOutput/index.js
```diff
@@ -0,0 +1,63 @@
...
+    const result = useMarkdownToReact(source, parserOptions, plugins, {
+        ...defaultRenderers,
+        ...customRenderers,
+    });
```
reiniergs, line 21: it will be just
```js
const result = useMarkdownToReact(value);
```

Issue #1475, opened 2020-04-10
fix: TimePicker am/pm select when clicking on the selected element


TahimiLeonBravo:
https://drive.google.com/drive/u/2/folders/1rZEIq4TxqbFkdyaCu8g-Sh7Zym1dHaBJ

#### Steps to reproduce
1. Go to TimePicker component https://react-rainbow.io/#!/TimePicker/3
2. Click on the TimePicker input
3. Click on the "PM" element

#### Current Behavior
When clicking on the selected element on the "AM-PM" it doesn't close.

#### Expected Behavior
When clicking any element on the "AM-PM" it should close them

Anyrob:
Hi! I sent request access for the added document,
meanwhile, would you mind to giving me more context about it, please?

TahimiLeonBravo:
Hi @Anyrob the bug happens when you try clicking on the selected value on the "am-pm", I am going to add the steps to reproduce the bug and the expected behavior

Anyrob:
Thank you! I would like to work on this one :)

TahimiLeonBravo:
> Thank you! I would like to work on this one :)

Great!! you are welcome!!

Anyrob:
Hi @TahimiLeonBravo I think I solve it :) I'm gonna make my PR soon, but I noticed another bug, should I write it here?

TahimiLeonBravo:
Great! @Anyrob I think you should make a PR with this fix and create a new issue for the new bug :)

Anyrob:
Hey! @TahimiLeonBravo, I'm experiencing some problems with my internet connection but as soon as it is restored I'll make the PR! :)

Anyrob:
Hi @TahimiLeonBravo just to notify you that my internet is better now :), and also, that I was trying to do the integration test before the PR, but I couldn't do it. I already identified where to write (Integration-specs-TimePicker-TimePricker-1.specs.js) and I was analyzing the ones that are there, however, I'm not being able to prove that it's working as expected. I already have a gif that shows the issue being solved, but I don't know if it's enough to show that it works. That's my only blocker to submit the PR, Do you have any recommendations for me?

LeandroTorresSicilia:
Submit the PR without the test, we can create another issue to test it later

PR #1663, opened 2020-06-27
fix: close AM / PM element by selecting

Anyrob:
**Close AM / PM element by selecting**

fix: TimePicker am/pm select when clicking on the selected element #1475

- When selecting an AM/PM option, the code is sending the value update but not the signal to close. The solution for this was to modify in the "AmPmSelect" render an action of "blur" (same as close) for the moment it is clicked, that means, AM / PM has been selected.

@nexxtway/react-rainbow

### DIFF
--- a/src/components/TimePicker/ampmSelect.js
+++ b/src/components/TimePicker/ampmSelect.js
@@ -49,8 +49,7 @@ export default class AmPmSelect extends PureComponent {
     }
 
     render() {
-        const { isFocused } = this.props;
-        const { tabIndex, onFocus, value } = this.props;
+        const { isFocused, tabIndex, onFocus, value, onClick } = this.props;
 
         if (isFocused) {
             return (
@@ -71,6 +70,7 @@ export default class AmPmSelect extends PureComponent {
                         value="AM"
                         checked={this.isInputChecked('AM')}
                         onChange={this.handleOnChange}
+                        onClick={onClick}
                         onBlur={handleAmPmBlur}
                     />
 
@@ -83,6 +83,7 @@ export default class AmPmSelect extends PureComponent {
                         value="PM"
                         checked={this.isInputChecked('PM')}
                         onChange={this.handleOnChange}
+                        onClick={onClick}
                         onBlur={handleAmPmBlur}
                     />
 
@@ -107,6 +108,7 @@ AmPmSelect.propTypes = {
     value: PropTypes.string,
     defaultValue: PropTypes.string,
     onChange: PropTypes.func,
+    onClick: PropTypes.func,
     onFocus: PropTypes.func,
     tabIndex: PropTypes.string,
     isFocused: PropTypes.bool,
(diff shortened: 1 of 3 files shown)


src/components/TimePicker/ampmSelect.js
```diff
@@ -71,6 +77,7 @@ export default class AmPmSelect extends PureComponent {
                         value="AM"
                         checked={this.isInputChecked('AM')}
                         onChange={this.handleOnChange}
+                        onClick={this.handleOnClick}
```
LeandroTorresSicilia, line 80: I think here you can simply use the onClick prop direclty:
```onClick={onClick}```
then there is no need for:
```
this.handleOnClick = this.handleOnClick.bind(this);
...
handleOnClick() {
        const { onClick } = this.props;
        onClick();
    }
```

**Comment by TahimiLeonBravo** [comment]
@Anyrob Good Job!!
Review the CLA because it found an inconsistency with your Github account, you can review this

https://help.github.com/en/github/committing-changes-to-your-project/why-are-my-commits-linked-to-the-wrong-user#commits-are-not-linked-to-any-user

File: src/components/TimePicker/ampmSelect.js (hunk 1 above)
Comment (Anyrob) [reply src/components/TimePicker/ampmSelect.js:80]: Hi, @Leandro. I keep it that way because I was attempting to respect the project structure, but if you rather that way I could change it. Thanks for the comment :)

**Comment by Anyrob** [comment]
Hi @TahimiLeonBravo, I already check the CLA, but it wasn't updating so I made another commit, and this time it works! Please let me know if there's another thing I need to do (:

File: src/components/TimePicker/ampmSelect.js (hunk 1 above)
Comment (LeandroTorresSicilia) [reply src/components/TimePicker/ampmSelect.js:80]: Yes, we normally use handlers in this way but only when doing more things like for example:
```
handleOnClick() {
        const { onClick, disabled } = this.props;
        if (disabled) {
                     onClick();
        }
}
```
in a case like this we use to do like:
```onClick={onClick}```
since it is more simple and also avoids the binding which is costly

Comment (Anyrob) [reply src/components/TimePicker/ampmSelect.js:80]: Thanks for the clarification I'll change it 👍

**Comment by Anyrob** [comment]
@TahimiLeonBravo I made the change suggested by @LeandroTorresSicilia 👍  I think that's all if there is anything else pending please let me know

File: src/components/TimePicker/ampmSelect.js [hunk 2]
```diff
@@ -49,8 +49,7 @@ export default class AmPmSelect extends PureComponent {
     }
 
     render() {
-        const { isFocused } = this.props;
-        const { tabIndex, onFocus, value } = this.props;
+        const { isFocused, tabIndex, onFocus, value, onClick } = this.props;
```
LeandroTorresSicilia, line 52: good job!

Anyrob: Thank you (: 💯

Issue #1652, opened 2020-06-24
test: fix test on the Picklist component


TahimiLeonBravo:
Currently, some test on the Picklist component is disabled. We need all of them working, and add more if is necessary.

###### picklist-1
- [ ] should open menu and not set focus to input when click it
- [ ] should close menu and set focus to input when menu is open and click on input
- [ ] should open menu and lose input focus when it is focused and press any arrow key
- [ ] should close menu when it is open and and press tab key
- [ ] should close menu when click outside it
- [ ] should close menu and set focus to input when menu is open and press esc key
- [ ] should open menu and lose input focus when it is focused and press space key
- [ ] should select Experimental Building option and set focus to input when menu is open and press enter key
- [ ] should open menu and set Experimental Building option active when click on input
- [ ] should select Empire State with keyboard
- [ ] should set active Empire State when press arrow down key
- [ ] should set active Empire State when hover the option
- [ ] should select Empire State when hover the option and press enter key
- [ ] should set active the first option when other is active and close the menu and open it again
###### picklist-3
- [ ] should scroll down to see the next option focused when initially is not visible
- [ ] should scroll up to see the first option
- [ ] should scroll down when hover the down arrow
- [ ] should up down when hover the up arrow
###### picklist-9
- [ ] should select the new option with keyboard after it is added dynamically
- [ ] should select the second option with keyboard after it is added and removed dynamically a new element

PR #1664, opened 2020-06-27
test: fix test on the Picklist component

rgah2107:
fix: #1652

- fix test on the Picklist component

@nexxtway/react-rainbow


integration/specs/Picklist/picklist-1.spec.js
```diff
@@ -11,10 +11,10 @@ const {
 } = require('../../constants');
 
 const PICKLIST = '#picklist-1';
+const PICKLISTMENU = 'div[role="listbox"]';
```
LeandroTorresSicilia, line 14: We can't do this since this is implementation details of the InternalDropdown and is a fragile selector too since the page can have more elements with this selector, then this should be part of the picklist page object, I think we can use the id passed to the InternalDropdow to make this selector unique, if you look at the picklist there is a unique id generated `this.listboxId` that is passed as id to the InternalDropdown and as `aria-controls` to the input, then maybe you can get this `aria-controls` value and init an instance of the internal dropdown page object with this value.
Note that the page object of the InternalDropdown needs to be created and discussed so you need to come with a proposal of this page object

src/components/InternalDropdown/pageObject/index.js
```diff
@@ -0,0 +1,80 @@
+const PagePicklistOption = require('../../PicklistOption/pageObject');
```
LeandroTorresSicilia, line 1: use Option component instead

src/components/Picklist/pageObject/index.js
```diff
@@ -99,63 +67,39 @@ class PagePicklist {
...
     }
 
     /**
-     * Wait until the options menu is open.
+     * Returns a new InternalDropdown page object for the element with the supplied id.
      * @method
      */
-    waitUntilOpen() {
-        browser.waitUntil(() => this.isMenuOpen());
+    getMenu() {
```
LeandroTorresSicilia, line 95: can we use this method as internal order to avoid calling it when need to get an option?
If it will be used internally then you can use Symbol in order to hide this method

Issue #1407, opened 2020-03-10
fix: return the focus to the trigger element when close the Modal component on "Safari" and "Firefox" browsers


TahimiLeonBravo:
https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/dialog.html

Charlitos96:
Hi, I would like to work on this issue 😄

LeandroTorresSicilia:
Great!, you can submit a pull request

Charlitos96:
Hi @TahimiLeonBravo ! I was looking to reproduce the issue, but I can't make it happen. Can you give me more info about how you lost the focus?

**My OS**
Windows 10

**What I do?**
I run on `localhost:6060` the project, on Mozila firefox and the focus of the component `Modal` returns to the button.

Let me know if I need to do something else to reproduce the problem 😄

LeandroTorresSicilia:
Seems this is happening in macOS with Mozilla and Safari

Charlitos96:
Hi, I made a PR with the changes, I had some problems at the moment of making the push because some warnings on the code that I don't made, so let me know if it's ok! 😄

Charlitos96:
Update: I found that `document.activeElement` don't work on iOS, it returns `body` that's the reason why the focus get lost.
We can use `event.target` but that will works through `handleOnClick()` not on the component Modal.
I tried with `document.querySelector()` but don't work for this.
Do you have any suggestions about how to access from `componentDidUptadate`?

LeandroTorresSicilia:
@Charlitos96  I think we should follow this:
https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/dialog.html
note that this modal works perfectly in all browsers, also here you can see its source code
https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/dialog.html

Charlitos96:
Hello,
Following the pattern from the [link](https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/dialog.html), they pass the reference of the element that triggered the Modal using `openModal('dialog1',this)`.
Using that as a guide, I update the method `handleOnClick()` to pass the reference of the button with `event.target` to update `modalTriggerElement` with it.
If this was not the approach you suggested please let me know. 😄

Charlitos96:
Hello, I can't keep working on this, the problem is with the way that Mac handle the focus, not on the code. I hope anyone can find another solution.

PR #1666, opened 2020-06-28
fix: return focus to modal on firefox and safari

Charlitos96:
Return the focus to the trigger button that calls the Modal Component for firefox and safari on MacOS

fix: #1407

- Added reference to the Button that calls the Modal Component.

@nexxtway/react-rainbow


src/components/Modal/readme.md
```diff
@@ -14,13 +14,15 @@ class EmptyModal extends React.Component {
         };
         this.handleOnClick = this.handleOnClick.bind(this);
         this.handleOnClose = this.handleOnClose.bind(this);
+        this.buttonRef = React.createRef();
```
LeandroTorresSicilia, line 17: We don't want to make this always, instead we should find a solution that works out of the box each time we use the modal

src/components/Modal/index.js
```diff
@@ -44,13 +44,12 @@ export default class Modal extends Component {
...
             CounterManager.increment();
             disableBodyScroll(this.contentRef.current);
-            this.modalTriggerElement = document.activeElement;
             this.modalRef.current.focus();
         }
     }
 
     componentDidUpdate(prevProps) {
-        const { isOpen, onOpened } = this.props;
+        const { isOpen, onOpened, modalTriggerElement } = this.props;
```
LeandroTorresSicilia, line 52: we should avoid adding props since this change the component API and this needs to be discussed carefully, we should try to find an internal solution about how to find the active element

Charlitos96: Ok, I will look for another soluton them, thank you

Issue #1667, opened 2020-06-28
fix: the page crash when `Tabset` component have only one Tab


TahimiLeonBravo:
<img width="404" alt="Screen Shot 2020-06-28 at 12 27 36 PM" src="https://user-images.githubusercontent.com/32083556/85954114-cdc98800-b93a-11ea-989f-6f7aeb3eac8c.png">

gerardolgvr:
Hello! Can I try to fix this issue?

TahimiLeonBravo:
> Hello! Can I try to fix this issue?

@gerardolgvr You are welcome :)
Go for it!

gerardolgvr:
> > Hello! Can I try to fix this issue?
> 
> @gerardolgvr You are welcome :)
> Go for it!

Thanks, I think I found the solution to this issue. 

The problem was at the function "IsNotSameChildren" at utils.js. This function is expecting an array of objects but when a single tab is given the parameter that receives that function is an object. So, method .some is available for arrays but when an object is received an error occurs.

I have updated the function to check when an object is passed to the function to make the proper cooperation. 

The tests associated to this component runs without problems :) 

I think its a good moment to make a pr and check if it seems ok for you or if some changes are required 😊

PR #1671, opened 2020-06-29
fix: tabset component with only one tab

gerardolgvr:
### fix: tabset component with only one tab

fix: #1667 

- The problem was at the function "IsNotSameChildren" at utils.js (Tabset component). This function is expecting an array of objects but when a single tab is given the parameter that receives that function is an object. So, method ".some" is available for arrays but when an object is received an error occurs.
- A new check was added to make the proper comparation

@nexxtway/react-rainbow

### DIFF
--- a/src/components/Tabset/utils.js
+++ b/src/components/Tabset/utils.js
@@ -44,19 +44,15 @@ export function getChildrenTotalWidthUpToClickedTab(children, index) {
 }
 
 export function isNotSameChildren(children, prevChildren) {
-    if (!Array.isArray(children)) {
-        if (children && prevChildren) {
-            return children.props.name !== prevChildren.props.name;
-        }
-        return false;
+    if (Array.isArray(children)) {
+        return children.some((child, index) => {
+            if (child && prevChildren[index]) {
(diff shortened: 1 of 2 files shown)


src/components/Tabset/utils.js
```diff
@@ -44,6 +44,13 @@ export function getChildrenTotalWidthUpToClickedTab(children, index) {
 }
 
 export function isNotSameChildren(children, prevChildren) {
+    if (!Array.isArray(children)) {
```
maxxgreene, line 47: I will avoid negating at the beginning of a function. Why? because negations in an `if` do not read well.

try this
```js
if (Array.isArray(children)) {
    return children.some ...
}
return children.props.name === prevChildren.props.name;
```
simple :) 

also, try to write some tests for you fix

gerardolgvr: thanks @maxxgreene . I'm gonna update it

LeandroTorresSicilia:
@gerardolgvr Good job!

gerardolgvr:
> @gerardolgvr Good job!

Thanks! :)

Issue #1156, opened 2019-11-20
fix: RadioButtonGroup and RadioGroup keyboard navigation does not work on firefox and safari


TahimiLeonBravo:
`it is a problem with focus, they do not get focus when clicking and then the keyboard does not work`

TahimiLeonBravo:
@yvmunayev go inside this issue

jonathanurias96:
Hello! May I work in this issue?

TahimiLeonBravo:
> Hello! Can I work in this issue?

Great!! Go for it!

jonathanurias96:
@TahimiLeonBravo  Hello! I noticed this is the styled-components code that fills the radio button when is clicked (or checked):
<img width="534" alt="Screen Shot 2020-06-29 at 1 40 53 PM" src="https://user-images.githubusercontent.com/57416786/86054097-62e98100-ba0e-11ea-962f-718ef77726d0.png">

I would like to add the same css attributes to the same radio button but instead of the checked attribute for reference I would like to add it when the tag has an attribute "aria-checked="true"". 
In regular css its done this way:
<img width="328" alt="Screen Shot 2020-06-29 at 1 46 42 PM" src="https://user-images.githubusercontent.com/57416786/86054499-20747400-ba0f-11ea-85c5-cbf0f76a5445.png">

Do you happen to know how can I do it? or maybe somewhere I can read how to do it?

jonathanurias96:
Hello @TahimiLeonBravo, I already added a PR that fixes this issue. Please review it and if theres anything that needs to be changed please let me know. Also feedback will be appreciated.

PR #1672, opened 2020-06-29
fix: [Added focus method when input is clicked]

jonathanurias96:
fix: #1156

fix: focused element in Firefox and Safari when focusable element is clicked.

@nexxtway/react-rainbow


src/components/Input/inputRadio/radio.js
```diff
@@ -114,7 +114,6 @@ Radio.defaultProps = {
     disabled: false,
     onChange: () => {},
     tabIndex: undefined,
-    onClick: () => {},
```
LeandroTorresSicilia, line 117: we should avoid changing the component API since there are uses that relies on this prop, we can't break it

jonathanurias96: Sorry for that it was giving me a lint error and I supposed it was not needed. However I added a commit. Adding the prop back. Please review.

src/components/Input/inputRadio/radio.js
```diff
@@ -67,7 +67,10 @@ export default class Radio extends Component {
                     tabIndex={tabIndex}
                     onFocus={onFocus}
                     onBlur={onBlur}
-                    onClick={onClick}
+                    onClick={event => {
+                        this.inputRef.current.focus();
+                        onClick(event);
+                    }}
```
LeandroTorresSicilia, line 73: we should create a class method to handle this like:
```
handleClick(event) {
    if (isMozilla() || isSafari()) {
        this.inputRef.current.focus();
    }
    onClick(event);
}

...

onClick={this.handleClick}
```
also we need to add tests to this and note the helpers isMozilla and isSafari should be added to src/libs/utils

src/components/RadioButtonGroup/radioButtonItem.js
```diff
@@ -37,6 +38,10 @@ export default class RadioButtonItem extends Component {
             >
                 <StyledInput
                     type="radio"
+                    ref={this.inputRef}
+                    onClick={() => {
+                        this.inputRef.current.focus();
+                    }}
```
LeandroTorresSicilia, line 44: same put this in a class method and also check for call this only when safari and mozilla

src/components/Input/inputRadio/radio.js
LeandroTorresSicilia: for these helpers take a look here:
https://jsfiddle.net/6spj1059/
only add the needed ones

Issue #700, opened 2019-04-24
fix: fix the Table header when the height is passed


TahimiLeonBravo:

PaulinaQuintero:
Hello! Can I work on this issue?

TahimiLeonBravo:
> Hello! Can I work on this issue?

Great!! Go for it!

PaulinaQuintero:
Thank you! How can I reproduce the issue? It's only in iphone 6 and iphone 8 right? Because I tried to reproduce it in Android 7 and Chrome in Windows 10 and I couldn't. Could you give me more context? please

TahimiLeonBravo:
> Thank you! How can I reproduce the issue? It's only in iphone 6 and iphone 8 right? Because I tried to reproduce it in Android 7 and Chrome in Windows 10 and I couldn't. Could you give me more context? please

I test it on IOS and it is currently happening, I reproduced them on iPhone 6, iPhone 8 Plus, and iPad Pro
I am not sure if it happens on Android but I guess it only happens on IOS

PaulinaQuintero:
I haven't been able to reproduce the error yet, could you tell me how to reproduce it? I don't understand what you mean by height, I'm a little confused

TahimiLeonBravo:
> I haven't been able to reproduce the error yet, could you tell me how to reproduce it? I don't understand what you mean by height, I'm a little confused

You can view the header content is hidden when you use the table on a container with less height than the Table content, you can view that if you open this example https://react-rainbow.io/#!/Table/3 on any iOS device

PaulinaQuintero:
Thank you I can see it now in iphone 6! I will work on it

PR #1673, opened 2020-06-30
fix: [Fixed Header table when using iOS]

PaulinaQuintero:
fix: [Fixed Header table when using iOS]  #700
 fix: #700 
Changes proposed in this PR:
The problem was the position of the table component Scrollable Y. I added a relative position, that makes the header scrollable 

 I have followed (at least) the PR section of the contributing guide.

@nexxtway/react-rainbow

### DIFF
--- a/src/components/Table/styled/scrollableY.js
+++ b/src/components/Table/styled/scrollableY.js
@@ -2,10 +2,12 @@ import styled from 'styled-components';
 import attachThemeAttrs from '../../../styles/helpers/attachThemeAttrs';
 
 const StyledScrollableY = attachThemeAttrs(styled.div)`
-    height: 100%;
-    overflow: hidden;
-    overflow-y: auto;
-    -webkit-overflow-scrolling: touch;
+padding: 0 0 40px 0;
+height: 100%;
+overflow: hidden;
+overflow-y: scroll;
+position:relative;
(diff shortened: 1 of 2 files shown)


TahimiLeonBravo:
@PaulinaQuintero I am reviewing the changes and the issue was resolved but I find some problems with this change:
- the header is not fixed on the top of the Table, only the body should scroll
- the body of the table is very highest
Show the video
https://drive.google.com/file/d/169dnsjsbg1AyUeNgg4HLLtY_wVVApIXR/view?usp=sharing

- I found another problems with the content of the table
Screen Shot 2020-06-30 at 12 07 58 AM
Screen Shot 2020-06-30 at 12 08 36 AM

src/components/Table/styled/scrollableY.js
```diff
@@ -2,10 +2,12 @@ import styled from 'styled-components';
 import attachThemeAttrs from '../../../styles/helpers/attachThemeAttrs';
 
 const StyledScrollableY = attachThemeAttrs(styled.div)`
-    height: 100%;
-    overflow: hidden;
-    overflow-y: auto;
-    -webkit-overflow-scrolling: touch;
+padding: 0 0 40px 0;
```
luislizarraga, line 5: You should keep the indentation consistent

Issue #1683, opened 2020-07-03
fix: Picklist and Input component styles when `readOnly`


TahimiLeonBravo:
#### Requirements
- [ ] It should not open the dropdown when `readOnly` prop is passed
- [ ] It should not show the shadow when `shaded` and `readOnly` is passed
- [ ] It should predominate the `disabled` styles on the Input component when `readOnly` and `disabled` is passed

PR #1684, opened 2020-07-04
fix: styles in the Picklist component when the prop readOnly is passed

rgah2107:
fix: #1683

- styles in the Picklist component when the prop readOnly is passed

@nexxtway/react-rainbow


src/components/Picklist/styled/input.js
```diff
@@ -43,6 +43,7 @@ const PickerInput = attachThemeAttrs(styled(StyledInput))`
         box-shadow: none;
         background-color: transparent;
         border: 2px solid transparent;
+        pointer-events: none;
```
yvmunayev, line 46: The problem with this solution is that it changes the course to the hand that indicates some action and the course should be the text

src/components/Input/label/index.js
```diff
@@ -29,6 +29,7 @@ Label.propTypes = {
     required: PropTypes.bool,
     inputId: PropTypes.string,
     readOnly: PropTypes.bool,
+    disabled: PropTypes.bool,
```
yvmunayev, line 32: To change the API of a component a previous discussion is necessary, this small change has a big impact, you had to change 5 component which in turn many more components depends on them for this reason it is not feasible. I think the best solution would be to add to the RenderIf where the Label is painted the conditions of disabled and readOnly

TahimiLeonBravo:
> Y
> Screen Shot 2020-07-10 at 4 37 26 AM
> ou have to take into account keyboard navigation

we have a bug when use the keyboard navigation, you can review the bug on the video here
https://drive.google.com/file/d/1FisnzjtXsylVphZBOBUdoaOdrcNz4cSz/view?usp=sharing

TahimiLeonBravo:
@rgah2107 the label should be centered on the Picklist component when `readOnly` and `disabled` are passed at the same time
Screen Shot 2020-07-12 at 5 06 16 PM

Issue #1669, opened 2020-06-28
fix: markup and improve accessibility on the Tree component


TahimiLeonBravo:
### Specs
- [x] It should render a `ul` element instead of `li` when an element has one or more children (here the console warns about that one li cannot contains li as children, then we need to find a good solution).
- [x] It should have a `span` element instead `h1` on the label element
- [ ] It should follow the accessibility standards on the markup
- [ ] It should make clickable all items on the tree
- [ ] It should work with the keyboard
- [ ] It should allow the user to do any action when clicking on the tree items

#### Design Guidelines
https://zpl.io/bAoorRB

#### Accessibility Sources
- https://www.w3.org/TR/wai-aria-practices/examples/treeview/treeview-2/treeview-2b.html
-https://www.lightningdesignsystem.com/components/trees/#site-main-content

PR #1703, opened 2020-07-10
fix: accessibility issues on Tree component, adjusting examples

hportales2000:
fix: #1669 

- adding new markup with accessibility attributes
- adding new props

@nexxtway/react-rainbow


src/components/Tree/index.js
```diff
@@ -37,24 +57,36 @@ Tree.propTypes = {
...
     /** The action triggered when the user clicks in the tree node expand or collapse button. */
-    onExpandCollapse: PropTypes.func,
+    onNodeExpand: PropTypes.func,
     /** The action triggered when the user clicks in the tree node checkbox. */
-    onSelect: PropTypes.func,
+    onNodeCheck: PropTypes.func,
+    /** The action triggered when the user clicks in the tree node label. */
+    onNodeSelect: PropTypes.func,
+    /** The tree node name. */
+    selectedNode: PropTypes.string,
```
yvmunayev, line 66: @LeandroTorresSicilia this API changes have been discussed?

LeandroTorresSicilia: @yvmunayev, Yes it was already discussed, these changes are approved

src/components/Tree/child.js
```diff
@@ -19,26 +19,56 @@ export default function Child(props) {
...
+        onNodeSelect,
+        isSelected,
+        name,
+        selectedNode,
     } = props;
     const hasChildren = Array.isArray(children);
     const hasCheckbox = typeof isChecked === 'boolean' || isChecked === 'indeterminate';
     const hasIcon = !!icon;
+    const levelMatch = name.match(/\./g);
+    const ariaLevelValue = name && levelMatch ? levelMatch.length + 1 : 1;
```
LeandroTorresSicilia, line 33: we should move these two lines and put in a helper and then use:
```const getTreeLevel(name)```

File: src/components/Tree/child.js [hunk 3]
```diff
@@ -19,26 +19,56 @@ export default function Child(props) {
...
+    const tabIndex = isSelected ? 0 : -1;
+
+    const handleNodeSelect = event => {
+        event.stopPropagation();
+        onNodeSelect({ name });
```
LeandroTorresSicilia, line 38: here we need to send also the nodePath value

src/components/Tree/child.js
```diff
@@ -19,26 +19,56 @@ export default function Child(props) {
...
-        <ItemContainerLi hasChildren={hasChildren} icon={icon} data-id="node-element-li">
-            <NodeContainer data-id="node-element">
+        <ItemContainerLi
+            id={name}
+            data-id="node-element-li"
+            icon={icon}
+            hasChildren={hasChildren}
+            onClick={handleNodeSelect}
+            role="treeitem"
+            aria-label={hasChildren ? label : undefined}
```
LeandroTorresSicilia, line 53: when discussed we said that this will not be used now, also label can be a node and you can't pass a react node as an attribute to an element

src/components/Tree/child.js
```diff
@@ -19,26 +19,56 @@ export default function Child(props) {
...
+            aria-level={ariaLevelValue}
+            aria-expanded={hasChildren ? isExpanded : undefined}
+            aria-selected={isSelected === true ? isSelected : undefined}
```
LeandroTorresSicilia, line 56: put these conditions in const and use above the markup in order to avoid mix js with jsx

src/components/Tree/child.js
```diff
@@ -67,9 +100,20 @@ Child.propTypes = {
     isLoading: PropTypes.bool,
     icon: PropTypes.node,
     children: PropTypes.array,
-    onExpandCollapse: PropTypes.func,
-    onSelect: PropTypes.func,
+    /** The action triggered when the user clicks in the tree node checkbox. */
+    onNodeCheck: PropTypes.func,
+    /** The action triggered when the user clicks in the tree node expand or collapse button. */
```
LeandroTorresSicilia, line 105: these descriptions are only needed in index.js since they are used to create the documentation but only are taken from there

src/components/Tree/index.js
```diff
@@ -37,24 +57,36 @@ Tree.propTypes = {
...
     /** A CSS class for the outer element, in addition to the component's base classes. */
     className: PropTypes.string,
     /** An object with custom style applied for the outer element. */
     style: PropTypes.object,
     /** The id of the outer element. */
     id: PropTypes.string,
+    /** The label of node with role="tree". */
```
LeandroTorresSicilia, line 73: we need to explain more here, also role="tree" is an implementation detail and they don't know who has this role

src/components/Tree/index.js
```diff
@@ -37,24 +57,36 @@ Tree.propTypes = {
...
+    ariaLabel: PropTypes.string,
+    /** The id of the outer element. */
```
LeandroTorresSicilia, line 75: same ^^

src/components/Tree/readme.md
```diff
@@ -79,10 +90,13 @@
         }
     ];
 
+    const selectedNode = '';
+
     import React, { useState, useEffect } from 'react';
 
     const TreeExample = () => {
         const [data, setData] = React.useState(initialData);
+        const [node, setSelectedNode] = React.useState(selectedNode);
```
LeandroTorresSicilia, line 99: since the import above has useState, then here you can remove `React.`:
```const [node, setSelectedNode] = useState(selectedNode);```
same for the data state

File: src/components/Tree/styled/itemContainerLi.js [hunk 10]
```diff
@@ -8,6 +8,14 @@ const ItemContainerLi = styled.li`
         `
         margin-left: 0px;
     `};
+
+    &:focus {
+        outline: 0;
+    }
+
+    &:focus > div {
+        border-left: #01b6f5 solid 4px;
```
LeandroTorresSicilia, line 17: here we need to use colors from the theme

src/components/Tree/styled/nodeContainer.js
```diff
@@ -4,6 +4,11 @@ const NodeContainer = styled.div`
     display: flex;
     flex-direction: row;
     align-items: center;
+    ${props =>
+        props.isSelected &&
+        `
+        background-color: #e4e4e4;
```
LeandroTorresSicilia, line 10: same ^^ all colors should come from theme

src/components/Tree/treeChildren.js
```diff
@@ -1,17 +1,70 @@
...
+                onNodeSelect={onNodeSelect}
+                selectedNode={selectedNode}
             />
         );
     });
 }
+
+TreeChildren.propTypes = {
+    /** A array with the nodes of the Tree. This is a recursive shape that is used for render the nested nodes passed on children of the array.
+     */
```
LeandroTorresSicilia, line 37: same ^^ descriptions are only needed in index.js

TahimiLeonBravo:
@hportales2000 the styles should look like the zeplin design

- right now it looks this
Screen Shot 2020-07-12 at 12 48 14 AM

- it should look like this
https://zpl.io/anqwg8O
Screen Shot 2020-07-12 at 12 52 15 AM

package.json
```diff
@@ -140,7 +140,7 @@
         "firebase-tools": "6.1.1",
         "glob": "^7.1.3",
         "husky": "^2.2.0",
-        "jest": "^24.9.0",
+        "jest": "^26.1.0",
```
LeandroTorresSicilia, line 143: We need to be careful about this updates since there may be Breakin Changes, I agree this needs to be done, but it should be another task where release notes are reviewed in order to not break our library.

src/components/Tree/expandCollapseButton.js
```diff
@@ -14,22 +14,18 @@ function getIcon(isExpanded) {
...
             <SpinnerContainer>
                 <Spinner variant="brand" size="x-small" />
             </SpinnerContainer>
         );
     }
-    if (hasChildren) {
-        return <Button size="x-small" icon={getIcon(isExpanded)} onClick={onClick} tabIndex={-1} />;
-    }
-    return null;
+    return <Button size="x-small" icon={getIcon(isExpanded)} onClick={onClick} tabIndex={-1} />;
```
LeandroTorresSicilia, line 25: This change implies that the expand collapse button will render always, we should avoid rendering it if not needed.

src/components/Tree/helpers/getNodeLevel.js
```diff
@@ -0,0 +1,9 @@
+const getNodeLevel = ({ name }) => {
+    let levelMatch = [];
+    if (name) {
+        levelMatch = name.match(/\./g);
+    }
+    return name && levelMatch ? levelMatch.length + 1 : 1;
+};
```
LeandroTorresSicilia, line 7: In order to improve readability here we can:
```
const getNodeLevel = ({ name = '' }) => {
    const levelMatch = name.match(/\./g);
    if (levelMatch) {
        return levelMatch.length + 1;
    }
    return 1;
};
```

src/components/Tree/readme.md
```diff
@@ -14,7 +14,24 @@
...
+                            children: [
+                                { label: 'Tree Item' },
+                                { label: 'Tree Item',
+                                children: [
+                                    { label: 'Tree Item' },
+                                    { label: 'Tree Item' },
+                                ],
+                                },
+                            ],
+                            },
```
LeandroTorresSicilia, line 31: fix indentation problems:
```
{ label: 'Tree Item' },
{ label: 'Tree Item' },
{
    label: 'Tree Item',
    children: [
        { label: 'Tree Item' },
        {
            label: 'Tree Item',
            children: [
                { label: 'Tree Item' },
                { label: 'Tree Item' },
            ],
        },
    ],
},
```

src/components/Tree/styled/childrenContainer.js
```diff
@@ -3,21 +3,10 @@ import attachThemeAttrs from '../../../styles/helpers/attachThemeAttrs';
 
 const ChildrenContainerUl = attachThemeAttrs(styled.ul)`
     margin-left: 10px;
-    margin-top: 2px;
-    margin-bottom: 2px;
+    // margin-top: 2px;
+    // margin-bottom: 2px;
```
LeandroTorresSicilia, line 7: we need to avoid having commented code

src/components/Tree/styled/itemContainerLi.js
```diff
@@ -2,19 +2,28 @@ import styled from 'styled-components';
...
+    `};
+    > div {
+        margin-left: ${props =>
+            props['aria-level'] > 1
+                ? `-${(props['aria-level'] - 1) * 20 + props['aria-level'] - 1}px`
+                : '0'};
+        padding-left: ${props =>
+            props['aria-level'] > 1
+                ? `${(props['aria-level'] - 1) * 20 + props['aria-level'] - 1}px`
+                : '0'};
```
LeandroTorresSicilia, line 26: I think this is too complex, we should find another way of doing this

src/components/Tree/styled/itemContainerLi.js
```diff
@@ -2,19 +2,28 @@ import styled from 'styled-components';
...
+    }
+    ${props =>
+        props.hasChildren &&
+        `
+        > div > button {
+            visibility: visible;
+        }
+    `};
+    > div {
+        margin-left: ${props =>
```
LeandroTorresSicilia, line 19: I think it is better to apply styles directly to the div, which is more maintainable and readable

PR #1727, opened 2020-07-16
test: tree component

LeandroTorresSicilia:
fix: #

-

@nexxtway/react-rainbow


LeandroTorresSicilia:
add these test cases:
- [ ] should not set aria-expanded when the node does not have children
- [ ] should set the aria-expanded value of the node when it has children
- [ ] should set the aria-selected value to true when the node is selected
- [ ] should not set the aria-selected value when the node is not selected
- [ ] should set the role="group" in the child ul element

PR #1732, opened 2020-07-19
test: add test when copy and paste on the PhoneInput component to accept only number

devrchancay:
fix: #1679 

- test: add test when copy and paste on the PhoneInput component to accept only number

@nexxtway/react-rainbow

### DIFF
--- a/src/components/PhoneInput/__tests__/phoneinput.spec.js
+++ b/src/components/PhoneInput/__tests__/phoneinput.spec.js
@@ -1,0 +1,27 @@
+import React from 'react';
+import { mount } from 'enzyme';
+import PhoneInput from '../';
+
+const Wrapper = () => {
+    const [value, setValue] = React.useState('');
+    return (
+        <PhoneInput
+            label="Phone Number"
+            value={value}
+            onChange={setValue}
+            placeholder="Enter your phone number"
+        />
+    );
+};
+
+describe('<PhoneInput />', () => {
+    it('Should accept only numbers.', () => {
+        const wrapper = mount(<Wrapper />);
+        wrapper
+            .find('input[type="tel"]')
+            .first()
+            .simulate('change', { target: { value: 'Your phone number is 0987256983' } });
+        wrapper.update();
+        expect(wrapper.find('input[type="tel"]').prop('value')).toBe('0987256983');
+    });
+});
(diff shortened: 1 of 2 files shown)


src/components/PhoneInput/__tests__/phoneinput.spec.js
```diff
@@ -0,0 +1,27 @@
...
+            value={value}
+            onChange={setValue}
+            placeholder="Enter your phone number"
+        />
+    );
+};
+
+describe('<PhoneInput />', () => {
+    it('Should accept only numbers.', () => {
+        const wrapper = mount(<Wrapper />);
```
LeandroTorresSicilia, line 19: I think here you can use the PhoneInput directly and instead check for the onChange event like this:
```
const onChangeMockFn = jest.fn();
const component = mount(< PhoneInput onChange={onChangeMockFn} />);
// simulate
expect(onChangeMockFn).toHaveBeenCalledWith(...);
```

devrchancay: In the way you suggest, how do I verify that the input only contains numbers?

LeandroTorresSicilia: @devrchancay instead of doing:
```expect(wrapper.find('input[type="tel"]').prop('value')).toBe('0987256983');```
you can do:
```
expect(onChangeMockFn).toHaveBeenCalledWith(expect.objectContaining({
    phone: '0987256983',
}));
```

Comment (devrchancay) [reply src/components/PhoneInput/__tests__/phoneinput.spec.js:19]: Oh thanks!

File: src/components/PhoneInput/__tests__/phoneinput.spec.js [hunk 2]
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import { mount } from 'enzyme';
+import PhoneInput from '../';
+
+describe('<PhoneInput />', () => {
+    it('Should accept only numbers.', () => {
```
yvmunayev, line 6: test descriptions are lowercase

LeandroTorresSicilia, line 6:
```suggestion
    it('should accept only numbers.', () => {
```

PR #1745, opened 2020-07-23
refactor: import useOutsideClick to rainbow-modules-hooks

yvmunayev:
fix: #1628 

- refactor: import useOutsideClick to rainbow-modules-hooks
- fix: fix a bug in InternalOverlay

@nexxtway/react-rainbow


LeandroTorresSicilia:
@yvmunayev take a look at CI since all unit tests are failing

src/components/InternalOverlay/index.js
```diff
@@ -95,9 +95,10 @@ const InternalOverlay = props => {
         if (isVisible && contentMeta) {
             onOpened();
             disableBodyScroll(undefined, { reserveScrollBarGap: true });
-        } else {
-            enableBodyScroll();
         }
+        return () => {
+            enableBodyScroll();
+        };
```
LeandroTorresSicilia, line 101: I have doubts, whats happen when set isVisible to false and the component is not unmounted? then enableBodyScroll will not be called when it should do.

yvmunayev: React also sanitizes the effects of previous renderings before running the effects of the current render. So when you change isVisible enableBodyScroll is always called

https://es.reactjs.org/docs/hooks-effect.html#example-using-hooks-1

src/components/MultiSelect/index.js
```diff
@@ -147,20 +141,19 @@ const MultiSelect = React.forwardRef((props, ref) => {
...
-        handleOutsideClick,
+        event => {
+            if (
+                event.target !== triggerRef.current.buttonRef.current &&
+                !triggerRef.current.buttonRef.current.contains(event.target)
+            ) {
+                setIsOpen(false);
+            }
+        },
+        isOpen,
```
LeandroTorresSicilia, line 154: Multiselect outside click is not working, we should add integration test in order to avoid this happening

Issue #1611, opened 2020-06-05
test: the InternalDropdown component


TahimiLeonBravo:
### Unit Tests Specs Proposal
Unit tests

- [x] should render svg and input elements when enableSearch is passed.
- [x] should not render svg and input elements when enableSearch is not passed.
- [x] should render loading icon when isLoading is passed.
- [x] should not render loading icon when isLoading is not passed.
- [x] should fire onChange with the right data when select one option and multiple is passed and there are no options selected.
- [x] should fire onChange with the right data when select one option and multiple is passed and have other options selected.
- [x] should fire onChange with the right data when deselecting one option and multiple is passed and have other options selected.

### Integration Tests Specs Proposal
- [x] should render only filtered options when type in search.
- [x] should render EmptyMessage component when no results found in the search.
- [x] should not render EmptyMessage component when displaying options.
- [x] should render Arrow component at top of the list of options when scrolling down the list.
- [x] should not render Arrow component at top of the list of options when scroll is at top of the list.
- [x] should render Arrow component at bottom of the list of options when scrolling top the list.
- [x] should not render Arrow component at bottom of the list of options when scroll is at bottom of the list.
- [x] should not render Arrow component when the list of options fits the container's height.
- [x] should Arrow component at top and bottom be visible both when first option and last options are not visible
- [x] should set active next option when press arrow down key
- [x] should select the option clicked
- [x] should select the option active when press enter key

LeandroTorresSicilia:
other tests to add:
- [ ] should fire onChange with the right data when select one option and multiple is passed and there are no options selected
- [ ] should fire onChange with the right data when select one option and multiple is passed and have other options selected
- [ ] should fire onChange with the right data when deselecting one option and multiple is passed and have other options selected
- [ ] should render only filtered options when type in search

PR #1751, opened 2020-07-24
test: internal dropdown

hportales2000:
fix: #1611 

-

@nexxtway/react-rainbow


src/components/InternalDropdown/__test__/internalDropdown.spec.js
```diff
@@ -51,4 +148,36 @@ describe('InternalDropdown', () => {
...
+    it('should render svg and input elements when enableSearch is passed', () => {
+        const component = mount(
+            <InternalDropdown label="Picklist" enableSearch>
+                <PicklistOption label="Option 1" name="option1" />
+                <PicklistOption label="Option 2" name="option2" />
+                <PicklistOption label="Option 3" name="option3" />
+            </InternalDropdown>,
+        );
+        expect(component.find('svg').exists()).toBe(true);
+        expect(component.find('input').exists()).toBe(true);
```
LeandroTorresSicilia, line 160: let's be more precise here and use:
```expect(component.find('input[type="search"]').exists()).toBe(true);```

File: src/components/InternalDropdown/__test__/internalDropdown.spec.js [hunk 2]
```diff
@@ -51,4 +148,36 @@ describe('InternalDropdown', () => {
...
+    it('should not render svg and input elements if enableSearch is not passed', () => {
+        const component = mount(
+            <InternalDropdown label="Picklist">
+                <PicklistOption label="Option 1" name="option1" />
+                <PicklistOption label="Option 2" name="option2" />
+                <PicklistOption label="Option 3" name="option3" />
+            </InternalDropdown>,
+        );
+        expect(component.find('svg').exists()).toBe(false);
+        expect(component.find('input').exists()).toBe(false);
```
LeandroTorresSicilia, line 171: same ^^

src/components/InternalDropdown/__test__/internalDropdown.spec.js
```diff
@@ -51,4 +148,36 @@ describe('InternalDropdown', () => {
...
+    });
+    it('should not render svg and input elements if enableSearch is not passed', () => {
+        const component = mount(
+            <InternalDropdown label="Picklist">
+                <PicklistOption label="Option 1" name="option1" />
+                <PicklistOption label="Option 2" name="option2" />
+                <PicklistOption label="Option 3" name="option3" />
+            </InternalDropdown>,
+        );
+        expect(component.find('svg').exists()).toBe(false);
```
LeandroTorresSicilia, line 170: here we should be more precise too, since in the future we add another svg elment it will fail, then we shoul find the element itself ```find(SearchIcon)```

src/components/InternalDropdown/pageObject/index.js
```diff
@@ -36,6 +36,15 @@ class PageInternalDropdown {
...
     }
 
+    /**
+     * It moves the pointer over the menu item.
+     * @method
+     */
+    hover() {
+        const itemElement = $(`${this.rootElement} > div[role="option"]`);
+        itemElement.moveTo();
+    }
```
LeandroTorresSicilia, line 46: this method is part of the Option component, look at other implementations where we use it like:
```
...
const option2 = dropdown.getOption(2);
option2.hover()
...
```

src/components/InternalDropdown/pageObject/index.js
```diff
@@ -75,6 +84,73 @@ class PageInternalDropdown {
...
+            .isExisting();
+    }
+    /**
+     * Returns true when the search no results found, false otherwise.
+     * @method
+     * @returns {bool}
+     */
+    emptyMessageExist() {
+        return $(this.rootElement)
+            .$('p*=Our robots')
```
LeandroTorresSicilia, line 116: What happen if we decide to change this text? We should use a unique selector that will not change.

src/components/InternalDropdown/pageObject/index.js
```diff
@@ -75,6 +84,73 @@ class PageInternalDropdown {
...
+            .isExisting();
+    }
+
+    /**
+     * Clicks the input element.
+     * @method
+     */
+    clickInputSearch() {
```
LeandroTorresSicilia, line 124: clickSearch

src/components/InternalDropdown/pageObject/index.js
```diff
@@ -75,6 +84,73 @@ class PageInternalDropdown {
...
+        return $(this.rootElement)
+            .$('[data-id="internal-dropdown-arrow-up"]')
+            .isExisting();
+    }
+    /**
+     * Returns true when the search no results found, false otherwise.
+     * @method
+     * @returns {bool}
+     */
+    emptyMessageExist() {
```
LeandroTorresSicilia, line 114: isEmpty

src/components/InternalDropdown/pageObject/index.js
```diff
@@ -75,6 +84,73 @@ class PageInternalDropdown {
...
+        $(this.rootElement)
+            .$('input[type="search"]')
+            .click();
+    }
+    /**
+     * Type in the input element.
+     * @method
+     * @param {string} value - The value to type in the input element.
+     */
+    setSearchCriteria(value) {
```
LeandroTorresSicilia, line 134: to be consistent with what we do in lookup, use setQuery

src/components/InternalDropdown/pageObject/index.js
```diff
@@ -75,6 +84,73 @@ class PageInternalDropdown {
...
+    isVisible() {
+        return $(this.rootElement).isDisplayedInViewport();
+    }
+    /**
+     *  Wait until the option is visible.
+     * @method
+     */
+    waitUntilIsVisible() {
+        browser.waitUntil(() => this.isVisible());
+    }
```
LeandroTorresSicilia, line 153: in this case we don't need this both last methods, for example when use it in picklist is where there is a functionality where the menu isVisible and hidden, but here there is not such functionality

integration/specs/InternalDropdown/internalDropdown-1.spec.js
```diff
@@ -0,0 +1,50 @@
...
+        component.waitForExist();
+    });
+    it('should render Arrow component at top of the list of options when scrolling down the list', () => {
+        const internalDropdown = new PageInternalDropdown(INTERNALDROPDOWN);
+        internalDropdown.hoverScrollDownArrow();
+        expect(internalDropdown.arrowUpExists()).toBe(true);
+    });
+    it('should not render Arrow component at top of the list of options when scroll is at top of the list', () => {
+        const internalDropdown = new PageInternalDropdown(INTERNALDROPDOWN);
+        internalDropdown.getOption(0);
```
LeandroTorresSicilia, line 22: here there is no need to get the first option

integration/specs/InternalDropdown/internalDropdown-3.spec.js
```diff
@@ -0,0 +1,39 @@
...
+        const internalDropdown = new PageInternalDropdown(INTERNALDROPDOWN);
+        internalDropdown.clickInputSearch();
+        internalDropdown.setSearchCriteria('c');
+        expect(internalDropdown.getOptionsLength()).toBe(2);
+    });
+    it('should not render Arrow component when the list of options fits the container', () => {
+        const internalDropdown = new PageInternalDropdown(INTERNALDROPDOWN);
+        internalDropdown.clickInputSearch();
+        internalDropdown.setSearchCriteria('c');
+        expect(internalDropdown.getOptionsLength()).toBe(2);
```
LeandroTorresSicilia, line 24: since in the above test this was checked, then remove this

integration/specs/InternalDropdown/internalDropdown-1.spec.js
```diff
@@ -0,0 +1,50 @@
...
+        expect(internalDropdown.arrowDownExists()).toBe(false);
+    });
+    it('should render Arrow component at top and bottom when first option and last options are not visible', () => {
+        const internalDropdown = new PageInternalDropdown(INTERNALDROPDOWN);
+        const option4 = internalDropdown.getOption(5);
+        option4.hover();
+        browser.keys(ARROW_DOWN_KEY);
+        expect(internalDropdown.arrowUpExists()).toBe(true);
+        expect(internalDropdown.arrowDownExists()).toBe(true);
+    });
```
LeandroTorresSicilia, line 49: here we should add tests for keyboard navigation, Option page object have an isActive method, then we should test that when press arrow key the next option is active:
- [ ] should set active next option when press arrow down key
- [ ] should select the option clicked
- [ ] should select the option active when press enter key

integration/specs/InternalDropdown/internalDropdown-3.spec.js
```diff
@@ -13,22 +13,14 @@ describe('InternalDropdown base example', () => {
...
+        internalDropdown.setQuery('c');
         expect(internalDropdown.getOptionsLength()).toBe(2);
     });
-    it('should not render Arrow component when the list of options fits the container', () => {
-        const internalDropdown = new PageInternalDropdown(INTERNALDROPDOWN);
-        internalDropdown.clickInputSearch();
-        internalDropdown.setSearchCriteria('c');
-        expect(internalDropdown.getOptionsLength()).toBe(2);
-        expect(internalDropdown.arrowUpExists()).toBe(false);
-        expect(internalDropdown.arrowDownExists()).toBe(false);
```
LeandroTorresSicilia: here I meant to remove only:
```expect(internalDropdown.getOptionsLength()).toBe(2);```
this assertion no the enterie test

Issue #1742, opened 2020-07-21
tests: unit and integration tests to InternalOverlay


TahimiLeonBravo:
### Description
Add full testing to the InternalOverlay component

#### internalOverlay
- [x] should render if isVisible is true
- [x] should not render if isVisible is false
- [x] should disable scroll when isVisible is true
- [x] should enable scroll when isVisible is false
- [x] should resolve content meta
- [ ] should call positionResolver with the right parameters when it is passed
- [ ] should call onOpened when isVisible is true
- [ ] should not call onOpened when isVisible is false
- [ ] should call triggerElementRef when triggerElementRef is a function

#### internalOverlay-1
- [x] should exist after one click on trigger and destroy on second click
- [x] should not exists after click outside of element
- [ ] should be rendered in the right position respect to the trigger element

PR #1753, opened 2020-07-24
test: unit and integration tests to InternalOverlay component

rgah2107:
fix: #1742

- add unit and integration tests to InternalOverlay component

@nexxtway/react-rainbow


src/components/InternalOverlay/pageObject/button.js
```diff
@@ -19,6 +19,15 @@ class PageInternalOverlayButton {
...
     }
+
+    /**
+     * Returns true when the trigger element has focus.
+     * @method
+     * @returns {bool}
+     */
+    hasFocusTrigger() {
+        return $(this.rootElement).isFocused();
+    }
```
LeandroTorresSicilia, line 30: we were wrong with this page object, I think we should remove it since the trigger element is not part of the InternalOverlay, instead simply get the trigger button and do whatever needed in the test

src/components/InternalOverlay/__tests__/internalOverlay.spec.js
```diff
@@ -85,4 +85,59 @@ describe('<InternalOverlay />', () => {
...
+        const positionResolverMock = jest.fn();
+        mount(
+            <InternalOverlay
+                render={() => <div id="test-id" />}
+                triggerElementRef={{}}
+                positionResolver={positionResolverMock}
+                isVisible
+            />,
+        );
+        expect(positionResolverMock).toHaveBeenCalledTimes(1);
```
LeandroTorresSicilia, line 99: I think we should remove this test since the below one already test that the position resolver was called

integration/specs/InternalOverlay/internalOverlay-1.spec.js
```diff
@@ -14,6 +15,20 @@ describe('InternalOverlay with default position resolver', () => {
         component.waitForExist();
     });
 
+    it('should set focus on trigger when click it', () => {
+        const button = new PageInternalOverlayButton(BUTTON);
+        button.click();
+        expect(button.hasFocusTrigger()).toBe(true);
+    });
```
LeandroTorresSicilia, line 22: we should remove this test since this is not behavior for the InternalOverlay, since we don't set the focus for anything inside the component this is the default browser behavior and then not make sense test it

integration/specs/InternalOverlay/internalOverlay-1.spec.js
```diff
@@ -14,6 +15,20 @@ describe('InternalOverlay with default position resolver', () => {
...
+
+    it('should remove focus from trigger when clicked outside it', () => {
+        const button = new PageInternalOverlayButton(BUTTON);
+        button.click();
+        expect(button.hasFocusTrigger()).toBe(true);
+        $('[data-preview="InternalOverlay"]').click();
+        expect(button.hasFocusTrigger()).toBe(false);
+    });
```
LeandroTorresSicilia, line 30: same ^^

integration/specs/InternalOverlay/internalOverlay-1.spec.js
```diff
@@ -31,4 +46,53 @@ describe('InternalOverlay with default position resolver', () => {
         $('[data-preview="InternalOverlay"]').click();
         expect(internalOverlay.exists()).toBe(false);
     });
+
+    it('should set focus on trigger when press tab key', () => {
+        const button = new PageInternalOverlayButton(BUTTON);
+        $('[data-preview="InternalOverlay"]').click();
+        browser.keys(TAB_KEY);
+        expect(button.hasFocusTrigger()).toBe(true);
```
LeandroTorresSicilia, line 54: same ^^

integration/specs/InternalOverlay/internalOverlay-1.spec.js
```diff
@@ -31,4 +46,53 @@ describe('InternalOverlay with default position resolver', () => {
...
+    });
+
+    it('should exist after set focus on trigger and press enter key', () => {
+        const button = new PageInternalOverlayButton(BUTTON);
+        const internalOverlay = new PageInternalOverlay(OVERLAY);
+        $('[data-preview="InternalOverlay"]').click();
```
LeandroTorresSicilia, line 60: where is this element?

integration/specs/InternalOverlay/internalOverlay-1.spec.js
```diff
@@ -31,4 +46,53 @@ describe('InternalOverlay with default position resolver', () => {
...
+    it('should not exist after set focus on trigger and press space key twice', () => {
+        const button = new PageInternalOverlayButton(BUTTON);
+        const internalOverlay = new PageInternalOverlay(OVERLAY);
+        $('[data-preview="InternalOverlay"]').click();
+        browser.keys(TAB_KEY);
+        expect(button.hasFocusTrigger()).toBe(true);
+        browser.keys(SPACE_KEY);
+        browser.keys(SPACE_KEY);
+        expect(internalOverlay.exists()).toBe(false);
+    });
```
LeandroTorresSicilia, line 97: I think we should keep only the two first test, simply checking once that is open and close, removing the button page object, and using the trigger directly in the test.
The other thing we should test is that the container render in the right position respect to the trigger element, but we need to discuss about it, think the approach to test this and any method to add to the page object and let's discuss before doing any test

rgah2107: I get it

integration/specs/InternalOverlay/internalOverlay-1.spec.js
yvmunayev: @rgah2107 I think the best solution is to add an id to the <Container> of the example, this way you can use it as a trigger for the OutsideClick.

integration/specs/InternalOverlay/internalOverlay-1.spec.js
```diff
@@ -15,18 +14,18 @@ describe('InternalOverlay with default position resolver', () => {
...
     });
 
     it('should not exists after click outside of element', () => {
-        const button = new PageInternalOverlayButton(BUTTON);
         const internalOverlay = new PageInternalOverlay(OVERLAY);
-        button.click();
+        const triggerButton = $(BUTTON);
+        triggerButton.click();
         expect(internalOverlay.exists()).toBe(true);
         $('[data-preview="InternalOverlay"]').click();
```
yvmunayev, line 30: It is not okay to use an element outside the example for integration testing because it makes them vulnerable.

I think the best solution is to add an id to the Container of the example, this way you can use it as a trigger for the OutsideClick.

integration/specs/InternalOverlay/internalOverlay-1.spec.js
LeandroTorresSicilia: > I think we should keep only the two first test, simply checking once that is open and close, removing the button page object, and using the trigger directly in the test.
> The other thing we should test is that the container render in the right position respect to the trigger element, but we need to discuss about it, think the approach to test this and any method to add to the page object and let's discuss before doing any test

@rgah2107 We need to add test to positioning, but first you need to make a page object methods proposal

rgah2107: received

rgah2107:
```
getPositionRespectToTheTriggerElement(triggerElement) {
        const triggerElementBoundsRect = {
            ...triggerElement.getLocation(),
            ...triggerElement.getSize(),
        };
        const internalOverlaytBoundsRect = {
            ...$(this.rootElement).getLocation(),
            ...$(this.rootElement).getSize(),
        };
        console.log('triggerElementBoundsRect');
        console.log(triggerElementBoundsRect);
        console.log('internalOverlaytBoundsRect');
        console.log(internalOverlaytBoundsRect);
        if (
            internalOverlaytBoundsRect.x === triggerElementBoundsRect.x &&
            internalOverlaytBoundsRect.y > triggerElementBoundsRect.y
        ) {
            return 'BottomLeft';
        }
        if (
            internalOverlaytBoundsRect.x + internalOverlaytBoundsRect.width ===
                triggerElementBoundsRect.x + triggerElementBoundsRect.width &&
            internalOverlaytBoundsRect.y > triggerElementBoundsRect.y
        ) {
            return 'BottomRight';
        }
        if (
            internalOverlaytBoundsRect.x < triggerElementBoundsRect.x &&
            internalOverlaytBoundsRect.x + internalOverlaytBoundsRect.width >
                triggerElementBoundsRect.x + triggerElementBoundsRect.width &&
            internalOverlaytBoundsRect.y > triggerElementBoundsRect.y
        ) {
            return 'BottomCenter';
        }
        if (
            internalOverlaytBoundsRect.x === triggerElementBoundsRect.x &&
            internalOverlaytBoundsRect.y < triggerElementBoundsRect.y
        ) {
            return 'UpLeft';
        }
        if (
            internalOverlaytBoundsRect.x + internalOverlaytBoundsRect.width ===
                triggerElementBoundsRect.x + triggerElementBoundsRect.width &&
            internalOverlaytBoundsRect.y < triggerElementBoundsRect.y
        ) {
            return 'UpRight';
        }
        if (
            internalOverlaytBoundsRect.x < triggerElementBoundsRect.x &&
            internalOverlaytBoundsRect.x + internalOverlaytBoundsRect.width >
                triggerElementBoundsRect.x + triggerElementBoundsRect.width &&
            internalOverlaytBoundsRect.y < triggerElementBoundsRect.y
        ) {
            return 'UpCenter';
        }
        if (
            internalOverlaytBoundsRect.x > triggerElementBoundsRect.x &&
            internalOverlaytBoundsRect.y < triggerElementBoundsRect.y
        ) {
            return 'CenterLeft';
        }
        if (
            internalOverlaytBoundsRect.x + internalOverlaytBoundsRect.width <
                triggerElementBoundsRect.x &&
            internalOverlaytBoundsRect.y < triggerElementBoundsRect.y
        ) {
            return 'CenterRight';
        }
        return null;
    }
```

rgah2107: this could be a spec if we use the above method:
```
it('should be rendered in the right position respect to the trigger element', () => {
        const internalOverlay = new PageInternalOverlay(OVERLAY);
        const triggerButton = $(BUTTON);
        triggerButton.click();
        expect(internalOverlay.exists()).toBe(true);
        expect(internalOverlay.getPositionRespectToTheTriggerElement(triggerButton)).toBe('UpLeft');
    });
```

rgah2107: in the method I assume that the dimensions of the container are always greater than the dimensions of the trigger element, as is the case in the examples

rgah2107: I also believe that this method could be partitioned into one for each possible position and would return a boolean instead of a string

example:
```
isPositionedToBottomLeft(triggerElement){...}
```

PR #1755, opened 2020-07-25
feat: add support multi-language to MarkdownOutput component

yvmunayev:
#1743

- feat: add support multi-language to MarkdownOutput component

@nexxtway/react-rainbow


TahimiLeonBravo:
@yvmunayev we should represent the styles of the code depending on the language defined by the user.
- when the user doesn't define any language all the code should look like with the same color
- when the user defines a language we should detect the language and represent the code with the styles of the language added

if we make that, it should look like this:
Screen Shot 2020-07-26 at 1 39 23 PM

Issue #1760, opened 2020-07-27
docs: add the "data table with custom actions and wrap description" Interactive Example to the Table component


TahimiLeonBravo:
### Requirements
- [ ] It should have a new example with the name "data table with custom actions and wrap description"
- [ ] It should look like the design
https://zpl.io/blPO9Kz

PR #1763, opened 2020-07-28
docs: add the data table with custom actions and wrap description

lyon93:
fix: #1760

-

@nexxtway/react-rainbow

### DIFF
--- a/src/components/Table/readme.md
+++ b/src/components/Table/readme.md
@@ -984,68 +984,71 @@ function TableExample(props) {
 ##### Data Table with custom actions and wrap description
 
 ```js
-import {
-    Table,
-    Column,
-    ButtonGroup,
-    ButtonIcon,
-} from 'react-rainbow-components';
-import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
-import styled from 'styled-components';
 import React, { useState } from 'react';
-import {faEllipsisV,
-        faPlus,
-        faTrashAlt } from '@fortawesome/free-solid-svg-icons';
+import styled from 'styled-components';
+import { Table, Column, ButtonGroup, ButtonIcon } from 'react-rainbow-components';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import { faEllipsisV, faPlus, faTrashAlt } from '@fortawesome/free-solid-svg-icons';
 
 const initialData = [
     {
         name: 'Leandro Torres',
-        description:'Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et. Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et.',
+        description:
+            'Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et. Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et.',
         id: '1234qwerty',
     },
     {
         name: 'Jose Torres',
-        description:'Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et. Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et.',
+        description:
+            'Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et. Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et.',
         id: '1234asdfgh',
     },
     {
         name: 'Reinier',
-        description:'Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et. Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et.',
+        description:
+            'Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et. Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et.',
         id: '1234zxcvbn',
     },
     {
         name: 'Sara',
-        description:'Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et. Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et.',
+        description:
+            'Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et. Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et.',
         id: '5678qwerty',
     },
     {
         name: 'Tahimi',
-        description:'Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et. Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et.',
+        description:
+            'Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et. Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et.',
         id: '5678asdfgh',
     },
     {
         name: 'Leo',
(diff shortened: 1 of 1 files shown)


src/components/Table/readme.md
```diff
@@ -981,6 +981,145 @@ function TableExample(props) {
...
+    const StyledActionCell = styled.div`
+        display: flex;
+        justify-content: center;
+        align-content: center;
+        align-items: center;
+        margin: auto;
+        box-sizing: border-box;
+    `;
+    const CustomAction = ({row})=>(
+            <StyledActionCell>
```
TahimiLeonBravo, line 1084: I think that if we use a `width = {60}` in the column, we don't have to add a wrapper or styles to the button. Also in this way, we better take advantage of the Table space because we keep always small the column.

src/components/Table/readme.md
```diff
@@ -981,6 +981,145 @@ function TableExample(props) {
...
+            <div className="rainbow-flex rainbow-justify_spread rainbow-align_center rainbow-p-around_large">
+                    <div>
+                        <Title>Members</Title>
+                        <Subtitle>Total • {data.length}</Subtitle>
+                    </div>
+                        
+            </div>
+            <Table keyField="id" data={data}>
+                <Column header="Name" field="name" />
+                <Column header="Description" field="description" />
```
TahimiLeonBravo, line 1114: I think it would be much more useful if we make the text appear complete in the description column, I mean to wrap it. It would be much more evident if we put the text of the description a little longer.
In this way, we would have an example that covers this use case that we use many times.
You can follow the design example

src/components/Table/readme.md
```diff
@@ -981,6 +981,145 @@ function TableExample(props) {
...
+                    <ButtonIcon variant="border-filled" disabled icon={<FontAwesomeIcon icon={faPlus} />} />
+                    <ButtonIcon
+                        variant="border-filled"
+                        disabled
+                        icon={<FontAwesomeIcon icon={faEllipsisV} />}
+                    />
+                </ButtonGroup>
+            </GlobalHeader>
+            <div className="rainbow-flex rainbow-justify_spread rainbow-align_center rainbow-p-around_large">
+                    <div>
```
TahimiLeonBravo, line 1106: I think you don't need this `div` element here
Remember, we should make the examples as simple and clean as possible so we avoid adding noise to the other developers who come to learn how to use the component.

src/components/Table/readme.md
```diff
@@ -981,6 +981,145 @@ function TableExample(props) {
...
+    const customActionStyles ={marginLeft: "auto"};
+    const StyledActionCell = styled.div`
+        display: flex;
+        justify-content: center;
+        align-content: center;
+        align-items: center;
+        margin: auto;
+        box-sizing: border-box;
+    `;
+    const CustomAction = ({row})=>(
```
yvmunayev, line 1083: We do not declare of component within another component.

The CustomAction component should be declared independent and receive as props {row, onDeleteElement}.
To use them you can make these

```js
<Column component={row => <CustomAction row={row} onDeleteElement={handleOnDeleteElement} />}/>
```

src/components/Table/readme.md
```diff
@@ -981,6 +981,145 @@ function TableExample(props) {
...
+    `;  
+
+const Subtitle = styled.h3.attrs(props => props.theme.rainbow)`
+    font-family: Lato;
+    font-size: 14px;
+    color: ${props => props.palette.text.title};
+    `;
+
+function TableCustomAction(props) {
+    const [data, setData] = useState(props.data);
```
yvmunayev, line 1067: Copying props to a state is bad practice. I think a cleaner solution would be this:

```js
 const [data, setData] = useState(initialData);
```

and remove the prop in the component call.

src/components/Table/readme.md
```diff
@@ -981,6 +981,145 @@ function TableExample(props) {
...
+    ButtonIcon,
+} from 'react-rainbow-components';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import styled from 'styled-components';
+import React, { useState } from 'react';
+import {faEllipsisV,
+        faPlus,
+        faTrashAlt } from '@fortawesome/free-solid-svg-icons';
+
+const data = [
```
yvmunayev, line 1000: rename initialData

src/components/Table/readme.md
```diff
@@ -981,6 +981,134 @@ function TableExample(props) {
...
+            />
+    );
+}
+
+function TableCustomAction() {
+    const [data, setData] = useState(initialData);
+    function handleOnDeleteElement(id) {
+            const newData = data.filter(e => e.id!==id);
+            setData(newData)
+        }
```
yvmunayev, line 1082: indentation

```js
const newData = data.filter(e => e.id !== id);
```

src/components/Table/readme.md
```diff
@@ -981,6 +981,134 @@ function TableExample(props) {
...
+                onClick={()=>onDeleteElement(row.id)}
+                buttonSize="small"
+                icon={<FontAwesomeIcon icon={faTrashAlt} />}
+            />
+    );
+}
+
+function TableCustomAction() {
+    const [data, setData] = useState(initialData);
+    function handleOnDeleteElement(id) {
```
yvmunayev, line 1079: change to handleDeleteElement

src/components/Table/readme.md
```diff
@@ -981,6 +981,134 @@ function TableExample(props) {
...
+            </GlobalHeader>
+            <div className="rainbow-p-around_large">
+                <Title>Members</Title>
+                <Subtitle>Total • {data.length}</Subtitle>
+            </div>
+            <Table keyField="id" data={data}>
+                <Column  defaultWidth={200} header="Name" field="name" />
+                <Column  header="Description" field="description" />
+                <Column  width={60} component={
+                    ({row}) =><CustomAction row={row} onDeleteElement={handleOnDeleteElement} />}/>
```
yvmunayev, line 1104: indentation

```js
<Column
    width={60}
    component={({row}) =><CustomAction row={row} onDeleteElement={handleDeleteElement} />}
/> 
```

src/components/Table/readme.md
TahimiLeonBravo: @lyon93 With this, I mean that we make the "Description" column have more than one line of text in the example (that is, do not put the ellipsis "...") just like the design
Screen Shot 2020-07-29 at 4 25 48 PM

src/components/Table/readme.md
```diff
@@ -981,6 +981,151 @@ function TableExample(props) {
...
+    `;
+
+function CustomAction(props){
+    const {row, onDeleteElement} = props;
+    return (
+            <ButtonIcon
+                onClick={()=>onDeleteElement(row.id)}
+                buttonSize="small"
+                icon={<FontAwesomeIcon icon={faTrashAlt} />}
+            />
```
yvmunayev, line 1073: indentation

src/components/Table/readme.md
```diff
@@ -981,6 +981,151 @@ function TableExample(props) {
...
+    white-space:pre-line;
+    line-height: 30px;
+`;
+
+function TableCustomAction() {
+    const [data, setData] = useState(initialData);
+    function handleDeleteElement(id) {
+        const newData = data.filter(e => e.id !== id);
+        setData(newData);
+        }
```
yvmunayev, line 1089: indentation

src/components/Table/readme.md
```diff
@@ -981,6 +981,151 @@ function TableExample(props) {
...
+            <div className="rainbow-p-around_large">
+                <Title>Members</Title>
+                <Subtitle>Total • {data.length}</Subtitle>
+            </div>
+            <Table keyField="id" data={data}>
+               <Column  defaultWidth={250} header="Name" field="name" />
+               <Column
+                header="Description"
+                field="description"
+                component={({value})=><WrapDescription>{value}</WrapDescription>}
```
yvmunayev, line 1116: indentation

src/components/Table/readme.md
```diff
@@ -981,6 +981,151 @@ function TableExample(props) {
...
+               />
+               <Column
+                width={60}
+                component={({row})=><CustomAction row={row} onDeleteElement={handleDeleteElement}/>}
```
yvmunayev, line 1120: indentation

src/components/Table/readme.md
```diff
@@ -981,6 +981,151 @@ function TableExample(props) {
...
+                        icon={<FontAwesomeIcon icon={faEllipsisV} />}
+                    />
+                </ButtonGroup>
+            </GlobalHeader>
+            <div className="rainbow-p-around_large">
+                <Title>Members</Title>
+                <Subtitle>Total • {data.length}</Subtitle>
+            </div>
+            <Table keyField="id" data={data}>
+               <Column  defaultWidth={250} header="Name" field="name" />
```
yvmunayev, line 1112: indentation

src/components/Table/readme.md
```diff
@@ -981,6 +981,151 @@ function TableExample(props) {
...
+    Column,
+    ButtonGroup,
+    ButtonIcon,
+} from 'react-rainbow-components';
+import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
+import styled from 'styled-components';
+import React, { useState } from 'react';
+import {faEllipsisV,
+        faPlus,
+        faTrashAlt } from '@fortawesome/free-solid-svg-icons';
```
yvmunayev, line 998: indentation

Issue #1738, opened 2020-07-20
feat: implement keyboard navigation on the Tree component


TahimiLeonBravo:
We should follow the keyboard navigation specs defined on w3c

#### Sources
- https://www.w3.org/TR/wai-aria-practices/examples/treeview/treeview-2/treeview-2b.html
- https://www.lightningdesignsystem.com/components/trees/#site-main-content

PR #1781, opened 2020-08-08
feat: add keyboard navigation to Tree component 

wildergd:
fix: #1738 

- add keyboard navigation to Tree component

@nexxtway/react-rainbow


src/components/Tree/child.js
```diff
@@ -30,14 +32,20 @@ export default function Child(props) {
         name,
         selectedNode,
         isFirstNode,
+        autoFocus,
+        focusedNode,
+        setFocusedNode,
+        clearFocusedNode,
+        privateKeyDown,
```
LeandroTorresSicilia, line 39: we should follow the convention to name the callbacks starting with the on prefix like:
onPrivateKeyDown, onFocusNode, ...

src/components/Tree/child.js
```diff
@@ -107,6 +123,10 @@ export default function Child(props) {
     );
 }
 
+export default function Child(props) {
+    return <Consumer>{values => <ChildComponent {...values} {...props} />}</Consumer>;
+}
```
LeandroTorresSicilia, line 128: it is more clean if we instead use the useContext hook, also since this component has only tree levels and children are no external component, we should simply pass props through and then avoid using context
https://reactjs.org/docs/context.html#before-you-use-context

src/components/Tree/helpers/buildPlainListFromTree.js
```diff
@@ -0,0 +1,34 @@
+import getNodeName from './getNodeName';
+import getNodeLevel from './getNodeLevel';
+
+export default function buildPlainListFromTree(tree, parentName, parentPath) {
+    return tree.reduce((result, node, index) => {
+        const nodeName = getNodeName({ parentName, index });
+        const level = getNodeLevel({ name: nodeName });
+        const nodePath = parentPath ? [...parentPath, index] : [index];
+        if (node.isExpanded)
```
LeandroTorresSicilia, line 9: we should use brackets to indicate blocks since it is more readable:
```
if (node.isExpanded) {
...
}
```

LeandroTorresSicilia:
When the tree gets the focus, if we press tab key then the focus should move to the next page element, now it gets stuck in the Tree, we should add test to this behavior

Issue #1777, opened 2020-08-07
feat: implement ColorPicker component


TahimiLeonBravo:
### Requirements
- [ ] Define API
- [ ] Implement component
- [ ] Add Tests

#### Design
- https://zpl.io/bLMWd1G

#### API
- https://docs.google.com/document/d/1DHR6KCk5MQSpEZ8ZlJNmoo0w62CrnHPiqm-FcWQ34yw/edit

PR #1795, opened 2020-08-25
feat: implement ColorPicker component

yvmunayev:
fix: #1777

- feat: implement ColorPicker component

@nexxtway/react-rainbow


LeandroTorresSicilia:
@TahimiLeonBravo when using the color picker you need to pass an height to its container in order to see it like
Screen Shot 2020-09-01 at 11 50 59 AM

Instead if you don't pass an height then it looks like:
Screen Shot 2020-09-01 at 11 51 59 AM
where the gradient is hidden
this is the expected behavior?

src/components/ColorPicker/commons/saturation/index.js
```diff
@@ -0,0 +1,108 @@
...
+
+    const unbindEventListeners = () => {
+        window.removeEventListener('mousemove', handleChange);
+        window.removeEventListener('mouseup', unbindEventListeners);
+    };
+
+    const handleMouseDown = event => {
+        handleChange(event);
+        window.addEventListener('mousemove', handleChange);
+        window.addEventListener('mouseup', unbindEventListeners);
```
LeandroTorresSicilia, line 52: can we avoid setting listener in window and instead attach it to specific element to reduce its scope?

TahimiLeonBravo:
> @TahimiLeonBravo when using the color picker you need to pass a height to its container in order to see it like
> Screen Shot 2020-09-01 at 11 50 59 AM
> 
> Instead if you don't pass a height then it looks like:
> Screen Shot 2020-09-01 at 11 51 59 AM
> where the gradient is hidden
> this is the expected behavior?

@LeandroTorresSicilia 
this is not expected, I think it should have a min-height

src/components/ColorPicker/commons/saturation/index.js
yvmunayev: I need to keep listening for the move event when it is outside the component so that the behavior is like the native component

TahimiLeonBravo (approved):

Issue #1789, opened 2020-08-17
fix: keep scrolling disabled while there is a visible InternalOverlay.


yvmunayev:
-   [x] I have searched the [issues](https://github.com/nexxtway/react-rainbow/issues) of this repository and believe that this is not a duplicate.

## Describe the bug

When there are several visible InternalOverlay and one is hidden, the scroll is enabled.

## To Reproduce 🕹

1. Go to HelpText
2. Click on HelpText info button
3. Hover HelpText error button
4. MouseLeave HelpText error button
5. See error

## Expected Behavior 🤔

As long as there is a visible Internal Overlay, the scroll must be disabled.

Issue #1031, opened 2019-09-27
fix: Modal behavior on iOS


Johnnycus:
-   [x] I have searched the [issues](https://github.com/nexxtway/react-rainbow/issues) of this repository and believe that this is not a duplicate.

## Describe the bug

Modal component doesn't have a scroll bar on iOS (both Safari and Chrome). Scrollbar is visible and is working in Google Dev Tools but not on the actual iOS device.

## To Reproduce 🕹

1. Create a Modal component
2. Add a lot of text or elements inside to make it scrollable
3. Check the Modal on iOS device
4. See that it's not scrollable

## Expected Behavior 🤔

It should allow scrolling when inside text or elements are taking a lot of space

## Current Behavior 😯

No scrolling on iOS devices which makes it impossible to read the text

## Your Environment 🌎

**Smartphone (please complete the following information):**

-   Device: iPhone X
-   OS: iOS 12
-   Browser Chrome, Safari

reiniergs:
probably adding `-webkit-overflow-scrolling: touch;` to the Modal content container It will solve that issue

Johnnycus:
> probably adding `-webkit-overflow-scrolling: touch;` to the Modal content container It will solve that issue

That doesn't help, unfortunately. I have spent some time today trying to fix this issue but wasn't able to.

reiniergs:
it's not going to work because this component is preventing scroll in the body `body lock approach`, check this https://github.com/nexxtway/react-rainbow/blob/master/src/components/Modal/index.js#L38 

@LeandroTorresSicilia this needs to be a more elegant solution. check this out https://github.com/willmcpo/body-scroll-lock 

/cc @TahimiLeonBravo

TahimiLeonBravo:
@Johnnycus We have already fixed the modal issue in IOS. This is now available in the new release v1.6.0

Johnnycus:
> @Johnnycus We have already fixed the modal issue in IOS. This is now available in the new release v1.6.0

Thanks for the Modal fix but after updating to a new version my app doesn't work because there is no "styled" folder in node_modules/react-rainbow-components. See the [issue](https://github.com/nexxtway/react-rainbow/issues/1050)

PR #1817, opened 2020-09-11
feat: implement useScrollLock hook

yvmunayev:
fix: #1789 #1826 

- feat: implement useScrollLock hook
- test: add unit test to useScrollLock hook
- refactor: refactor InternalOverlay to use useScrollLock
- refactor: refactor Drawer to use useScrollLock

@nexxtway/react-rainbow


TahimiLeonBravo:
@yvmunayev I don't remember we discuss these hook, please you can explain the uses cases for them

src/components/Drawer/index.js
```diff
@@ -57,22 +51,14 @@ export default function Drawer(props) {
     );
 
     useEffect(() => {
-        const contentElement = contentRef.current;
```
LeandroTorresSicilia, line 60: seems that contentRef is not being used anymore then let's clean all code related to it

src/components/Drawer/index.js
```diff
@@ -85,6 +71,8 @@ export default function Drawer(props) {
         }
     }, [drawerState, isOpen, onOpened]);
 
+    useScrollLock(isOpen);
```
LeandroTorresSicilia, line 74: last example in drawer (https://react-rainbow.io/#!/Drawer/7) is not working properly, when open drawer then select birthday input which open a modal, when click outside it closes both the modal and the drawer. It must only close the modal. (Before fix this issue let’s add an integration test to this case)
Maybe it need to be used in Modal in order to count the locks, but Modal is class then I think we should leave Drawer as before for now and use the hook only in the internal dropdown for now

src/libs/hooks/useScrollLock/helpers/handleOverflow.js
```diff
@@ -0,0 +1,43 @@
...
+    });
+}
+
+export default function handleOverflow(locks) {
+    if (locks.length === 1) {
+        hiddenOverflow();
+    }
+    if (locks.length === 0) {
+        restoreOverflow();
+    }
```
LeandroTorresSicilia, line 42: what happens when locks have more than 1 element?

src/libs/hooks/useScrollLock/helpers/handleOverflow.js
```diff
@@ -0,0 +1,43 @@
...
+
+            if (scrollBarGap > 0) {
+                previousPaddingRight = document.body.style.paddingRight;
+                document.body.style.paddingRight = `${scrollBarGap}px`;
+            }
+        }
+
+        if (previousOverflowSetting === undefined) {
+            previousOverflowSetting = document.body.style.overflow;
+            document.body.style.overflow = 'hidden';
```
LeandroTorresSicilia, line 17: I have doubts about this implementation since we have implemented a more complex here src/libs/scrollController/index.js in order to support all browser and mobile take a look at this old issue:
https://github.com/nexxtway/react-rainbow/issues/1031

src/libs/hooks/useScrollLock/helpers/handleOverflow.js
yvmunayev: When the length of lock is 1 the scroll will become lock, when it is greater than 1 it is not necessary to do anything because it is already lock.

src/components/Drawer/index.js
```diff
@@ -85,6 +73,8 @@ export default function Drawer(props) {
...
```
HellWolf93, line 76: Last example in drawer ([https://react-rainbow.io/#!/Drawer/7](https://react-rainbow.io/#!/Drawer/7)) still does not work. Now when open drawer, then birthday modal, close modal and close drawer, scroll remains locked.

HellWolf93 (changes requested):

Issue #1814, opened 2020-09-10
feat: hide the selected info on the FileSelector component


TahimiLeonBravo:
### Requirements
- [ ] do not render selected file(s) by passing value={null}, but onChange callback will be invoke every time file(s) were selected
- [ ] add tests

PR #1820, opened 2020-09-12
feat: hide in FileSelector selected info

amontalvof:
fix: #1814 

-

@nexxtway/react-rainbow


src/components/FileSelector/__test__/icon.spec.js
```diff
@@ -0,0 +1,74 @@
...
+    it('should render a FileIcon', () => {
+        const files = [
+            {
+                name: 'filename.jpg',
+            },
+        ];
+        const isDragOver = false;
+        const component = mount(<Icon files={files} isDragOver={isDragOver} />);
+        const IconUpload = component.find(FileIcon);
+        expect(IconUpload.exists()).toBe(true);
```
LeandroTorresSicilia, line 57: let's use a more generic name to the icon since here there is not IconUpload at all, you can use:
```const icon = component.find(FileIcon);```

File: src/components/FileSelector/icon.js [hunk 2]
```diff
@@ -0,0 +1,35 @@
...
+            if (files && files.length > 1) {
+                return <FilesIcon />;
+            }
+        }
+    }
+    return uploadIcon || <UploadIcon />;
+}
+Icon.propTypes = {
+    files: PropTypes.array,
+    error: PropTypes.string,
```
LeandroTorresSicilia, line 23: check at index file the error can be a node too then you need:
```error: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),```

File: src/components/FileSelector/index.js [hunk 3]
```diff
@@ -234,6 +240,8 @@ FileSelector.propTypes = {
     onFocus: PropTypes.func,
     /** The action triggered when the element releases focus. */
     onBlur: PropTypes.func,
+    /** An object to do not let change the icon and text information  */
```
LeandroTorresSicilia, line 243: check we @TahimiLeonBravo we need more explicit description

LeandroTorresSicilia (approved):

PR #1840, opened 2020-09-21
docs: add example with an initial value set in the input component

amontalvof:
fix: #1831 

-

@nexxtway/react-rainbow

### DIFF
--- a/src/components/Input/readme.md
+++ b/src/components/Input/readme.md
@@ -17,7 +17,7 @@ const containerStyles = {
 />;
 ```
 
-##### input base with an initial value set
+##### input base with a value controlled through a state and an initial value set by default
 
 ```js
 import React, { useState } from 'react';
--- a/src/components/Input/readme.md
+++ b/src/components/Input/readme.md
@@ -17,33 +17,6 @@ const containerStyles = {
 />;
 ```
 
-##### input base with a value controlled through a state and an initial value set by default
-
-```js
-import React, { useState } from 'react';
-import { Input } from 'react-rainbow-components';
-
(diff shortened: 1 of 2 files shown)


src/components/Input/readme.md
```diff
@@ -17,6 +17,33 @@ const containerStyles = {
 />;
 ```
 
+##### input base with an initial value set
```
TahimiLeonBravo, line 20: I think we should start adding a useful and short little description for the new interactive examples

TahimiLeonBravo (changes requested):

src/components/Input/readme.md
```diff
@@ -17,6 +17,33 @@ const containerStyles = {
...
+const handleChange = (e) => {
+    setValue(e.target.value)
+}
+
+const containerStyles = {
+    maxWidth: 700,
+};
+
+<Input
+    id="input-component-1"
```
LeandroTorresSicilia, line 37: The id must be unique on a web page, the first example already use `id="input-component-1"`
Also keep in mind when adding new examples and use the ids that the integration tests depend on this ids and on the example order too

TahimiLeonBravo (approved):

TahimiLeonBravo:
@amontalvof let's do the new examples in this way:
- Each example we should have a title and a description
- It should look like the design https://zpl.io/29LEldY

TahimiLeonBravo:
@amontalvof we should use a smaller font-size for this example
show as it looks like now
Screen Shot 2020-09-28 at 12 25 54 AM

PR #1852, opened 2020-09-23
docs: example of dynamic row action in table component

amontalvof:
fix: #1815 

-

@nexxtway/react-rainbow

### DIFF
--- a/src/components/Table/exampleData.js
+++ b/src/components/Table/exampleData.js
@@ -1,0 +1,294 @@
+export const data = [
+    {
+        name: 'Leandro Torres',
+        company: 'Nexxtway',
+        email: 'leandro@gmail.com',
+        status: 'verified',
+        id: '1234qwerty',
+    },
+    {
+        name: 'Reinier',
+        company: 'Nexxtway',
+        email: 'reinier@gmail.com',
+        status: 'verified',
+        id: '1234asdfgh',
+    },
+    {
+        name: 'Jose Torres',
+        company: 'Google',
+        email: 'jose@gmail.com',
+        status: 'verified',
+        id: '1234zxcvbn',
+    },
+    {
+        name: 'Sara',
+        company: 'Nexxtway',
+        email: 'sara@gmail.com',
+        status: 'verified',
+        id: '5678qwerty',
+    },
+    {
+        name: 'Tahimi',
+        company: 'Nexxtway',
+        email: 'tahimi@gmail.com',
+        status: 'verified',
+        id: '5678asdfgh',
+    },
+    {
+        name: 'Alejandro',
+        company: 'Google',
+        email: 'alejandro@gmail.com',
+        status: 'verified',
+        id: '5678zxcvbn',
+    },
+    {
+        name: 'Carlos',
+        company: 'Oracle',
+        email: 'carlos@gmail.com',
+        status: 'verified',
+        id: '9012qwerty',
+    },
+    {
+        name: 'Luis',
+        company: 'Google',
+        email: 'luis@gmail.com',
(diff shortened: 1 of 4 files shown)


src/components/Table/readme.md
```diff
@@ -577,6 +577,114 @@ const StatusBadge = ({ value }) => <Badge label={value} variant="lightest" style
...
+        id: '5678asdfgh',
+    },
+    {
+        name: 'Carlos',
+        company: 'Oracle',
+        email: 'carlos@gmail.com',
+        status: 'pending',
+        id: '5678zxcvbn',
+    },
+];
```
reiniergs, line 631: @TahimiLeonBravo I think we are repeating this list on multiple examples, we should have an internal files that export a list if persons that we can use in the examples like:
```
import persons from '/path/to/files';
```

TahimiLeonBravo: I agree, let's do this @amontalvof

src/components/Table/readme.md
```diff
@@ -577,6 +577,114 @@ const StatusBadge = ({ value }) => <Badge label={value} variant="lightest" style
...
+                icon={<FontAwesomeIcon icon={faEllipsisV} />}
+            />
+        </ButtonGroup>
+    </GlobalHeader>
+    <Table keyField="id" data={data}>
+        <Column header="Name" field="name" />
+        <Column header="Status" field="status" component={StatusBadge} />
+        <Column header="Company" field="company" />
+        <Column header="Email" field="email" />
+        <Column field="status" component={ButtonAction}/>
```
TahimiLeonBravo, line 683: in order to have a more beautiful example we should add a `width={60}` to this column

src/components/Table/readme.md
```diff
@@ -577,6 +577,114 @@ const StatusBadge = ({ value }) => <Badge label={value} variant="lightest" style
...
+    );
+};
+
+const ButtonAction = (props) => {
+    const {value, row:{name}}=props;
+    return (
+        <ButtonMenu
+            id="button-menu-2"
+            menuAlignment="right"
+            menuSize="x-small"
```
TahimiLeonBravo, line 658: we should use the `buttonVariant="base"` on the ButtonMenu component

TahimiLeonBravo (changes requested):

src/components/Table/readme.md
```diff
@@ -577,51 +304,81 @@ const StatusBadge = ({ value }) => <Badge label={value} variant="lightest" style
...
+            <MenuItem label="Edit" onClick={() => console.log(`Edit ${name}`)}/>
+        </>
+    );
+};
+
+const ButtonAction = (props) => {
+    const {value, row:{name}}=props;
+    return (
+        <ButtonMenu
+            id="button-menu-2"
```
LeandroTorresSicilia, line 340: take care this id is not used already in other Table example

src/components/Table/exampleData.js
```diff
@@ -0,0 +1,294 @@
+export const data = [
```
LeandroTorresSicilia, line 1: we should move this file to this folder `library/data` where we already have a `user.js` file and it is set globally in the `library/setup.js` file

TahimiLeonBravo:
@amontalvof let's do the new examples in this way:
- Each example we should have a title and a description
- It should look like the design https://zpl.io/29LEldY

src/components/Table/readme.md
amontalvof: I checked and the id is not used in any other example

LeandroTorresSicilia (approved):

PR #1834, opened 2020-09-19
feat: add `orientation` prop on the `RadioGroup` and `CheckboxGroup` components

luisFilipePT:
fix: #1824 

## Screenshots

image

image

image

image

## Changes proposed in this PR:
- Add optional orientation prop to radio group and checkbox group (default vertical)
- Possible values ['horizontal', 'vertical]

- [x] I have followed (at least) the [PR section of the contributing guide](https://github.com/nexxtway/react-rainbow/blob/master/CONTRIBUTING.md#submitting-a-pull-request).

@nexxtway/react-rainbow

PR #1855, opened 2020-09-25
feat: add prop labelAlignment to select component

### LINKED ISSUE #1844: feat: add `labelAlignment` prop to the `Select` component
### Requirements
- [ ] Implement feature
- [ ] Add Tests
- [ ] add InteractiveExample

### Specs
- [ ] It should have a `labelAlignment` prop to control the position of the label
- [ ] It should receive the `center | right | left ` values
- [ ] The default value should be `center`

luisFilipePT:
feat(select): add prop labelAlignment to select component

closes #1844

fix: #1844 

-
image

image

image

image

@nexxtway/react-rainbow

### DIFF
--- a/src/components/Select/__test__/select.spec.js
+++ b/src/components/Select/__test__/select.spec.js
@@ -41,30 +41,4 @@ describe('Select component', () => {
         const component = mount(<Select label="Select Label" required />);
         expect(component.find('SelectStyledLabel').prop('labelAlignment')).toBe('center');
     });
-    it('renders correctly with label aligned left', () => {
-        const component = mount(<Select label="Select Label" required labelAlignment="left" />);
-        const elem = component.find('SelectStyledLabel');
-
-        expect(getComputedStyle(elem.getDOMNode()).getPropertyValue('align-self')).toBe(
-            'flex-start',
-        );
-    });
-    it('renders correctly with label aligned right', () => {
-        const component = mount(<Select label="Select Label" required labelAlignment="right" />);
-        const elem = component.find('SelectStyledLabel');
-
-        expect(getComputedStyle(elem.getDOMNode()).getPropertyValue('align-self')).toBe('flex-end');
-    });
-    it('renders correctly with label centered (default)', () => {
-        const component = mount(<Select label="Select Label" required />);
-        const elem = component.find('SelectStyledLabel');
-
-        expect(getComputedStyle(elem.getDOMNode()).getPropertyValue('align-self')).toBe('center');
-    });
-    it('renders correctly with label centered by passing prop (explicit)', () => {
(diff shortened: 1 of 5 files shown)


luisFilipePT:
Some tests are failing, but they run fine locally (see printscreen in the description) and they follow the same logic as ones ina previous pull request that passed :/ (https://github.com/nexxtway/react-rainbow/pull/1834)

@LeandroTorresSicilia can you check?

TahimiLeonBravo (approved):

src/components/Select/__test__/select.spec.js
```diff
@@ -29,4 +29,42 @@ describe('Select component', () => {
...
+        const component = mount(<Select label="Select Label" required />);
+        const elem = component.find('SelectStyledLabel');
+
+        expect(getComputedStyle(elem.getDOMNode()).getPropertyValue('align-self')).toBe('center');
+    });
+    it('renders correctly with label centered by passing prop (explicit)', () => {
+        const component = mount(<Select label="Select Label" required labelAlignment="center" />);
+        const elem = component.find('SelectStyledLabel');
+
+        expect(getComputedStyle(elem.getDOMNode()).getPropertyValue('align-self')).toBe('center');
```
LeandroTorresSicilia, line 68: for me also fails locally, but since these are simply styles and we usually don't test them then we can remove these test and just leave the other where you check that SelectStyledLabel component receives the right labelAlignment prop

src/components/Select/styled/label.js
```diff
@@ -3,13 +3,19 @@ import attachThemeAttrs from '../../../styles/helpers/attachThemeAttrs';
...
+};
+
 const StyledLabel = attachThemeAttrs(styled.label)`
     color: ${props => props.palette.text.label};
     font-size: ${FONT_SIZE_TEXT_MEDIUM};
     line-height: 1.5;
     margin-right: ${MARGIN_SMALL};
     margin-bottom: ${MARGIN_XX_SMALL};
-    align-self: center;
+    align-self: ${props => LABEL_ALIGNMENT_MAPPING[props.labelAlignment]};
```
LeandroTorresSicilia, line 18: when a user passes a wrong labelAlignment value that does not exist in the map we should fallback to 'center' like:
``` align-self: ${props => LABEL_ALIGNMENT_MAPPING[props.labelAlignment] || 'center'};```

src/components/Select/__test__/select.spec.js
luisFilipePT: done

src/components/Select/styled/label.js
luisFilipePT: done

LeandroTorresSicilia (approved):

TahimiLeonBravo (approved):

PR #1874, opened 2020-09-30
refactor:  TimePicker component

gtallesb:
fix: #

-update TimePicker component to be functional
-use hooks

@nexxtway/react-rainbow


src/components/TimePicker/index.js
```diff
@@ -14,170 +14,148 @@ import StyledModal from './styled/modal';
...
-        const { value } = this.props;
-        if (prevValue !== value) {
-            this.updateValue();
-        }
-    }
-
-    getTriggerInputValue() {
-        const { value } = this.state;
-        const { placeholder, hour24 } = this.props;
+const TimePicker = props => {
```
yvmunayev, line 17: You must use React.forwardRef () to keep the ref

src/components/TimePicker/index.js
```diff
@@ -14,170 +14,148 @@ import StyledModal from './styled/modal';
...
+        error,
+        readOnly,
+        disabled,
+        tabIndex,
+        id,
+        cancelLabel,
+        okLabel,
+        onChange,
+        hour24,
+    } = props;
```
yvmunayev, line 37: you must use useReduxForm (props) to maintain integration with ReduxForm

src/components/TimePicker/index.js
```diff
@@ -14,170 +14,148 @@ import StyledModal from './styled/modal';
...
-            this.setState({ isOpen: true });
+            setIsOpen(true);
         }
-    }
+    };
 
-    handleClick(event) {
-        const { onClick, readOnly } = this.props;
+    const handleClick = event => {
+        const { onClick } = props;
```
yvmunayev, line 65: delete this onClick you can define it at the beginning

src/components/TimePicker/index.js
```diff
@@ -14,170 +14,148 @@ import StyledModal from './styled/modal';
...
+            setIsOpen(true);
             onClick(event);
         }
-    }
+    };
 
-    handleBlur() {
-        const { onBlur, value } = this.props;
+    const handleBlur = () => {
+        const { onBlur } = props;
```
yvmunayev, line 73: same before

src/components/TimePicker/index.js
```diff
@@ -14,170 +14,148 @@ import StyledModal from './styled/modal';
...
         onBlur(value);
-    }
+    };
 
-    handleFocus() {
-        const { onFocus, value } = this.props;
+    const handleFocus = () => {
+        const { onFocus } = props;
```
yvmunayev, line 78: same before

src/components/TimePicker/index.js
```diff
@@ -14,170 +14,148 @@ import StyledModal from './styled/modal';
...
     /**
      * Sets focus on the element.
      * @public
      */
-    focus() {
-        this.inputRef.current.focus();
-    }
+    const focus = () => {
+        inputRef.current.focus();
+    };
```
yvmunayev, line 97: you must use useImperativeHandle () to export these functions.

You can be guided by PhoneInput

TahimiLeonBravo:
@gtallesb please check the tests that are failing

src/components/TimePicker/__test__/timePicker.spec.js
```diff
@@ -82,14 +83,14 @@ describe('<TimePicker/>', () => {
         });
     });
     it('should call to focusHourInput when the modal is open', () => {
```
HellWolf93, line 85: As this test no longer checks this, we should change this description to something like "should focus the hour input when modal is open"

src/components/TimePicker/index.js
```diff
@@ -9,182 +10,147 @@ import withReduxForm from '../../libs/hocs/withReduxForm';
...
+            </StyledModal>
+        </StyledContainer>
+    );
+});
 
 TimePicker.propTypes = {
     /** Sets the date for the TimePicker programmatically. */
     value: PropTypes.string,
     /** Override the label of the 'Cancel' button. */
+    // eslint-disable-next-line react/no-unused-prop-types
```
HellWolf93, line 153: This rule is already disabled at file level

HellWolf93 (changes requested):
If you are using `useReduxForm` hook there is no need to use the `withReduxForm` hoc on the export. (Github won't let me comment the specific line, is the last line in the file)

HellWolf93:
@gtallesb integration tests are failing, please check this

src/components/TimePicker/index.js
```diff
@@ -1,185 +1,149 @@
...
+        cancelLabel,
+        okLabel,
+        onChange,
+        hour24,
+        onClick,
+        onBlur,
+        onFocus,
+    } = useReduxForm(props);
+    const [isOpen, setIsOpen] = useState(false);
+    const [value, setValue] = useState(hour24 ? props.value : get12HourTime(props.value));
```
LeandroTorresSicilia, line 43: in case we use redux form we need to take the value prop from useReduxForm like all other props to get the right value

LeandroTorresSicilia, line 43: to distinguish form state `value` you can:
```value: valueProp```
in props destructuring

File: src/components/TimePicker/index.js [hunk 10]
```diff
@@ -1,185 +1,149 @@
...
+    };
+
+    const closeModal = () => {
+        setIsOpen(false);
+    };
+    useEffect(() => {
+        if (prevValue !== props.value) {
+            updateValue();
+        }
+    });
```
LeandroTorresSicilia, line 101: this should be used to substitute the  `componentDidUpdate` functionality, then you don't need to maintain the prevValue by yourself, instead you can use:
```
useEffect(() => {
    updateValue();
}, [valueProp]);
```
this do the same since it will enter in the useEffect every time the value prop change

src/components/TimePicker/__test__/timePicker.spec.js
```diff
@@ -81,15 +82,15 @@ describe('<TimePicker/>', () => {
...
-            ).toHaveBeenCalledTimes(1);
-        }, 0);
+    it('should focus hour input when modal is open', () => {
+        const wrapper = mount(<TimePicker label="unit-testing-timePicker" />);
+        wrapper.find('input').simulate('click');
+        expect(
+            wrapper
+                .find('input')
+                .at(1)
+                .is(':focus'),
```
LeandroTorresSicilia, line 92: this test is very well, I think we can use a more specific selector to the input like:
```
.find('input[aria-label="hour"]')
```
to be more confident we are selecting the right input since there are four of them

LeandroTorresSicilia (approved):

PR #1897, opened 2020-10-09
docs: improve doc in badge component

YulyGP:
Badges border and Badges variant were united into a single category. The category name is Badges variant.

fix: #

-

@nexxtway/react-rainbow


TahimiLeonBravo (approved):

src/components/Badge/readme.md
```diff
@@ -1,4 +1,5 @@
-##### badges border
+# Badges variant
+##### Badges change color according to the variant applied to them. You can use a lightest or outline brand for borders, or even change it for a default, inverse, brand, warning, success, or error variant that we provide you.
```
TahimiLeonBravo, line 3: I believe that the names of the values of the component variant should not be in the description of the example, this is more difficult to maintain because if we add a new variant we would have to add it to the description and these values are already well described in the table from "props and methods"

TahimiLeonBravo (changes requested):

TahimiLeonBravo:
@YulyGP please review the "Contributor License Agreement". Needs to be signed

TahimiLeonBravo (approved):

Issue #1880, opened 2020-10-03
feat: add `variant` prop on the Tabset component


TahimiLeonBravo:
### Requirements
- [ ] Implement new prop `variant` on the Tabset component
- [ ] Add test
- [ ] Add a new interactive example on the Tabset

#### Design
https://zpl.io/VODL45r

#### Usage
```
<Tabset
    id="tabset-7"
    onSelect={}
    activeTabName={}
    variant="line"
  >
       <Tab
             label=""
             name=""
             id=""
        />
</Tabset>
```

### Specs
- [ ] It should have a new `variant` prop
- [ ] The variant prop should have this possibles values `"card" | "line"` and by default the value should be "card"

PR #1899, opened 2020-10-09
feat: add variant prop on the Tabset component

rgah2107:
fix: #1880

- add variant prop
- add tests
- add a new interactive example

@nexxtway/react-rainbow


src/components/Tabset/index.d.ts
```diff
@@ -5,6 +5,7 @@ export interface TabsetProps extends BaseProps {
     activeTabName?: string;
     onSelect?: (event: MouseEvent<HTMLElement>, name: string) => void;
     fullWidth?: boolean;
+    variant?: 'card' | 'linear';
```
LeandroTorresSicilia, line 8: the new variant is `line`

src/components/Tabset/index.js
```diff
@@ -280,6 +326,8 @@ Tabset.propTypes = {
     /** If true, the tabs will grow to use all the available space.
      * This value defaults to false. */
     fullWidth: PropTypes.bool,
+    /** The variant changes the appearance of the Tabset. Accepted variants include card and linear. The default value is card. */
+    variant: PropTypes.oneOf(['card', 'linear']),
```
LeandroTorresSicilia, line 330: same ^^

src/components/Tabset/index.js
```diff
@@ -76,6 +83,22 @@ export default class Tabset extends Component {
...
+                const currentTabMeta = getTabMeta(activeTabName, tabsetChildren);
+                if (
+                    activeTabName !== prevProp.activeTabName ||
+                    tabMeta.left !== currentTabMeta.left ||
+                    tabsetMeta.width !== currentTabsetMeta.width
+                ) {
+                    this.updateIndicator();
+                }
+            }
+        }
```
LeandroTorresSicilia, line 101: I think it is more simply pass the logic of painting the new variant to the Tab component, this avoiding doing all this computation.
If it is complex to mix the two variants then you can do in Tab/index.js:
```
if (variant === 'line') {
    return ...
}
return ...
```
then you can render two different components based on the variant

src/components/Tabset/readme.md
```diff
@@ -791,3 +791,155 @@ class TabsetExample extends React.Component {
...
+const facebookIconStyles = { color: '#3c5997', height: 30 };
+const twitterIconStyles = { color: '#00b0f3', width: 30 };
+const googleIconStyles = { height: 30 };
+
+class TabsetExample extends React.Component {
+    constructor(props) {
+        super(props);
+        this.state = { selected: 'companies' };
+        this.handleOnSelect = this.handleOnSelect.bind(this);
+    }
```
LeandroTorresSicilia, line 856: we should add new examples using functional components with hooks instead of class components

src/components/Tab/styled/button.js
```diff
@@ -43,30 +44,42 @@ const StyledAnchor = attachThemeAttrs(styled.button)`
...
-        content: '';
-        position: absolute;
-        left: -2px;
-        height: 20px;
-        width: 1px;
-        background-color: ${props => props.palette.border.divider};
-        box-sizing: border-box;
-    }
+    ${props =>
+        (props.variant === 'card' || !variants.includes(props.variant)) &&
```
LeandroTorresSicilia, line 57: there is not much implication since there are only two elements in the array but here you can use a map instead, thus avoiding the iteration

LeandroTorresSicilia (approved):

PR #1930, opened 2020-10-19
fix: hover styles on variant=inverse or border-inverse on ButtonIcon

rgah2107:
fix: #1885

- add background color when variant is inverse or border-inverse in the ButtonIcon component

@nexxtway/react-rainbow


src/components/ButtonIcon/styled/button.js
```diff
@@ -280,6 +280,7 @@ const StyledButton = attachThemeAttrs(styled.button).attrs(props => {
             &:focus,
             &:active {
                 color: ${props.inverse.active};
+                background-color: ${props.palette.background.inverse};
```
LeandroTorresSicilia, line 283: we can compute here the needed value instead of defining a new color

LeandroTorresSicilia (approved):

Issue #346, opened 2018-10-26
feat: implement the Tooltip component


TahimiLeonBravo:
### API
https://docs.google.com/document/d/10kuIRE07OE7_dQ8BZKJyN4o4yfby0KyaNXMwnrLMoCE/edit?ts=5f8f6381

### Examples
https://zpl.io/bPZPBGy
https://zpl.io/b6q1Arj

PR #1947, opened 2020-10-27
feat: implement the Tooltip component

HellWolf93:
fix: #346

- implement Tooltip component


TahimiLeonBravo:
@HellWolf93 @LeandroTorresSicilia @maxxgreene I think we should find a way to document this

HellWolf93:
> @HellWolf93 @LeandroTorresSicilia @maxxgreene I think we should find a way to document this

In my opinion, we should document all the functions defined in libs, it is getting difficult to track all the functionality in there.

src/libs/showTooltip/tooltip/styled/index.js
```diff
@@ -0,0 +1,89 @@
+import styled from 'styled-components';
+import attachThemeAttrs from '../../../../styles/helpers/attachThemeAttrs';
+
+const StyledTooltip = attachThemeAttrs(styled.div)`
+    background-color: ${props => props.palette.text.main};
```
LeandroTorresSicilia, line 5: this is not working with theme

Issue #1958, opened 2020-10-28
fix: `Picklist` when loading and after render options.


TahimiLeonBravo:
### Show bug video
https://drive.google.com/file/d/1LXQ1i0EEaQc4iahBzmR5bczWe-ZzDBaA/view?usp=sharing

PR #1961, opened 2020-10-30
fix: `Picklist` when loading and after render options

wildergd:
fix: #1958 

- fix bug in`Picklist` when loading and after render options.

@nexxtway/react-rainbow


src/components/Picklist/index.js
```diff
@@ -68,12 +54,16 @@ class Picklist extends Component {
...
+        const { isOpen: wasOpen, dropdownSize } = prevState;
         const { isOpen } = this.state;
         if (!wasOpen && isOpen) {
             this.outsideClick.startListening(this.containerRef.current, () => this.closeMenu());
             this.windowScrolling.startListening(this.handleWindowScroll);
         }
+        const currentSize = this.dropdownRef.current && this.dropdownRef.current.getSize();
+        if (isOpen && JSON.stringify(dropdownSize) !== JSON.stringify(currentSize)) {
+            this.updateDropdownSize(this.dropdownRef.current.getSize());
+        }
```
LeandroTorresSicilia, line 66: here instead of doing this, we should simply change the bottom default resolvers in InternalOverlay, thus this will work fine always

LeandroTorresSicilia (approved):

PR #1976, opened 2020-11-04
fix: buttonmenu drawer bug

amontalvof:
fix: #

-

@nexxtway/react-rainbow


src/components/Drawer/__test__/drawer.spec.js
```diff
@@ -122,16 +122,6 @@ describe('<Drawer />', () => {
...
-    it('should fire an event when the drawer backdrop is clicked and the drawer is open', () => {
-        const closeMockFn = jest.fn();
-        const component = mount(
-            <Drawer isOpen onRequestClose={closeMockFn}>
-                <p />
-            </Drawer>,
-        );
-        component.find('div[role="presentation"]').simulate('click');
-        expect(closeMockFn).toHaveBeenCalledTimes(1);
-    });
```
HellWolf93, line 134: I don't see why we should remove this test case, this still is the expected behavior, we only need to make it work.

LeandroTorresSicilia: the problem is that now we use ref.current.addEventListener(), then enzyme does not work when attaching a listener in this way, what we should do at least is add an integration test

LeandroTorresSicilia (approved):

Issue #2004, opened 2020-11-18
feat: implement `Path` and `PathStep` component


TahimiLeonBravo:
### Requirements
- implement component
- add tests
- document example

### Design
https://zpl.io/2jzGNR4
https://zpl.io/V0K3vjl

### Usage
```
 <Path 
          currentStepName (string) 
          onClick (func)
>
           
           <PathStep
                   hasError (bool)
                   label (string/node)
                   name (string)
           />
 </Path>
```

TahimiLeonBravo:
@wildergd we should make some fixes to the component
[ ] Selected PathStep should have an icon
[ ] Steps should be buttons
[ ] Animation for visual feedback when clicking on the step
[ ] Animation when hover step is a "nice to have"
[ ] Step with error should not disable previous steps
[ ] Error should be independent of the step selected state

wildergd:
> @wildergd we should make some fixes to the component
> [ ] Selected PathStep should have an icon
> [ ] Steps should be buttons
> [ ] Animation for visual feedback when clicking on the step
> [ ] Animation when hover step is a "nice to have"
> [ ] Step with error should not disable previous steps
> [ ] Error should be independent of the step selected state

ok... i will do changes ASAP

PR #2028, opened 2020-11-26
feat: implement `Path` and `PathStep` components

wildergd:
fix: #2004 

- add `Path` and `PathStep` components

@nexxtway/react-rainbow


src/components/Path/index.js
```diff
@@ -0,0 +1,85 @@
...
+export default function Path(props) {
+    const { currentStepName, onClick, children, id, className, style } = props;
+    const [hoveredStepName, setHoveredStepName] = useState(null);
+
+    const steps = useMemo(() => {
+        const selectedIndex = children.findIndex(child => child.props.name === currentStepName);
+        const hoveredIndex = children.findIndex(child => child.props.name === hoveredStepName);
+        const someStepHasError = children.some(child => child.props.hasError);
+
+        return React.Children.map(children, (child, index) => {
```
LeandroTorresSicilia, line 18: we should use context for this since with this approach we can do some things, for example, if we want to use a RenderIf to show some PathStep dynamically, this can't be done because RenderIf will be the child, this is a drawback in our table component right now

src/components/Path/index.d.ts
```diff
@@ -0,0 +1,11 @@
+import { ReactNode, MouseEvent } from 'react';
+import { BaseProps } from '../types';
+
+export interface PathProps extends BaseProps {
+    currentStepName?: string;
+    onClick?: (event: MouseEvent<HTMLElement>) => void;
```
LeandroTorresSicilia, line 6: onClick send the name then it is a string instead of MouseEvent<HTMLElement>

src/components/Path/index.js
```diff
@@ -0,0 +1,85 @@
...
+                    isSelected={isSelected}
+                    hasError={hasError}
+                    isChecked={isChecked}
+                    key={name}
+                    onMouseEnter={() => setHoveredStepName(name)}
+                    onMouseLeave={() => setHoveredStepName(null)}
+                    onClick={() => onClick(name)}
+                >
+                    {label}
+                    <RenderIf isTrue={isChecked && index !== hoveredIndex}>
```
LeandroTorresSicilia, line 45: let's use these conditions in const with meaningful names

src/components/Path/index.js
```diff
@@ -0,0 +1,85 @@
...
+                        <CheckMark />
+                    </RenderIf>
+                    <RenderIf isTrue={hasError && index !== hoveredIndex}>
```
LeandroTorresSicilia, line 48: same ^^

src/components/Path/readme.md
```diff
@@ -0,0 +1,35 @@
...
+const initialState = { currentStep: 'arrived' };
+
+    <div className="rainbow-p-around_x-large rainbow-align-content_center">
+        <Path currentStepName={state.currentStep} onClick={stepName => setState({ currentStep: stepName })}>
+            <PathStep name="scheduled" label="Scheduled" />
+            <PathStep name="in-progress" label="InProgress" />
+            <PathStep name="arrived" label="Arrived" />
+            <PathStep name="delivered" label="Delivered" />
+        </Path>
+    </div>
```
LeandroTorresSicilia, line 16: for new examples we will not use this anymore, instead make it a functional component with hooks, note that this form the user can't copy and paste

src/components/PathStep/index.js
```diff
@@ -0,0 +1,28 @@
...
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export default function PathStep(props) {
+    return <></>;
+}
+
+PathStep.propTypes = {
+    /** The name of the PathStep. */
+    name: PropTypes.string,
```
LeandroTorresSicilia, line 11: here name does not match with typescript where it also accepts a number, I think we should accept numbers too

src/components/PathStep/index.d.ts
```diff
@@ -0,0 +1,9 @@
+import { BaseProps } from '../types';
+
+export interface PathStepProps extends BaseProps {
+    label?: string;
```
LeandroTorresSicilia, line 4: label is a node

src/components/PathStep/readme.md
```diff
@@ -0,0 +1,35 @@
...
+const initialState = { currentStep: 'delivered', hasError: true };
+
+    <div className="rainbow-p-around_x-large rainbow-align-content_center">
+        <Path currentStepName={state.currentStep}>
+            <PathStep name="scheduled" label="Scheduled" />
+            <PathStep name="in-progress" label="InProgress" />
+            <PathStep name="arrived" label="Arrived" />
+            <PathStep name="delivered" label="Delivered" hasError={state.hasError} />
+        </Path>
+    </div>
```
LeandroTorresSicilia, line 34: same ^^ for these examples

src/components/PathStep/index.js
LeandroTorresSicilia, line 11: let's only use string for now

src/components/Path/index.js
```diff
@@ -0,0 +1,111 @@
...
+                        hoveredIndex,
+                        someStepHasError: hasErrors,
+                        privateGetStepIndex: getStepIndex,
+                        privateGetStepZIndex,
+                        privateRegisterStep,
+                        privateUnregisterStep,
+                        privateUpdateStepProps,
+                        privateOnClick: onClick,
+                        privateUpdateHoveredStep: setHoveredStepName,
+                    }}
```
yvmunayev, line 80: declare a variable for this context

yvmunayev (commented):
@TahimiLeonBravo I think that when hovering in the steps the global size of the component should not change.

wildergd:
> @TahimiLeonBravo I think that when hovering in the steps the global size of the component should not change.

ok fixed

src/components/Path/readme.md
```diff
@@ -0,0 +1,60 @@
...
+import React, { useState, useCallback } from 'react';
+import { Path, PathStep } from 'react-rainbow-components';
+
+
+const BasicPath = () => {
+    const [currentStep, setCurrentStep] = useState('arrived');
+
+    const handleClick = useCallback(stepName => {
+        setCurrentStep(stepName);
+    }, []);
```
LeandroTorresSicilia, line 14: here you can simply use ```onClick={setCurrentStep}```

src/components/PathStep/readme.md
```diff
@@ -0,0 +1,60 @@
...
+
+import React, { useState, useCallback } from 'react';
+import { Path, PathStep } from 'react-rainbow-components';
+
+
+const BasicPath = () => {
+    const [currentStep, setCurrentStep] = useState('arrived');
+
+    const handleClick = useCallback(stepName => {
+        setCurrentStep(stepName);
```
LeandroTorresSicilia, line 13: same ^^

src/components/index.js
```diff
@@ -51,6 +51,8 @@ export { default as MultiSelect } from './MultiSelect';
 export { default as Notification } from './Notification';
 export { default as Option } from './Option';
 export { default as Pagination } from './Pagination';
+export { default as Path } from './Path';
+export { default as PathStep } from './PathStep';
```
LeandroTorresSicilia, line 55: add to index.d.ts too

yvmunayev (approved):

LeandroTorresSicilia (approved):

PR #2072, opened 2020-12-29
fix: the Modal component close unexpected when selecting an option on…

HellWolf93:
… the `MultiSelect` component

fix: #2067

@nexxtway/react-rainbow

### DIFF
--- a/src/components/MultiSelect/index.js
+++ b/src/components/MultiSelect/index.js
@@ -156,6 +156,8 @@ const MultiSelect = React.forwardRef((props, ref) => {
         }
     };
 
+    const handleContainerClick = event => event.stopPropagation();
+
     useOutsideClick(
         dropdownContainerRef,
         event => {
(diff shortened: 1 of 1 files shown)


src/components/MultiSelect/index.js
```diff
@@ -156,6 +156,8 @@ const MultiSelect = React.forwardRef((props, ref) => {
         }
     };
 
+    const handleContainerClick = event => event.stopPropagation();
```
LeandroTorresSicilia, line 159: we need to avoid stopPropagation since can be a user that listens for clicks in containers up in the tree, then these clicks will not get in its listener, we should follow the same approach of the drawer

HellWolf93: Fixed

LeandroTorresSicilia (approved):

LeandroTorresSicilia (approved):

Issue #1909, opened 2020-10-12
fix Codesandbox call stack size and Lato-Black.svg


ghost:
react-rainbow-component v1.18.1
react v16.8.6

Codesandbox console: RangeError: Maximum call stack size exceeded
![Screenshot 2020-10-12 at 8 06 10 AM](https://user-images.githubusercontent.com/67747427/95707382-abdde200-0c59-11eb-906f-70b92b3f81bc.png)

LeandroTorresSicilia:
@organization983 can you provide more context? Like the code, you used in the codexandbox

wildergd:
problem is related to embed assets for legacy styles... (same problem happened when the library within a Next.js project).. is not exactly a bug... however, one solution is to move legacy styles to a cdn and add the corresponding <link ....> to head when Application is mount

LeandroTorresSicilia:
@ghost this has been addressed in version 1.24.1, when using Codesandbox I recommend you to import the components direct like:
```
import Application from 'react-rainbow-components/components/Application';
```
since the other way:
```
import { Application } from 'react-rainbow-components';
```
require all components and some of them still have problems with SSR

PR #2091, opened 2021-01-11
fix: font imports  in `Application` component

wildergd:
fix: #1909 

- remove font assets from lib
- add font css from google-fonts 

@nexxtway/react-rainbow


src/components/Application/rainbowLegacyStyles.js
```diff
@@ -1,133 +1,5 @@
...
-    font-weight: normal;
-    font-style: normal; }
-  
-  @font-face {
-    font-family: 'Lato';
-    src: url(${LatoRegularEot});
-    src: url(${LatoRegularWoff2}) format("woff2"), url(${LatoRegularWoff}) format("woff"), url(${LatoRegularTtf}) format("truetype"), url("${LatoRegularSvg}#Lato-Regular") format("svg"), url("${LatoRegularEot}?#iefix") format("embedded-opentype");
-    font-weight: normal;
-    font-style: normal; }
+  @import url('https://fonts.googleapis.com/css2?family=Lato:ital,wght@0,100;0,300;0,400;0,700;0,900;1,100;1,300;1,400;1,700;1,900&display=swap');
```
LeandroTorresSicilia, line 2: Here we need to be careful since we are using these font-faces defined and it can stop working since they can have other names.
For example, users can be used in their code: ```font-family: 'Lato Bold Italic'```

wildergd: i just included all Lato available fonts in google-fonts... using google-font source url for fonts i don't think is a good idea as it can change in time... lets see if there is a way to add an alias

wildergd: there is no way to add an alias to web fonts, only to local
sol 1: private font hosting
sol 2:  what if instead of using a direct url to the font  asset.. we set the url in base64.. like in this page https://www.filamentgroup.com/lab/font-loading.html... it should increase the size of the lib if we include all font-assets but we can include only the most supported like ttf, woff, woff2...

LeandroTorresSicilia (approved):

Issue #2099, opened 2021-01-13
feat: improve `ActivityTimeline` and `TimelineMarker` components


TahimiLeonBravo:
### ActivityTimeline
- Implement `variant` prop (`acordion` | `default`) default to `default`
- Implement `activeSectionNames` prop
- Implement `multiple` prop
- Implement `onToggleSection` prop

### TimelineMarker
- Implement `isLoading` prop

### Design
https://zpl.io/2Zld4ql

PR #2100, opened 2021-01-15
feat: improve `ActivityTimeline` and `TimelineMarker` components

wildergd:
fix: #2099 

- add `variant` prop to `ActivityTimeline` component
- add `isLoading` prop to TimelineMarker component 

@nexxtway/react-rainbow

### DIFF
--- a/src/components/ActivityTimeline/index.js
+++ b/src/components/ActivityTimeline/index.js
@@ -7,7 +7,7 @@ import { Provider } from './context';
 import StyledUl from './styled/ul';
 
 /**
- * The ActivityTimeline displays each of any item upcoming, current, and past activities.
+ * The ActivityTimeline displays each of any item upcoming, current, and past activities in cronological order (ascending or descending).
  * @category Layout
  */
 export default function ActivityTimeline(props) {
--- a/src/components/ActivityTimeline/index.js
+++ b/src/components/ActivityTimeline/index.js
@@ -1,5 +1,9 @@
-import React from 'react';
+import React, { useRef, useState, useCallback, useMemo, useEffect } from 'react';
 import PropTypes from 'prop-types';
+import isChildRegistered from '../InternalDropdown/helpers/isChildRegistered';
+import insertChildOrderly from '../InternalDropdown/helpers/insertChildOrderly';
+import { getChildTimelineMarkersNodes } from './helpers';
+import { Provider } from './context';
 import StyledUl from './styled/ul';
 
 /**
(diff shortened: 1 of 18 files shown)


TahimiLeonBravo:
@wildergd let's use a `gray color` for the arrows

TahimiLeonBravo:
@wildergd we need to fix the line color, in some cases it is not visible, I think we should use here an alpha color that guarantees that it is darker than the background
Screen Shot 2021-01-15 at 10 06 18 AM

src/components/ActivityTimeline/index.js
```diff
@@ -1,22 +1,101 @@
...
+        privateUnregisterMarker,
+        privateOnToggleMarker,
+    ]);
 
     return (
-        <StyledUl className={className} style={style}>
-            {children}
+        <StyledUl id={id} className={className} style={style} ref={containerRef} variant={variant}>
+            <Provider value={context}>{children}</Provider>
         </StyledUl>
```
LeandroTorresSicilia, line 92: I think here, we should execute all this new code only if the variant is accordion:
```
if (variant === 'accordion') {
   return <NewAccordionComponent />;
}
return <OriginalComponent />;
```

src/components/ActivityTimeline/index.js
```diff
@@ -26,10 +105,32 @@ ActivityTimeline.propTypes = {
...
+    /** It contain the name of the TimelineMarker that is expanded.
+     * It is an array of string when multiple is true,
+     * or a string when when multiple is false.
+     * It must match the name of the TimelineMarker. */
+    activeSectionNames: PropTypes.oneOfType([
+        PropTypes.arrayOf(PropTypes.string),
+        PropTypes.string,
+    ]),
+    /** Action fired when a TimelineMarker is selected.
+     * The event params include the `name` of the selected TimelineMarker. */
```
LeandroTorresSicilia, line 123: we need to explain that this props (multiple, activeSectionNames and onToggleSection) are only used for accordion variant

src/components/TimelineMarker/index.js
```diff
@@ -10,18 +13,97 @@ import StyledLabel from './styled/label';
...
+        isLoading,
+        label,
+        description,
+        datetime,
+        children,
+        className,
+        style,
+    } = props;
+    const {
+        isVariantAccordion,
```
LeandroTorresSicilia, line 37: I think we should pass variant, since if we add new ones in the future then is more easy to make things here based on the variant

src/components/TimelineMarker/index.js
```diff
@@ -10,18 +13,97 @@ import StyledLabel from './styled/label';
...
+            return activeNames === currentName;
+        }
+        return true;
+    }, [isVariantAccordion, currentName, activeNames, multiple]);
+
+    const shouldRenderChildren = useMemo(() => isExpanded && !isLoading, [isLoading, isExpanded]);
 
     return (
-        <StyledLi className={className} style={style}>
+        <StyledLi data-id="timeline-marker-li" className={className} style={style}>
```
LeandroTorresSicilia, line 95: same ^^ here we should only execute all this new code if it is accordion variant

src/components/TimelineMarker/styled/columnLeft.js
```diff
@@ -9,7 +12,7 @@ const StyledColumnLeft = attachThemeAttrs(styled.div)`
 
     ::before {
         content: '';
-        background-color: ${props => props.palette.background.highlight};
+        background-color: ${props => getColor(props)};
```
LeandroTorresSicilia, line 15: can pass `getColor` directly

LeandroTorresSicilia (approved):

PR #2103, opened 2021-01-17
fix: enhance code

wildergd:
fix: #2099 

- some code enhancements

@nexxtway/react-rainbow

### DIFF
--- a/src/components/ActivityTimeline/index.d.ts
+++ b/src/components/ActivityTimeline/index.d.ts
@@ -3,10 +3,14 @@ import { BaseProps } from '../types';
 
 type Names = string[] | string;
 
+interface onToggleSectionEventShape {
+    name: Names;
+}
+
 export interface ActivityTimelineProps extends BaseProps {
     children?: ReactNode;
     multiple?: boolean;
-    onToggleSection?: (event: MouseEvent<HTMLElement>, name: Names) => void;
+    onToggleSection?: (value: onToggleSectionEventShape) => void;
     activeSectionNames?: Names;
 }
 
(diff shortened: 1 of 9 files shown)


src/components/ActivityTimeline/index.d.ts
```diff
@@ -3,10 +3,14 @@ import { BaseProps } from '../types';
...
 
+interface onToggleSectionEventShape {
+    name: Names;
+}
+
 export interface ActivityTimelineProps extends BaseProps {
     children?: ReactNode;
     multiple?: boolean;
-    onToggleSection?: (event: MouseEvent<HTMLElement>, name: Names) => void;
+    onToggleSection?: (value: onToggleSectionEventShape) => void;
```
LeandroTorresSicilia, line 13: we rethink this, and the use case we have is to make a fetch when opening a section, then if here we get an array with all section opened we don't know which was toggled, then we will change this API to:
```({ activeSectionNames: Names, toggledSection: string }) => void;```

File: src/components/ActivityTimeline/readme.md [hunk 2]
```diff
@@ -127,7 +127,8 @@ const AccordionActivityTimeline =  ({ eventsList }) => {
         setTimeout(() => setLoading(false), 3000);
     }, []);
 
-    const handleToggleEvent = useCallback((event, name) => {
+    const handleToggleEvent = useCallback(({ name }) => {
+        console.log(name);
```
LeandroTorresSicilia, line 131: here you forgot to remove the console

LeandroTorresSicilia (approved):

Issue #2106, opened 2021-01-18
fix: iconPosition does not exist on ButtonMenu


maximegheraille:
-   [x] I have searched the [issues](https://github.com/nexxtway/react-rainbow/issues) of this repository and believe that this is not a duplicate.

## Describe the bug

I cannot use the iconPosition with ButtonMenu

I'm getting this error

![image](https://user-images.githubusercontent.com/32335849/104961584-e7b1e180-59d6-11eb-8e50-1dc21f46d344.png)

The interface does not seem to have to property

![image](https://user-images.githubusercontent.com/32335849/104961609-f7c9c100-59d6-11eb-81e2-04764b8a5baf.png)

PR #2131, opened 2021-01-25
fix: iconPosition does not exist on ButtonMenu

yvmunayev:
fix: #2106 

- fix: iconPosition does not exist on ButtonMenu

@nexxtway/react-rainbow


src/components/ButtonMenu/index.d.ts
```diff
@@ -3,6 +3,7 @@ import { BaseProps, ButtonIconVariant, ButtonIconSize, MenuSize, MenuAlignment }
 
 export interface ButtonMenuProps extends BaseProps {
     icon?: ReactNode;
+    iconPosition: string;
```
LeandroTorresSicilia, line 6: we should add `?` that means it will not be required

LeandroTorresSicilia (approved):

Issue #1471, opened 2020-04-08
fix: Lato-Black font error


krichdev:
Receiving an error about the Lato-Black font in the Application component.

-   [x] I have searched the [issues](https://github.com/nexxtway/react-rainbow/issues) of this repository and believe that this is not a duplicate.

## Describe the bug
Add library to project and tried using the ProgressIndicator and ProgressStep components. Would get the error below in console and browser.

## Screenshots
Console error
```
.../node_modules/react-rainbow-components/components/Application/fonts/Lato-Black/Lato-Black.eot:1
��


SyntaxError: Invalid or unexpected token
    at Module._compile (internal/modules/cjs/loader.js:891:18)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:991:10)
    at Module.load (internal/modules/cjs/loader.js:811:32)
    at Function.Module._load (internal/modules/cjs/loader.js:723:14)
    at Module.require (internal/modules/cjs/loader.js:848:19)
    at require (internal/modules/cjs/helpers.js:74:18)
    at Object.<anonymous> (/node_modules/react-rainbow-components/components/Application/rainbowLegacyStyles.js:10:41)
    at Module._compile (internal/modules/cjs/loader.js:955:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:991:10)
    at Module.load (internal/modules/cjs/loader.js:811:32)
    at Function.Module._load (internal/modules/cjs/loader.js:723:14)
    at Module.require (internal/modules/cjs/loader.js:848:19)
    at require (internal/modules/cjs/helpers.js:74:18)
    at Object.<anonymous> (/node_modules/react-rainbow-components/components/Application/index.js:22:51)
    at Module._compile (internal/modules/cjs/loader.js:955:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:991:10)
```

## To Reproduce 🕹

1. Install library
2. Import ProgressIndicator and ProgressStep component
3. Include above components in your own component

## Expected Behavior 🤔

Ability to render the components without error.

## Current Behavior 😯

Page will not load and displays the error about the font

## Context 🔦
Just simply trying to add two components to a page

## Your Environment 🌎

**Desktop (please complete the following information):**

-   OS: [Mac OS]
-   Browser [chrome]
-   Version [80.0.3987.149]
-   Framework: [NextJS v. 9.3.4]

TahimiLeonBravo:
[...]

PR #2140, opened 2021-01-31
fix: fix warning with legacy styles in SSR

wildergd:
#1471 

- fix legacy styles string escaping in SSR

NOTE: 
this will remove warning described in this comment `https://github.com/nexxtway/react-rainbow/issues/1471#issuecomment-765706675`. However a new Warning will appear `Warning: Expected server HTML to contain a matching <style> in <div>`. it is a common warning and research indicates that is not a real problem. So, it can be ignored.   

@nexxtway/react-rainbow


src/components/Application/rainbowLegacyStyles.js
```diff
@@ -1014,3 +1018,16 @@ export default `
   .rainbow-font-size-text_large {
     font-size: 1rem; }  
 `;
+
+const GlobalStyles = createGlobalStyle`
+    ${legacyStyles}
+`;
```
LeandroTorresSicilia, line 1024: let's put this inside the ```if (isServer) {``` block thus it is only created when is in server

wildergd: i did this 

```js
const RainbowLegacyStyles = isServer
    ? createGlobalStyle`
    ${legacyStyles}
`
    : () => <style>{legacyStyles}</style>;

export default RainbowLegacyStyles;
```
cos i a not sure how will be the impact on performance if we put `createGlobalStyle` insinde component function

LeandroTorresSicilia (approved):

PR #2156, opened 2021-02-10
fix: update z-index on portals

HellWolf93:
fix: #2155


src/styles/z-index.js
```diff
@@ -0,0 +1,4 @@
+export const zindexSpinner = 9000;
+export const zindexModal = 5000;
+export const zindexDrawer = 5000;
+export const zindexOverlay = 5000;
```
LeandroTorresSicilia, line 4: good job! here we can keep it centralized, only let's call this file `zIndex.js` in camelCase to keep consistency in the name convention

LeandroTorresSicilia: also to keep consistency, name these variables as ```ZINDEX_SPINNER``` as we do in colors.js, fontSizes.js etc

LeandroTorresSicilia (approved):

Issue #2205, opened 2021-03-25
feat: option to set fill and stroke colors on the Rating component


ldelia:
It would be nice to be able to change the color of the stars on the Rating component.

PR #2217, opened 2021-05-05
feat: add option to set fill and stroke colors on rating component

gtallesb:
fix: #2205 

-add starFillColor prop to change the color of the star as well as the stroke

@nexxtway/react-rainbow


TahimiLeonBravo:
@gtallesb I think that we should not add a new prop for this, we should add the color to the theme palette and use this color on the component. When the customer wants to change the component color he can change this color on the palette

PR #2267, opened 2021-08-03
fix: close lookup when `Esc` key is pressed

HellWolf93:
- Close Lookup when `Esc` key is pressed
- Add integration test

### DIFF
--- a/src/components/Lookup/index.js
+++ b/src/components/Lookup/index.js
@@ -256,10 +256,7 @@ class Lookup extends Component {
         const { searchValue } = this.state;
         const { keyCode } = event;
 
-        console.log('handleKeyDOwn');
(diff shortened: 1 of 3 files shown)


src/components/Lookup/index.js
```diff
@@ -254,8 +255,17 @@ class Lookup extends Component {
     handleKeyDown(event) {
         const { searchValue } = this.state;
         const { keyCode } = event;
-        if (keyCode === ESCAPE_KEY && !!searchValue) {
-            event.stopPropagation();
+
+        console.log('handleKeyDOwn');
+        if (keyCode === ESCAPE_KEY) {
+            console.log(searchValue);
+            console.log(this.isLookupOpen());
```
LeandroTorresSicilia, line 262: remove all these console

PR #2292, opened 2021-09-26
fix: drawer closing when resizing a column from a table inside drawer to outside

wildergd:
fix: #

-

@nexxtway/react-rainbow


wildergd:
the only way i was able to make it work was to use stopPropagation as the issue is caused by the `mouseup` event itself... that also triggers `click` event on the same element... and also taget for both `mouseup` and `click` takes the drawer backdrop as target..

Issue #2285, opened 2021-09-21
fix: Picklist onBlur fires on click


jkauszler:
The Picklist component onBlur prop is firing at the incorrect time. As of V1.27.0 onBlur fires right after onClick before a selection or clicking away happens.

-   [x] I have searched the [issues](https://github.com/nexxtway/react-rainbow/issues) of this repository and believe that this is not a duplicate.

## Describe the bug

When using the Picklist component onBlur prop, onBlur fires as the selection window is being displayed after the initial click.

## To Reproduce 🕹

1. Go to the Picklist example page (https://react-rainbow.io/#/Picklist).
2. Edit one of the code examples to include the `onBlur` prop
3. Set the prop value equal to `{(x) => console.log(x)}`
4. Open your console
5. Click on the example picklist ui to open the window
6. Notice that `null` or it's value has been returned
7. Click outside of the picklist ui or an option to close the picklist window.
8. Notice that there is not an onBlur console fire being returned.

## Expected Behavior 🤔

I expect onBlur to fire after either after a selection within the picklist window has been made, a tab keypress to another input, or a click outside of the picklist window. It would also be nice if onBlur returned itself rather than just its value, like other inputs.

## Current Behavior 😯

Currently onBlur fires before a selection or closing of the Picklist window.

## Context 🔦

Im trying to validate selections made from the picklist after one has been selected or if the window has been closed without selecting an option.

**Desktop (please complete the following information):**

-   OS: OS X
-   Browser: Chrome 
-   Version: 91.0.4472.164 (Official Build) (arm64)

PR #2303, opened 2021-10-19
fix: `onBlur` event on Picklist fires on click

HellWolf93:
fix: #2285 

- Fixed bug that was firing onBlur event when Picklist dropdown opens


HellWolf93:
The issue also mentions changing the `onBlur` parameter from value to the synthetic event. I think this is not feasible because it will be a breaking change, but maybe we could add the event as a second parameter

PR #2320, opened 2021-11-30
fix: improve ColorInput component

HellWolf93:


TahimiLeonBravo (approved):

src/components/ColorInput/__test__/colorInput.spec.js
```diff
@@ -69,13 +69,78 @@ describe('<ColorInput />', () => {
...
+        const value = { hex: '#ccc', alpha: 1, isValid: true };
+        const component = mount(<ColorInput value={value} />);
+        expect(component.find('ColorSample').prop('value')).toEqual(value);
+    });
+
+    it('should update the color sample on change', () => {
+        const value = { hex: '#ccc', alpha: 1, isValid: true };
+        const component = mount(<ColorInput value={value} />);
+        component.find('input[type="text"]').simulate('change', { target: { value: '#fff' } });
+        expect(component.find('ColorSample').prop('value')).toEqual({
```
LeandroTorresSicilia, line 119: in component testing we should focus on the final result instead of internal details, for example instead of checking for props passed to internal component or checking states we should check for the color rendered, then we should find the HTML element that render this and check what color it have

LeandroTorresSicilia (approved):

src/components/ColorInput/__test__/colorInput.spec.js
HellWolf93: fixed in #2323

PR #2362, opened 2022-01-26
feat: add `headerComponent` prop to the column component

yvmunayev:
feat: add `headerComponent` prop to the column component

-

@nexxtway/react-rainbow

### DIFF
--- a/src/components/Table/head/header.js
+++ b/src/components/Table/head/header.js
@@ -52,12 +52,8 @@ export default class Header extends Component {
 
     render() {
         const {
-            content,
-            sortDirection,
             minColumnWidth,
             maxColumnWidth,
-            sortable,
-            isSorted,
             computedWidth,
             type,
             onSelectAllRows,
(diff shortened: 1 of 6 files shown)


src/components/Table/head/header.js
```diff
@@ -98,25 +97,31 @@ export default class Header extends Component {
...
-                        role="presentation"
-                        headerAlignment={headerAlignment}
-                        onClick={this.handleSort}
-                    >
-                        <StyledContent
-                            title={this.getHeaderContent()}
-                            className="rainbow-table_header-content"
+                    <RenderIf isTrue={HeaderComponent}>
+                        {/* eslint-disable-next-line react/jsx-props-no-spreading */}
+                        <HeaderComponent {...rest} onSort={this.handleSort} />
```
LeandroTorresSicilia, line 102: here we need to rename props, the props need to be the same as main props, for example the column does not have any content props, I assume that this is mapped to header, if so then we need to use header instead
Also I think we should avoid passing sortable since I don't see for what they will need this, it is only used internally to send the onSort event

LeandroTorresSicilia, line 102: here change content for header

LeandroTorresSicilia (approved):

LeandroTorresSicilia (approved):

PR #2385, opened 2022-03-17
fix: add empty component prop

amontalvof:
fix: #

-

@nexxtway/react-rainbow

### DIFF
--- a/src/components/InternalDropdown/index.js
+++ b/src/components/InternalDropdown/index.js
@@ -418,10 +418,7 @@ const InternalDropdown = forwardRef((props, reference) => {
                     </Content>
                 </Ul>
                 <RenderIf isTrue={showEmptyMessage && EmptyComponent}>
-                    <EmptyComponent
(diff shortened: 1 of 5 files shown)


src/components/InternalDropdown/index.js
```diff
@@ -416,7 +417,13 @@ const InternalDropdown = forwardRef((props, reference) => {
                         </Provider>
                     </Content>
                 </Ul>
-                <RenderIf isTrue={showEmptyMessage}>
+                <RenderIf isTrue={showEmptyMessage && EmptyComponent}>
+                    <EmptyComponent
+                        searchValue={searchValue}
+                        hasTimeout={!!searchTimeout.current}
```
LeandroTorresSicilia, line 423: here only pass searchValue, for now we don't have use case to use hasTimeout also it is not a meaningful prop name, then for now only searchValue

LeandroTorresSicilia (approved):

PR #2435, opened 2022-07-22
feat: add type declarations to libs and styles

HellWolf93:


src/libs/debounce/index.d.ts
```diff
@@ -0,0 +1 @@
+export default function(callback: Function, time: number): Function;
```
LeandroTorresSicilia, line 1: here time is not required

src/libs/ResizeSensor/index.d.ts
```diff
@@ -0,0 +1,9 @@
+declare class ResizeSensor {
+    constructor(
+        element: Element,
+        resizeListener: (size: { width: number; height: number }) => void,
```
LeandroTorresSicilia, line 4: I see that resizeListener is used inside attachResizeEvent and there we don't pass any argument, then size argument should be optional

src/libs/hooks/useDisclosure.d.ts
```diff
@@ -0,0 +1,9 @@
+interface Disclosure {
+    isOpen: boolean;
+    open: () => void;
+    close: () => void;
+    toggle: () => void;
+}
+
+declare function useDisclosure(defaultIsOpen: any): Disclosure;
```
LeandroTorresSicilia, line 8: we should not use any

LeandroTorresSicilia, line 8: also here defaultIsOpen should be optional

src/libs/hooks/useErrorMessageId.d.ts
```diff
@@ -0,0 +1,2 @@
+declare function useErrorMessageId(error: any): string | undefined;
```
LeandroTorresSicilia, line 1: same ^^ no any and optional

src/libs/hooks/useLabelId.d.ts
```diff
@@ -0,0 +1,2 @@
+declare function useLabelId(label: any): string | undefined;
```
LeandroTorresSicilia, line 1: same ^^

src/libs/hooks/useLocale.d.ts
```diff
@@ -0,0 +1,2 @@
+declare function useLocale(localProp: string): string;
```
LeandroTorresSicilia, line 1: optional

src/libs/hooks/useUniqueIdentifier.d.ts
```diff
@@ -0,0 +1,2 @@
+declare function useUniqueIdentifier(prefix: string): string;
```
LeandroTorresSicilia, line 1: optional

LeandroTorresSicilia (approved):

Issue #2452, opened 2022-09-25
feat: add input prop valueAlignment and deprecate isCentered


yvmunayev:
feat: add input prop valueAlignment = “left” | “center” | “right” and deprecate isCentered

## Context 🔦

**Describe the solution you'd like**

**Describe alternatives you've considered**

PR #2453, opened 2022-09-25
feat: add input prop valueAlignment and deprecate isCentered

yvmunayev:
fix: #2452

- feat: add input prop valueAlignment and deprecate isCentered

@nexxtway/react-rainbow

### DIFF
--- a/src/components/Input/inputBase/index.js
+++ b/src/components/Input/inputBase/index.js
@@ -122,9 +122,11 @@ export default class InputBase extends Component {
             isCentered,
             iconPosition,
             variant,
+            valueAlignment,
         } = this.props;
         const isReadOnly = !!(!disabled && readOnly);
         const isPassword = type === 'password';
+        const alignment = isCentered ? 'center' : valueAlignment;
 
         return (
             <StyledContainer id={id} className={className} style={style}>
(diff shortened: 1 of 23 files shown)


src/components/Input/inputBase/index.js
```diff
@@ -122,9 +122,11 @@ export default class InputBase extends Component {
             isCentered,
             iconPosition,
             variant,
+            valueAlignment,
         } = this.props;
         const isReadOnly = !!(!disabled && readOnly);
         const isPassword = type === 'password';
+        const alignment = isCentered ? 'center' : valueAlignment;
```
LeandroTorresSicilia, line 129: here you are giving priority, to `isCentered` but since we will deprecate it, we should do the inverse, then let's give priority to the `valueAlignment` over the `isCentered`

src/components/Input/pickerInput/index.js
```diff
@@ -62,7 +62,8 @@ export default class Input extends Component {
 
     renderInput() {
         const { props } = this;
-        const { readOnly } = props;
+        const { readOnly, isCentered, valueAlignment } = props;
+        const alignment = isCentered ? 'center' : valueAlignment;
```
LeandroTorresSicilia, line 66: same ^^ here

LeandroTorresSicilia (approved):

Issue #2458, opened 2022-09-26
feat: add options declaration type to chart component


yvmunayev:
feat: add options declaration type to chart component

PR #2459, opened 2022-09-26
feat: add options declaration type to chart component

yvmunayev:
fix: #2458

- feat: add options declaration type to chart component

@nexxtway/react-rainbow

### DIFF
--- a/src/components/Chart/index.d.ts
+++ b/src/components/Chart/index.d.ts
@@ -1,3 +1,4 @@
+import { ChartOptions } from 'chart.js';
 import { ReactNode, ComponentType } from 'react';
 import { BaseProps } from '../types';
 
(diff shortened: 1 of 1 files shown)


src/components/Chart/index.d.ts
```diff
@@ -19,6 +20,8 @@ export interface ChartProps extends BaseProps {
     maintainAspectRatio?: boolean;
     plugins?: Array<object>;
     children?: ReactNode;
+    options?: ChartOptions;
+    [key: string]: unknown;
```
LeandroTorresSicilia, line 24: I think here we only need to add options, the `[key: string]: unknown;` is only needed for dataset where you can pass extra props defined by plugins but here I don't think we need this

LeandroTorresSicilia (approved):

Issue #2533, opened 2022-11-12
fix: carousel card component using set timeout on react18


yvmunayev:
-   [X] I have searched the [issues](https://github.com/nexxtway/react-rainbow/issues) of this repository and believe that this is not a duplicate.

## Describe the bug

fix: carousel card component using set timeout on react18

PR #2534, opened 2022-11-12
fix: carousel card component using set timeout on react18

yvmunayev:
fix: #2533 

- fix: carousel card component using set timeout on react18

@nexxtway/react-rainbow


LeandroTorresSicilia:
@yvmunayev there are two integration tests failing

Screen Shot 2022-12-03 at 2 13 09 PM

src/components/Accordion/index.js
```diff
@@ -46,7 +46,7 @@ const Accordion = props => {
 
     const selectAccordionSection = side => {
         const accordionSectionIndex = childrenRegistered.findIndex(
-            section => section.name === currentSection,
+            section => section.id === currentSection,
```
LeandroTorresSicilia, line 49: why change name for id here?

src/libs/hooks/usePrevious.js
```diff
@@ -0,0 +1,11 @@
+import { useRef, useEffect } from 'react';
+
+const usePrevious = value => {
+    const ref = useRef();
+    useEffect(() => {
+        ref.current = value;
+    });
```
LeandroTorresSicilia, line 7: I think here you should use an array of dependencies with `value`

LeandroTorresSicilia (approved):

Issue #2539, opened 2022-11-19
feat: finish to add prop size to some inputables


LeandroTorresSicilia:
ColorInput
StripeCardInput
Textarea
FileSelector

Note: remember to change label size too

PR #2574, opened 2023-01-17
feat: finish to add prop size to some inputables

yvmunayev:
fix: #2539 

- feat: add prop size to ColorInput
- feat: add prop size to StripeCardInput
- feat: add prop size to Textarea
- feat: add prop size to FileSelector
- fix: behavior of FileSelector when have error and is disabled
- fix: behavior of FileSelector when have error and is drag over

@nexxtway/react-rainbow


src/components/FileSelector/styled/index.js
```diff
@@ -156,7 +190,18 @@ export const StyledBackdrop = attachThemeAttrs(styled.div)`
         background: transparent;
         height: 100%;
         width: 100%;
-        `}
+        
+        ${props.size === 'large' &&
```
LeandroTorresSicilia, line 194: we need to discuss with @TahimiLeonBravo how multiline will look like when using sizes

TahimiLeonBravo (approved):

Issue #2577, opened 2023-01-18
fix: assigned field on the mapping modal can't be removed [ImportRecordFlow]


TahimiLeonBravo:

TahimiLeonBravo:
As a customer, I want to remove an assigned field on the mapping table.

**Current Behavior:**
When the customer assigns a value to a field he can't remove it anymore. The "Assign" button is disabled.

https://user-images.githubusercontent.com/32083556/213309650-e4b3f71c-ad0f-41fc-bdef-e399ffd88c7d.mov

**Expected Behavior:**
The customer should be able to remove assigned fields at any time.

PR #2580, opened 2023-01-23
fix: assigned field on the mapping modal cant be removed on importRecordFlow

yvmunayev:
fix: #2577 

- fix: assigned field on the mapping modal cant be removed on importRecordFlow

@nexxtway/react-rainbow


src/components/ImportRecordsFlow/stepThree/assignFieldModal.js
```diff
@@ -72,13 +71,7 @@ export default function AssignFieldModal(props) {
             isOpen={isAssignFieldModalOpen}
             size="small"
             onRequestClose={onRequestClose}
-            footer={
-                <AssignFieldModalFooter
-                    onCancel={onRequestClose}
-                    onAssign={handleAssign}
-                    isAssignButtonDisabled={isAssignButtonDisabled}
```
LeandroTorresSicilia, line 79: the intention here is to keep this button disabled only when the field is not assigned, then we need to fix the current behavior but also keep the behavior of having this button disabled when the field is not assigned

LeandroTorresSicilia (approved):

PR #2592, opened 2023-01-30
fix: change step one import record flow styles

adrianestevez1:
fix: #

-changing styles of the step one in the ImportRecordFlow component

@nexxtway/react-rainbow

### DIFF
--- a/src/components/ImportRecordsFlow/stepOne/index.js
+++ b/src/components/ImportRecordsFlow/stepOne/index.js
@@ -61,6 +61,7 @@ export default function StepOne(props) {
 
                 <RenderIf isTrue={isMergeOption}>
                     <Select
+                        style={{ marginTop: '15px' }}
                         label="Match Field"
                         placeholder="Select the Field do you want match"
                         options={fieldOptions}
--- a/src/components/ImportRecordsFlow/stepOne/index.js
+++ b/src/components/ImportRecordsFlow/stepOne/index.js
@@ -1,11 +1,11 @@
 import React, { useState, useEffect } from 'react';
 import PropTypes from 'prop-types';
 import RenderIf from '../../RenderIf';
-import Select from '../../Select';
 import getSchemaFieldOptions from '../helpers/getSchemaFieldOptions';
 import StyledContainer from './styled/container';
 import { StyledRadioGroup } from './styled/optionPicker';
 import OptionPicker from './optionPicker';
+import StyledSelect from './styled/styledSelect';
 
 export default function StepOne(props) {
     const { schemaFields, actionOption, onChangeAction, matchField, onChangeMatchField } = props;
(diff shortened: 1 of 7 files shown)


TahimiLeonBravo:
@adrianestevez1 fix layout of the modal is moving when hover over item.

https://user-images.githubusercontent.com/32083556/216672937-8a4a862f-77d2-4abc-9312-0619b379722f.mov

src/components/ImportRecordsFlow/stepOne/index.js
```diff
@@ -19,36 +16,61 @@ export default function StepOne(props) {
...
+    ];
+
+    return (
+        <>
+            <StyledContainer>
+                <StyledRadioGroup
+                    onChange={onChange}
+                    value={actionOption}
+                    options={options}
+                    id="radio-group-component-1"
```
LeandroTorresSicilia, line 59: what is this ID used for? remember that ids needs to be unique, first we need to avoid using id at all inside components but if it is completely necessary then we need to generate one instead of hard coded

src/components/ImportRecordsFlow/stepOne/index.js
```diff
@@ -19,36 +16,61 @@ export default function StepOne(props) {
...
+                />
+
+                <RenderIf isTrue={isMergeOption}>
+                    <Select
+                        style={{ marginTop: '15px' }}
```
LeandroTorresSicilia, line 64: no inline styles, intead use a styled component

src/components/ImportRecordsFlow/stepOne/styled/optionPicker.js
```diff
@@ -0,0 +1,94 @@
+import styled from 'styled-components';
+import File from '../../icons/file';
+import attachThemeAttrs from '../../../../styles/helpers/attachThemeAttrs';
+import RadioGroup from '../../../RadioGroup';
+import MergeFiles from '../../icons/mergeFiles';
+
+export const OptionContainer = attachThemeAttrs(styled.li)`
```
LeandroTorresSicilia, line 7: you can use li elements to define each item but for this you need to wrap it in ol or ul element, look here: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/li
`It must be contained in a parent element: an ordered list <ol>, an unordered list <ul>`

Here there is an example with all the accessibility specs where you can use ul and li for the radio buttons combined with the attribute role: https://www.w3.org/WAI/ARIA/apg/patterns/radio/examples/radio-activedescendant/

Another option is to simply use div everywhere with the use of role like this: https://www.w3.org/WAI/ARIA/apg/patterns/radio/examples/radio/

But since we are using input and labels I think the easiest solution is to simply change li for a span and keep the picker like this one: https://www.lightningdesignsystem.com/components/visual-picker/

TahimiLeonBravo (approved):

LeandroTorresSicilia (approved):

Issue #2628, opened 2023-07-28
fix: GoogleAddressLookup size grows for selected address with long text 


wildergd:
When using GoogleAddressLookup and we select a place with a long text.. the input is enlarged instead of cropping the text.

-   [x] I have searched the [issues](https://github.com/nexxtway/react-rainbow/issues) of this repository and believe that this is not a duplicate.

## Describe the bug

When using GoogleAddressLookup and we select a place with a long text.. the input is enlarged instead of cropping the text.

##Screenshots

https://github.com/nexxtway/react-rainbow/assets/3240995/67a2f273-d23a-4cc8-8b98-17651469d8cb

## To Reproduce 🕹

1. Go to GoogleAddressLookup example
2. Click on Show Code
3. change maxWith in containerStyles to 300px
4. in the input search for San Francisco and select San Francisco Airport

## Expected Behavior 🤔

it should cropt the text and keep input size

## Current Behavior 😯
the input value is enlarged

PR #2629, opened 2023-07-28
style: fix Lookup styles

wildergd:
fix: #2628

- fix styles in Lookup component

@nexxtway/react-rainbow


src/components/Lookup/styled/chip.js
```diff
@@ -6,6 +6,12 @@ const StyledChip = attachThemeAttrs(styled(Chip))`
     flex: 1 1 auto;
     margin: 2px;
     background-color: ${props => props.palette.background.main};
+    width: 0px;
+
+    & > span {
+        min-width: 0px;
+        flex-basis: 99%;    
+    }
```
wildergd, line 14: this is because changing the value directly in the Structured component... a lot of components are affected .. so better to change it here to only affect the Chip associated with the lookup value

